PR #2, opened 2022-04-28
Initial cleanup

### LINKED ISSUE #1: Replace copyright holder
Use F. Hoffmann-La Roche Ltd., 2022 instead of openpharma. Can discuss again when somebody else is contributing code.

danielinteractive:
closes #1


danielinteractive:
@dinakar29 I had some issue here with Pkgdown Docs/ Generate check, but I assume this might be because there is almost nothing in the pkgdown yet? maybe you can have a quick look? thx!

dinakar29:
> @dinakar29 I had some issue here with Pkgdown Docs/ Generate check, but I assume this might be because there is almost nothing in the pkgdown yet? maybe you can have a quick look? thx!

Yes, this is actually an easy fix. The issue herein is related to one or more missing topics in the `_pkgdown.yml` config or one or more stray files. Unfortunately the new version of `pkgdown` doesn't provide any proper logging or error reporting for this, so it's difficult to troubleshoot this.

I'll provide a fix.

danielinteractive:
thx @dinakar29 !

PR #4, opened 2022-04-28
Make pkgdown work

dinakar29:
`pkgdown` thinks that the PR template is a legitimate markdown file without any content in it 😞 
Adding a header to it so it passes.


dinakar29:
@danielinteractive This is the fix for the pkgdown issue 😸

danielinteractive (approved):
:-D I would never ever have guess this!

PR #13, opened 2022-05-13
adding h_build_formula

juliadedic1:
adding the h_build_formula helper function. There is an error when running the checks. Here is the message in the log:
* checking PDF version of manual ... WARNING
LaTeX errors when creating PDF version.
This typically indicates Rd problems.
* checking PDF version of manual without hyperrefs or index ... ERROR
Re-running with no redirection of stdout/stderr.
* DONE
Status: 1 ERROR, 1 WARNING

Also, I'm not sure if my unittests cover everything. Do we need to add an additional test if the `vars` argument is not specified properly?


danielinteractive:
Thanks @juliadedic1 , in terms of checks the relevant thing in the end is what we see here in github checks, and there I just find one misspelling:
image

R/formula.R
```diff
@@ -0,0 +1,64 @@
+#' Building formula for model
```
danielinteractive, line 1:
```suggestion
#' Building Model Formula
```
(just a bit more title style)

R/formula.R
```diff
@@ -0,0 +1,64 @@
...
+#'
+#' @description `r lifecycle::badge("experimental")`
+#'
+#' @param vars (`list`)\cr variables to use in the model
```
danielinteractive, line 5:
```suggestion
#' @param vars (`list`)\cr variables to use in the model.
```

R/formula.R
```diff
@@ -0,0 +1,64 @@
...
+#' @param cor_struct (`character`)\cr speicify the covariance structure to use
```
danielinteractive, line 6:
```suggestion
#' @param cor_struct (`string`)\cr specify the covariance structure to use.
```
note: string here because the user should just specify a single covariance structure

tests/testthat/test-formula.R
```diff
@@ -0,0 +1,12 @@
+library(testthat)
+
+test_that("h_build_formula gives error message if incorrect inputs given", {
```
danielinteractive, line 3:
```suggestion
test_that("h_build_formula gives error message if incorrect covariance structure is given", {
```

danielinteractive (commented):
thx @juliadedic1 , looks good! re: tests, you can here 
1) add one more test when no `arm` is specified in `vars`
2) when the function works, not just test for silence, but save the `result` and compare with `expected` result

danielinteractive:
@juliadedic1 thx, next time no need to close this PR and make a new one - you can just continue working on the old PR.

PR #17, opened 2022-05-16
12 add h build formula@main

juliadedic1:
Added additional unittests based on previous PR. 

I get this message with the new unittests:

Error in reporter$stop_if_needed() : Test failed
Error: `result` (`actual`) not identical to `expected` (`expected`).

`attr(actual, '.Environment')` is <env:0x0000023b4392eb50>
`attr(expected, '.Environment')` is <env:global>

They are identical results besides that bit above about the attributes.


danielinteractive (commented):
thx Julia, you can try either:
1) `expect_equivalent` instead of `expect_identical`
2) compare the character representation instead of the formulas themselves, so `expect_identical(as.character(result), as.character(expected))`
I guess 1) or otherwise 2) should work.

juliadedic1:
@danielinteractive Thanks!! Option 2 worked. just FYI it says that `expect_equivalent()` was deprecated in the 3rd edition.

juliadedic1:
Thanks @danielinteractive :)

danielinteractive (approved):

Issue #16, opened 2022-05-16
Add function `fit_single_optimizer()`


danielinteractive:
To do:

- [x] See design document [here](https://github.com/openpharma/mmrm/blob/9da1d5ef01ec78b7cc030a2fce2c4edf18f961a3/design/design_fit.Rmd#L371)
- [x] Add function `fit_single_optimizer()` in file `R/fit.R`
	- [x] Modify the prototype such that also `nlminb` could be used
	- [x] Don't include `Nelder-Mead` as an option since that does not make use of the automatic differentation in `glmmTMB`
- [x] Add unit tests in file `testthat/test-fit.R`
  - [x] If needed please improve the function code
  - [x] for the `start` values you can use `list(theta = c(3, 0.5))` e.g. in the case of AR1 model
- [x] Add `roxygen2` documentation chunks
  - [x] Run Build > Document, Build > Load All, and check that documentation looks good
- [x] Add function name to `pkgdown.yaml`
- [x] Run Build > Check to ensure that checks pass

```r
fit_single_optimizer <- function(formula,
                                 data,
                                 start = NULL,
                                 optimizer = c("L-BFGS-B", "BFGS", "CG", "nlminb")) 
```

and then inside the function:

```r
control <- glmmTMB::glmmTMBControl(
    optimizer = if (optimizer == "nlminb") stats::nlminb else stats::optim,
    optArgs = if (optimizer == "nlminb") list() else list(method = optimizer),
    parallel = 1L
  )
```

juliadedic1:
@danielinteractive no questions as of yet. Hoping to finish it up today. Thanks for checking in!

juliadedic1:
@danielinteractive should this function be added to the "functions" section in the `pkgdown.yaml`? Also, I am not sure about the unit-tests I created for this one. I will push what I have and we can discuss tomorrow in our stand-up!

danielinteractive:
@juliadedic1 yeah please add to the functions section, yes let's have a look tomorrow!

PR #23, opened 2022-06-02
adding fit_single_optimizer function. Unit tests are not fully workin…

juliadedic1:
PR for issue #16. I am not sure about the unit tests for this one. The ones commented out fail as the results are not the same. I've added another test that works but it's not testing the result of the function.


tests/testthat/test-fit.R
```diff
@@ -0,0 +1,55 @@
...
+#     data = dat,
+#     dispformula = ~0,
+#     start = NULL,
+#     control = glmmTMB::glmmTMBControl(optimizer = stats::optim, optArgs = list(method = "BFGS"), parallel = 1)
+#   )
+#   expect_identical(result, expected) # failed, not getting the same values
+#   # expect_equal(result, expected, tolerance = 1)
+# })
+
+#or should the test be more like this
```
danielinteractive, line 20: yeah I would do more like this, no need to compare identity of the whole result

danielinteractive (approved):

PR #35, opened 2022-06-23
Add Satterthwaite d.f. and emmeans support.

### LINKED ISSUE #33: Add more functions towards Satterthwaite and LS means functionality
To do:

- [x] See https://github.com/openpharma/mmrm/blob/main/design/design_fit.Rmd
- [x] `h_covbeta_fun()` (this should be much simpler now)
- [x] `h_jac_col_as_matrix()` (factored out helper)
- [x] `h_jac_list()`
- [x] modify `fit_model()` a la `mmrm()` and rename it to `mmrm()`
  - add the Jacobian calculation on top
  - add `mmrm` class to existing classes for return object, such that we know that we have the Jacobian in there
  - rename `mmrm_tmb()` to `h_mmrm_tmb()` to avoid confusion
- [x] diagnostics:
  - [x] `deviance()`
  - [x] `AIC()`
  - [x] `BIC()`
- [x] `h_quad_form_vec()`
- [x] `h_gradient()`
- [x] `h_df_1d_list()`
- [x] `df_1d()`
- [x] `h_quad_form_mat()`
- [x] `h_df_md_list()`
- [x] `h_md_denom_df()`
- [x] `df_md()`
- [x] `recover_data.mmrm()` needs
  - [x] call saved
  - [x] `terms()` method
- [x] `emm_basis.mmrm()` - also here more needs to be done

Note that we don't need:
- `h_general_jac_list()` (since no other jacobian calculations besides from our TMB fit)
- `vars()` (since we can stick with the formula interface for this package)
- `h_vcov_theta()` (since we have that explicitly as part of the object already)
- separate `mmrm()` as another wrapper of `fit_model()` (since we don't need vars interface and we can just calculate Jacobian in `fit_model()` directly)
- `diagnostics()` (this is more a wrapper for `tern.mmrm`)

danielinteractive:
closes #33

To review:

- [x] Install from this feature branch
- [x] Try out:
  - [x] `mmrm()` (replaces `fit_model()`)
  - [x] `df_1d()`
  - [x] `df_md()`
  - [x] `emmeans()` -> see `?emmeans_support`, need to have `emmeans` package installed


ywang-gilead:
Tried out functions in two datasets, one is the built-in dataset fev_data and the other is from a study. Both worked well.

ywang-gilead (approved):

PR #58, opened 2022-07-04
54 add a getter function component for mmrm tmb objects

BrianLang:
proposal for `component()`.
* updated `summary.mmrm` to use `component`
* updated `print.mmrm_tmb` to use `component`
* updated testing for `summary.mmrm` for trivial changes to ordering in list. 

probably still need to dig into using `component` in other places, how thorough should I be?


BrianLang:
did the classic mistake of developing and pushing on the wrong branch. this quasi undid the rebase (squash). This doesn't matter, does it?

danielinteractive:
thanks @BrianLang , does not matter I think since we anyway squash and merge at the end.

R/mmrm-methods.R
```diff
@@ -48,26 +48,12 @@ h_coef_table <- function(object) {
...
-      n_subjects = object$tmb_data$n_subjects,
-      n_timepoints = object$tmb_data$n_visits,
-      n_obs = length(object$tmb_data$y_vector),
-      coefficients = h_coef_table(object),
-      vcov = vcov(object),
-      varcor = VarCorr(object),
-      aic_list = aic_list,
-      call = stats::getCall(object)
-    ),
+    Reduce(append,
```
danielinteractive, line 52: hm this makes the code less readable, maybe you could instead just `c()` together the `aic_list` with the `components` list?

R/tmb-methods.R
```diff
@@ -166,33 +166,112 @@ print.mmrm_tmb <- function(x,
                            ...) {
   cat("mmrm fit\n\n")
 
-  n_subjects <- x$tmb_data$n_subjects
-  n_timepoints <- x$tmb_data$n_visits
-  n_obs <- length(x$tmb_data$y_vector)
-  cov_type <- x$formula_parts$corr_type
-  n_theta <- length(x$theta_est)
+  components <- component(x, c(
```
danielinteractive, line 169: here instead of saving everything in large list I think it is better readable to just call each component where needed below.

R/tmb-methods.R
```diff
@@ -166,33 +166,112 @@ print.mmrm_tmb <- function(x,
...
+    "deviance", "cov_type", "n_theta", "n_subjects",
+    "n_timepoints", "n_obs", "reml", "method",
+    "convergence", "evaluations", "conv_message", "call"
+  ))
 
-  h_print_call(x$call, n_obs, n_subjects, n_timepoints)
-  h_print_cov(cov_type, n_theta)
+  h_print_call(
+    components$call, components$n_obs,
+    components$n_subjects, components$n_timepoints
```
danielinteractive, line 177: e.g. here spend one line for each argument, name arguments and then use `component(x, ...)` to get it.

R/tmb-methods.R
```diff
@@ -166,33 +166,112 @@ print.mmrm_tmb <- function(x,
...
+component <- function(object,
+                           name = c(
+                             "AIC", "BIC", "logLik", "deviance",
+                             "cov_type", "n_theta", "n_subjects", "n_timepoints",
+                             "n_obs", "vcov", "varcor", "formula", "dataset",
+                             "reml", "method", "convergence", "evaluations",
+                             "conv_message", "call"
+                           ),
+                           ...) {
+  inherits(object, "mmrm_tmb")
```
danielinteractive, line 223:
```suggestion
  assert_class(object, "mmrm_tmb")
```

R/tmb-methods.R
```diff
@@ -166,33 +166,112 @@ print.mmrm_tmb <- function(x,
...
+  if (missing(name)) stop("'name' must not be missing")
```
danielinteractive, line 224: instead could be better to use `match.arg()` and allow multiple matches

R/tmb-methods.R
```diff
@@ -166,33 +166,112 @@ print.mmrm_tmb <- function(x,
...
+  ## Deal with multiple names -- "FIXME" is inefficiently redoing things
```
danielinteractive, line 225: if something needs to be done later, please make a new issue and refer to this function, generally we don't want any `fixme` or `todo` messages in the code base. for comments please use `# Sentence style.`

R/tmb-methods.R
```diff
@@ -166,33 +166,112 @@ print.mmrm_tmb <- function(x,
...
+  if (length(name <- as.character(name)) > 1) {
+    names(name) <- name
+    return(lapply(name, component, object = object))
+  }
+  if (tolower(name) == "all") { ## recursively get all provided components
+    return(sapply(eval(formals()$name),
+      component,
+      object = object, simplify = FALSE
+    ))
+  }
```
danielinteractive, line 235: this one I am not sure we need. maybe it could be elegant that by default when not specifying a name explicitly all components are returned?

R/tmb-methods.R
```diff
@@ -166,33 +166,112 @@ print.mmrm_tmb <- function(x,
...
+
+  stopifnot(inherits(object, "mmrm_tmb"))
```
danielinteractive, line 237:
```suggestion
```
we had this above now already

R/tmb-methods.R
```diff
@@ -166,33 +166,112 @@ print.mmrm_tmb <- function(x,
...
+  name <- match.arg(name)
```
danielinteractive, line 238: move to above

R/tmb-methods.R
```diff
@@ -166,33 +166,112 @@ print.mmrm_tmb <- function(x,
...
+
+  ### Start of the switch
+  switch(name,
```
danielinteractive, line 241: suggest to have a little helper function e.g. `h_single_component` that has this switch in it. then in `component` you could do an `lapply` over all `name`s.

danielinteractive: or even just lapply with the switch instead of separate helper.

R/tmb-methods.R
```diff
@@ -166,33 +166,112 @@ print.mmrm_tmb <- function(x,
...
+    "vcov" = vcov(object),
+    "varcor" = VarCorr(object),
+    "formula" = deparse(object$call$formula),
+    "dataset" = object$call$data,
+
+    "..foo.." = # placeholder!
+      stop(gettextf(
+        "'%s' is not implemented yet",
+        sprintf("component(*, \"%s\")", name)
+      )),
```
danielinteractive, line 270: this is not needed when we have below?

danielinteractive: ah ok this would be for future things. well I would omit this.

R/tmb-methods.R
```diff
@@ -166,33 +166,112 @@ print.mmrm_tmb <- function(x,
...
+    ## otherwise
+    stop(sprintf(
+      "Mixed-Effects extraction of '%s' is not available for class \"%s\"",
```
danielinteractive, line 273:
```suggestion
      "component '%s' is not available for class `mmrm_tmb`,
```

danielinteractive (commented):
thanks Brian for being so active! very cool. please see comments inside

R/tmb-methods.R
BrianLang: okay, noted, thanks :-)

R/tmb-methods.R
BrianLang: I agree. When we remove that, the default argument is then the list of all, meaning that we immediately return all.

R/tmb-methods.R
BrianLang: thanks!

R/tmb-methods.R
BrianLang: done

R/tmb-methods.R
BrianLang: done

R/tmb-methods.R
BrianLang: done

R/mmrm-methods.R
BrianLang: updated but unclear if it's an improvement.

R/tmb-methods.R
BrianLang: essentially have this implementation with the recursion. not quite clear what the lapply on the switch would look like (just learned about switch!).

danielinteractive: right, I guess it could be slightly more efficient to avoid the recursion here. otherwise it is fine of course.
I think something like `lapply(name, switch, "reml" = ..., "method" = ... etc)` could work, basically replacing the single switch call you have now

danielinteractive (commented):
thanks Brian, sorry for delay (was off today), please see `lapply()` pointer inside, good to start with tests I think

BrianLang:
Thanks, no worries on timing! Will update tomorrow

R/tmb-methods.R
BrianLang: thanks, makes sense. I've never really passed secondary arguments into `lapply` like that before. Nice.

danielinteractive:
@BrianLang yeah that would be cool!

tests/testthat/_snaps/tmb-methods.md
```diff
@@ -748,6 +748,33 @@
     $neg_log_lik
     [1] 1821.197
     
+    $jac_list
```
danielinteractive, line 751: I would maybe not save the whole matrix list etc.... otherwise we are going to get lots of failures on other servers etc.

danielinteractive: e.g. just expect a list with certain names (`expect_list`, `expect_named`) is probably sufficient

BrianLang: good tip, I had already noticed this when I had been adding the report(theta) function to the list. the environment is updated every run of the snapshot test and it fails!
To be clear, we leave $jac_list in the components, but only test the length and names of components(object)?

PR #60, opened 2022-07-06
fixes mismatches between dropped levels in data and in tmb_data

### LINKED ISSUE #59: Error when `AVISIT` column has SCREENING variable
please refer to the discussion on [tern#628](https://github.com/insightsengineering/tern/issues/628). This error should be more informative or fixed before the call.

To do:

- [x] define reprex for `mmrm`
- [x] include as unit tests
- [x] fix problem
- [x] confirm all checks pass

Melkiades:
The error was based on the fact that the TLG data-set had NAs on both sides of the formula used. These values were dropped by `tmb_data` but called again in `h_mmrm_tmb_fit` with the dropped rows. This mismatch was the cause of the buggy behavior.

Fixes #59


danielinteractive:
Thanks @Melkiades for working on this - can you please first create a reprex and add it as a unit test in this PR as well? (see checklist in issue)
second I am not sure whether the current fix will be the right one since this changes the semantic of what `data` is in the fit result. but let's look at that again with the reprex unit test in place

Melkiades:
> Thanks @Melkiades for working on this - can you please first create a reprex and add it as a unit test in this PR as well? (see checklist in issue) second I am not sure whether the current fix will be the right one since this changes the semantic of what `data` is in the fit result. but let's look at that again with the reprex unit test in place

The last commit provides unit testing to see the error. I decomposed the function call into parts to be able to imitate the same situation

danielinteractive (approved):
thanks again @Melkiades , I did a few changes to preserve meaning of the `data` element in `mmrm_tmb` result

PR #68, opened 2022-07-07
65: add cs

### LINKED ISSUE #65: Add compound symmetry covariance structure
To do:

- [ ] Add definition of compound symmetry covariance structure to covariance vignette 
- [ ] Adjust covariance structure choices in https://github.com/openpharma/mmrm/blob/709f392e966b12514525ea6f6118d8e9af32089c/R/tmb.R#L60
- [ ] Amend codes in https://github.com/openpharma/mmrm/blob/709f392e966b12514525ea6f6118d8e9af32089c/R/tmb.R#L157
- [ ] Initialize start parameters in https://github.com/openpharma/mmrm/blob/709f392e966b12514525ea6f6118d8e9af32089c/R/tmb.R#L208
	- [ ] Amend tests in https://github.com/openpharma/mmrm/blob/709f392e966b12514525ea6f6118d8e9af32089c/tests/testthat/test-tmb.R#L80
- [ ] Amend codes in https://github.com/openpharma/mmrm/blob/709f392e966b12514525ea6f6118d8e9af32089c/src/correlation.h#L23
- [ ] Amend printing of covariance structure in https://github.com/openpharma/mmrm/blob/b1b612263623c9d8f9e972adf19d09e23558664c/R/mmrm-methods.R#L109
- [ ] Add lower Cholesky factor function and include it in branching in 
 https://github.com/openpharma/mmrm/blob/709f392e966b12514525ea6f6118d8e9af32089c/src/correlation.h#L33
	- [ ] Test it in https://github.com/openpharma/mmrm/blob/main/src/test-correlation.cpp
- [ ] Amend integration tests in https://github.com/openpharma/mmrm/blob/709f392e966b12514525ea6f6118d8e9af32089c/tests/testthat/test-tmb.R#L256
- [ ] Run Clean and Rebuild > Check to ensure that checks pass

danielinteractive:
closes #65

### DIFF
--- a/tests/testthat/test-cpp-covariance.R
+++ b/tests/testthat/test-cpp-covariance.R
@@ -1,4 +1,4 @@
-# get_ante_dependence ----
+# ante-dependence ----
 
 test_that("prepare numbers for ante_dependence tests", {
   theta <- c(log(1), log(2), log(3), 1, 2)
@@ -26,7 +26,7 @@ test_that("prepare numbers for ante_dependence tests", {
   expect_equal(result, expected, tolerance = 1e-5)
 })
 
-# get_toeplitz ----
+# toeplitz ----
 
 test_that("prepare numbers for toeplitz tests", {
   theta <- c(log(1), log(2), log(3), 1, 2)
@@ -102,12 +102,12 @@ test_that("prepare numbers for heterogeneous autoregressive tests", {
   expect_equal(result, expected, tolerance = 1e-5)
 })
 
-# compound symemtry ----
+# compound symmetry ----
 
 test_that("prepare numbers for compound symmetry correlation function test", {
   x <- 1.2
   cor <- map_to_cor(x)
-  expect_equal(cor, 0.7682213)
+  expect_equal(cor, 0.7682213, tolerance = 1e-4)
 })
 
 test_that("prepare numbers for homogeneous compound symmetry tests", {


R/tmb.R
```diff
@@ -171,7 +171,9 @@ h_mmrm_tmb_data <- function(formula_parts,
     toep = 2,
     ar1 = 3,
     ar1h = 4,
-    ad = 5
+    ad = 5,
+    cs = 6,
+    csh = 7
```
dgkf, line 176: Still digesting the code base, but this stood out to me as a curious design choice. If I'm understanding this correctly, this maps the covariance types to their enum counterparts, but I would expect this to happen closer to wherever the C interface is called instead of losing the information at this point in the processing. As it is now, it would be rather difficult to recover this information.

src/test-covariance.cpp
```diff
@@ -91,6 +91,38 @@ context("autoregressive") {
   }
 }
 
+context("compound symmetry") {
+  test_that("corr_fun_compound_symmetry works as expected") {
+    vector<double> theta {{1.0}};
+    corr_fun_compound_symmetry<double> test_fun(theta);
+    expect_equal(test_fun(1, 0), 1 / sqrt(2));
+    expect_equal(test_fun(4, 1), 1 / sqrt(2));
+    expect_equal(test_fun(3, 1), 1 / sqrt(2));
```
dgkf, line 100: Using a theta of 1 doesn't test for whether the correlation parameter is squared (eg, with `theta == 1`, `theta / sqrt(1 + theta * theta)` == `theta / sqrt(1 + theta)`. 

Probably unlikely to be problematic, but it would be more thorough to use something other than the multiplicative identity for this.

dgkf (approved):
Mostly just using the review as an opportunity to understand the code base. Minor comments and suggestions, and for what it's worth I spot checked the SAS-produced results against the ones used for testing.

As far as my newcomer impressions, I'd say it's approved and changes would be up to you. Of course, would totally understand wanting further review from someone who's more familiar.

R/tmb.R
danielinteractive: Good catch! I know that this is not pretty. Thanks for commenting on it, I will make a separate issue for this as it will require some medium refactoring.

danielinteractive: https://github.com/openpharma/mmrm/issues/69

src/test-covariance.cpp
danielinteractive: Good idea, thanks, done

PR #70, opened 2022-07-08
Handle mapping of covariance type abbreviations to functions within Cpp-level API (Resolves #69)

### LINKED ISSUE #69: Avoid or improve `enum` mapping
Motivation: Currently we are mapping the covariance structures to their C++ `enum` integer codes here https://github.com/openpharma/mmrm/blob/b1b612263623c9d8f9e972adf19d09e23558664c/R/tmb.R#L169

Ideal outcome: Can we avoid this completely and just pass the `cov_type` string (e.g. `ar1`) directly to C++ and then decide based on that?

Otherwise: Somehow do the `enum` conversion in a more robust way.

dgkf:
Closes #69

# Description

Previously, `h_mmrm_tmb_data` would store a numeric value which corresponds to a covariance function. This is a somewhat superficial change that provides the covariance type to the model function generator as a string, allowing it to be more meaningful without knowledge of the enum mapping. 

I also added an error condition when the string was unmatched. Previously, an unmatched enum would silently return an empty matrix. This was somewhat guarded against by verifying the condition type in `h_mmrm_tmb_formula_parts`, but perhaps this helps to focus that checking in one place and reduce the need to keep parity between the two bits of code? Of course, if a user would ever need to tease apart `formula_parts` without building a model function, then it's nice to keep the guard in both places.

---

As an aside, this PR took me an embarrassingly long time to put together, so thanks for your patience as I re-familiarize myself with Cpp. I had a ton of issues trying to debug some of the templated code, which seems to give rather uninformative error messages. Any tips on how to better debug this?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,14 +1,14 @@
 Type: Package
 Package: mmrm
 Title: Mixed Models for Repeated Measures
-Version: 0.0.0.9011
+Version: 0.0.0.9010
 Authors@R: c(
     person("Daniel", "Sabanes Bove", , "daniel.sabanes_bove@roche.com", role = c("aut", "cre")),
     person("Julia", "Dedic", , "julia.dedic@roche.com", role = "aut"),
     person("Brian Matthew", "Lang", , "brian.lang@msd.com", role = "aut"),
+    person("Doug", "Kelkhoff", , "kelkhoff.douglas@gene.com", role = "aut"),
     person("Craig", "Gower-Page", , "craig.gower-page@roche.com", role = "aut"),
-    person("Alessandro", "Noci", , "alessandro.noci@roche.com", role = "aut"),
-    person("Doug", "Kelkhoff", , "kelkhoff.douglas@gene.com", role = "aut")
+    person("Alessandro", "Noci", , "alessandro.noci@roche.com", role = "aut")
   )
 Description: Mixed models for repeated measures (MMRM) are a popular
     choice for individually randomized trials with longitudinal continuous
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,13 +1,14 @@
 Type: Package
 Package: mmrm
 Title: Mixed Models for Repeated Measures
-Version: 0.0.0.9010
+Version: 0.0.0.9011
 Authors@R: c(
     person("Daniel", "Sabanes Bove", , "daniel.sabanes_bove@roche.com", role = c("aut", "cre")),
     person("Julia", "Dedic", , "julia.dedic@roche.com", role = "aut"),
     person("Brian Matthew", "Lang", , "brian.lang@msd.com", role = "aut"),
     person("Craig", "Gower-Page", , "craig.gower-page@roche.com", role = "aut"),
-    person("Alessandro", "Noci", , "alessandro.noci@roche.com", role = "aut")
+    person("Alessandro", "Noci", , "alessandro.noci@roche.com", role = "aut"),
+    person("Doug", "Kelkhoff", , "kelkhoff.douglas@gene.com", role = "aut")
   )
 Description: Mixed models for repeated measures (MMRM) are a popular
     choice for individually randomized trials with longitudinal continuous
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,3 @@
-# mmrm 0.0.0.9010
+# mmrm 0.0.0.9011
 
 Initialized package.
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,3 @@
-# mmrm 0.0.0.9011
+# mmrm 0.0.0.9010
 
 Initialized package.
--- a/R/tmb.R
+++ b/R/tmb.R
@@ -121,7 +121,7 @@ h_mmrm_tmb_formula_parts <- function(formula) {
 #'     indices for each subject.
 #' - `subject_n_visits`: length `n_subjects` `integer` containing the number of
 #'     observed visits for each subjects. So the sum of this vector equals `n`.
-#' - `cov_type`: `int` specifying the covariance type.
+#' - `cov_type`: `chr` value specifying the covariance type.
 #' - `reml`: `int` specifying whether REML estimation is used (1), otherwise ML (0).
 #'
 #' @details Note that the `subject_var` must not be factor but can also be character.
@@ -165,7 +165,6 @@ h_mmrm_tmb_data <- function(formula_parts,
   subject_zero_inds <- which(!duplicated(full_frame[[formula_parts$subject_var]])) - 1L
   subject_n_visits <- c(utils::tail(subject_zero_inds, -1L), nrow(full_frame)) - subject_zero_inds
   assert_true(identical(subject_n_visits, as.integer(table(full_frame[[formula_parts$subject_var]]))))
-  cov_type <- formula_parts$cov_type
 
   structure(
     list(
@@ -177,7 +176,7 @@ h_mmrm_tmb_data <- function(formula_parts,
       n_subjects = n_subjects,
       subject_zero_inds = subject_zero_inds,
       subject_n_visits = subject_n_visits,
-      cov_type = cov_type,
+      cov_type = formula_parts$cov_type,
       reml = as.integer(reml)
     ),
     class = "mmrm_tmb_data"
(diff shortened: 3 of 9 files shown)


DESCRIPTION
```diff
@@ -1,13 +1,14 @@
 Type: Package
 Package: mmrm
 Title: Mixed Models for Repeated Measures
-Version: 0.0.0.9010
```
danielinteractive, line 4: please leave this as is, we will get automatic version bump from automation

DESCRIPTION
```diff
@@ -1,13 +1,14 @@
...
+Version: 0.0.0.9011
 Authors@R: c(
     person("Daniel", "Sabanes Bove", , "daniel.sabanes_bove@roche.com", role = c("aut", "cre")),
     person("Julia", "Dedic", , "julia.dedic@roche.com", role = "aut"),
     person("Brian Matthew", "Lang", , "brian.lang@msd.com", role = "aut"),
     person("Craig", "Gower-Page", , "craig.gower-page@roche.com", role = "aut"),
-    person("Alessandro", "Noci", , "alessandro.noci@roche.com", role = "aut")
+    person("Alessandro", "Noci", , "alessandro.noci@roche.com", role = "aut"),
+    person("Doug", "Kelkhoff", , "kelkhoff.douglas@gene.com", role = "aut")
```
danielinteractive, line 11: I would suggest to move your name between Brian and Craig, since Craig and Alessandro are here mostly because we copied some function from rbmi, but they are not direct authors with new contributions here

NEWS.md
```diff
@@ -1,3 +1,3 @@
-# mmrm 0.0.0.9010
```
danielinteractive, line 1: also here please leave as is

R/tmb.R
```diff
@@ -165,16 +165,7 @@ h_mmrm_tmb_data <- function(formula_parts,
...
-  cov_type <- as.integer(switch(formula_parts$cov_type,
-    us = 1,
-    toep = 2,
-    ar1 = 3,
-    ar1h = 4,
-    ad = 5,
-    cs = 6,
-    csh = 7
-  ))
+  cov_type <- formula_parts$cov_type
```
danielinteractive, line 168: since it is now so simple suggest to skip this here and instead do it in line 180 as part of the list assembly

danielinteractive: please also adjust return value description in line 124

src/covariance.h
```diff
@@ -111,44 +111,29 @@ matrix<T> get_compound_symmetry_heterogeneous(const vector<T>& theta, int n_visi
...
-  case compound_symmetry_cov:
+  } else if (cov_type == "cs") {
     result = get_compound_symmetry<T>(theta, n_visits);
-    break;
-  case compound_symmetry_heterogeneous_cov:
+  } else if (cov_type == "csh") {
     result = get_compound_symmetry_heterogeneous<T>(theta, n_visits);
-    break;
+  } else {
+    Rf_error(("Unknown covariance type '" + cov_type + "'.").c_str());
```
danielinteractive, line 134: I wonder if you can do without the `Rf_` prefix, see e.g. https://github.com/glmmTMB/glmmTMB/blob/bb79005420e100244182bbe144040af2de67ec41/glmmTMB/src/glmmTMB.cpp#L96 for comparison

src/covariance.h
```diff
@@ -111,44 +111,29 @@ matrix<T> get_compound_symmetry_heterogeneous(const vector<T>& theta, int n_visi
...
-  auto_regressive_heterogeneous_cov = 4,
-  ante_dependence_cov = 5,
-  compound_symmetry_cov = 6,
-  compound_symmetry_heterogeneous_cov = 7
-};
-
 // Creates a new correlation object dynamically.
 template <class T>
-matrix<T> get_covariance_lower_chol(const vector<T>& theta, int n_visits, int cov_type) {
+matrix<T> get_covariance_lower_chol(const vector<T>& theta, int n_visits, const std::string& cov_type) {
```
danielinteractive, line 116: I  wonder if it was a bit more flexible re: tests here to use a copy in the signature rather than a const reference:
```suggestion
matrix<T> get_covariance_lower_chol(const vector<T>& theta, int n_visits, std::string cov_type) {
```

src/test-covariance.cpp
```diff
@@ -126,7 +126,7 @@ context("compound symmetry") {
 context("get_covariance_lower_chol") {
   test_that("get_covariance_lower_chol gives expected unstructured result") {
     vector<double> theta {{log(1.0), log(2.0), 3.0}};
-    matrix<double> result = get_covariance_lower_chol(theta, 2, unstructured_cov);
+    matrix<double> result = get_covariance_lower_chol(theta, 2, "us");
```
danielinteractive, line 129: like here I would have thought we need to make a copy of the string, I remember I had issues before when I tried to pass an int instead of int variable to a const-ref int argument... but probably my C++ is not good enough to understand the difference :-D

tests/testthat/test-tmb.R
```diff
@@ -335,6 +335,25 @@ test_that("h_mmrm_tmb_fit works as expected", {
...
+test_that("h_mmrm_tmb_fit errors when an invalid covariance type is used", {
+  formula <- FEV1 ~ RACE + us(AVISIT | USUBJID)
+  formula_parts <- h_mmrm_tmb_formula_parts(formula)
+
+  tmb_data <- h_mmrm_tmb_data(formula_parts, fev_data, reml = FALSE)
+  tmb_parameters <- h_mmrm_tmb_parameters(formula_parts, tmb_data, start = NULL)
+
+  tmb_data$cov_type <- "gaaah"
+  expect_error(regex = c("Unknown covariance type 'gaaah'"), {
+    tmb_object <- TMB::MakeADFun(
```
danielinteractive, line 347:
```suggestion
    TMB::MakeADFun(
```

tests/testthat/test-tmb.R
```diff
@@ -335,6 +335,25 @@ test_that("h_mmrm_tmb_fit works as expected", {
...
 
+test_that("h_mmrm_tmb_fit errors when an invalid covariance type is used", {
+  formula <- FEV1 ~ RACE + us(AVISIT | USUBJID)
+  formula_parts <- h_mmrm_tmb_formula_parts(formula)
+
+  tmb_data <- h_mmrm_tmb_data(formula_parts, fev_data, reml = FALSE)
+  tmb_parameters <- h_mmrm_tmb_parameters(formula_parts, tmb_data, start = NULL)
+
+  tmb_data$cov_type <- "gaaah"
+  expect_error(regex = c("Unknown covariance type 'gaaah'"), {
```
danielinteractive, line 346: suggest to use same pattern as in line 213 ff. e.g. for consistency

danielinteractive (commented):
Thanks a lot Doug! Very nice. Please see few minor comments inside. Re: debugging the TMB stuff here - yeah the templates make it harder. I don't have really good tips for that. The TMB reference does also not say anything really: https://kaskr.github.io/adcomp/_book/Errors.html
Run time errors can be done with gdb etc (which is also painful) but compilation is hard. Sometimes

src/covariance.h
dgkf: Let me know if you'd prefer that I change this. Initially I was just copying whatever consensus seemed to surface when I was searching around for best practices. With your comment, I tried to search for something more official and came across [this recommendation](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f16-for-in-parameters-pass-cheaply-copied-types-by-value-and-others-by-reference-to-const) from the iso cpp guidelines:

> For “in” parameters, pass cheaply-copied types by value and others by reference to `const`

Then the distinction is whether covariance type indicators are "cheaply-copied". They use `int`s as examples of cheaply copied, and `string`s as examples of "expensive" copies, but presumably they're considering strings that are longer than 4 characters. 

In short, it _seems_ like this is unlikely to be problematic and aligns with the recommended practice, but I'd be fine to change it if you'd prefer.

danielinteractive: thanks Doug, yeah I vaguely remember this kind of rule, I would say in this case it is still cheap so let's change

danielinteractive (approved):
thx Doug! after minor modification this will be good to merge. so I approve already now so that you don't need to wait for me.

R/tmb.R
```diff
@@ -121,7 +121,7 @@ h_mmrm_tmb_formula_parts <- function(formula) {
 #'     indices for each subject.
 #' - `subject_n_visits`: length `n_subjects` `integer` containing the number of
 #'     observed visits for each subjects. So the sum of this vector equals `n`.
-#' - `cov_type`: `int` specifying the covariance type.
+#' - `cov_type`: `chr` value specifying the covariance type.
```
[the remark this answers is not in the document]
dgkf: Ah gotcha - first time seeing this. Thanks for the heads up.

PR #78, opened 2022-07-25
64 clean up exports

BrianLang:
First Draft
Cleaned up exports.
Added a single test for single_optimizer which seemed to maybe not have been covered?
Need input into how we choose badges. 

# Description


BrianLang:
Any idea why the linter is failing for snake case on the emmeans.R functions? I didn't touch change anything that I would expect to affect this...

R/emmeans.R
```diff
@@ -15,20 +15,10 @@
 
 #' Support for `emmeans`
 #'
-#' @description `r lifecycle::badge("experimental")`
```
danielinteractive, line 18: I would keep this and below example

NAMESPACE
```diff
@@ -19,29 +19,12 @@ export(df_1d)
 export(df_md)
 export(fit_single_optimizer)
 export(h_coef_table)
```
danielinteractive, line 21: also remove this from exports

NAMESPACE
```diff
@@ -19,29 +19,12 @@ export(df_1d)
...
-export(h_mmrm_tmb_assert_opt)
-export(h_mmrm_tmb_assert_start)
 export(h_mmrm_tmb_control)
-export(h_mmrm_tmb_data)
-export(h_mmrm_tmb_fit)
-export(h_mmrm_tmb_formula_parts)
-export(h_mmrm_tmb_parameters)
 export(h_print_aic_list)
 export(h_print_call)
 export(h_print_cov)
```
danielinteractive, line 27: also remove these from exports

danielinteractive: (print helpers)

R/tmb.R
```diff
@@ -415,7 +392,19 @@ h_mmrm_tmb_fit <- function(tmb_object,
 #' @param control (`mmrm_tmb_control`)\cr list of control options produced
 #'   by [h_mmrm_tmb_control()].
 #'
-#' @return List of class `mmrm_tmb`, see [h_mmrm_tmb_fit()] for details.
+#' @return List pf class `mmrm_tmb` with:
```
danielinteractive, line 395:
```suggestion
#' @return List of class `mmrm_tmb` with:
```

R/tmb.R
```diff
@@ -415,7 +392,19 @@ h_mmrm_tmb_fit <- function(tmb_object,
 #' @param control (`mmrm_tmb_control`)\cr list of control options produced
 #'   by [h_mmrm_tmb_control()].
 #'
-#' @return List of class `mmrm_tmb`, see [h_mmrm_tmb_fit()] for details.
```
danielinteractive, line 418: hm but then we have duplicated documentation... I think once a user is here, then it is ok to forward them to internal helper function documentation. so I would keep this as it was.

_pkgdown.yml
```diff
@@ -19,37 +19,20 @@ reference:
...
-      - h_quad_form
-      - h_gradient
-      - h_df_1d_list
-      - h_md_denom_df
-      - h_df_md_list
-      - h_df_md_from_1d
       - h_coef_table
       - h_print_aic_list
       - h_print_call
       - h_print_cov
```
danielinteractive, line 27:
```suggestion
```
(in line with my comment above on not exporting these)

danielinteractive (commented):
Thanks a lot Brian! Please see few comments inside

danielinteractive (approved):

R/emmeans.R
BrianLang: thanks

PR #88, opened 2022-08-10
84: set `weights` to obtain weighted MMRM

### LINKED ISSUE #84: Add possibility to set `weights` to obtain weighted MMRM
To do:

- [x] Amend algorithm vignette to understand how the weights would be included in the fitting process
- [x] Add argument to `mmrm()` and related functions (to be detailed here)
	- [x] Add argument to `mmrm()` and `fit_single_optimizer()` in file `R/fit.R`
    - [x] Add argument to `h_mmrm_tmb` and `h_mmrm_tmb_data` in file `R/tmb.R`
    - [x] Update file `/src/mmrm.cpp`
- [x] For lower level functions: Add unit tests for the weights functionality for all functions which are involved
- [x] For `mmrm()`: Add integration test where we compare results with SAS results
	- [x] seems ok for weights = 1 (no weights)
	- [x] resolve differences for random weights
	- [x] including d.f.
- [x] For `print()` method: It would be good if we could include a line that mentions the weights there
	- [x] e.g. in here: https://github.com/openpharma/mmrm/blob/97373f29793a4b04563b91a91f3c4482807a9c6d/R/tmb-methods.R#L185
	- [x] can maybe do same trick as here: https://github.com/openpharma/mmrm/blob/97373f29793a4b04563b91a91f3c4482807a9c6d/R/fit.R#L220
- [x] Document in the introduction vignette

ywang-gilead:
closes #84


vignettes/algorithm.Rmd
```diff
@@ -77,7 +77,7 @@ S_i = \begin{pmatrix}
 Note that this follows from the well known property of the multivariate normal
 distribution that linear combinations of the random vector again have a
 multivariate normal distribution with the correspondingly modified mean
-vector and covariance matrix.
+vector and covariance matrix. When weights are specified, we replace the covariance matrix $\Sigma_i$ with $Z_i^{-1/2} \Sigma_i Z_i^{-1/2}$, where $Z_i$ is the diagonal weight matrix.
```
danielinteractive, line 80: maybe it would be easier to read and explain when we always use this more general form? i.e. already in line 60 define $\Sigma_i$ accordingly and introduce what $Z_i$ is. If no weights then it is just the identity matrix.

vignettes/algorithm.Rmd
```diff
@@ -188,7 +188,7 @@ and similarly
 X^\top \Omega^{-1} Y = X^\top W Y =  \sum_{i=1}^{n} X_i^\top W_i Y_i
 \]
 where $W_i = \Sigma_i^{-1}$ is the weight matrix for subject $i$, the inverse
-of its covariance matrix.
+of its covariance matrix. When the diagonal weight matrix $Z_i$ is specified, we replace $W_i$ with $W_i = (Z_i^{-1/2} \Sigma_i Z_i^{-1/2})^{-1} = Z_i^{1/2} \Sigma_i^{-1} Z_i^{1/2}$.
```
danielinteractive, line 191: I think in any case we don't need this here since we refer to $\Sigma_i$
```suggestion
of its covariance matrix.
```

vignettes/algorithm.Rmd
```diff
@@ -218,7 +218,7 @@ and we can thereby calculate the left and right hand sides for the WLS estimatin
 equation. We solve this equation with a robust Cholesky decomposition with pivoting.
 The advantage is that we can reuse this decomposition for calculating the
 covariance matrix of $\hat{\beta}$, i.e. $K = (X^\top W X)^{-1}$, by supplying the
-identity matrix as alternative right hand side.
+identity matrix as alternative right hand side. When weights $Z_i$ is specified, we replace $L_i$ with $Z_i^{-1/2} L_i$ in the above calculation.
```
danielinteractive, line 221: Also here I don't think we need this since in line 195 we refer to $\Sigma_i$
```suggestion
identity matrix as alternative right hand side.
```

danielinteractive: We could instead describe additional details how $L_i$ is calculated but not sure if needed?

vignettes/algorithm.Rmd
```diff
@@ -238,6 +238,7 @@ where $l_{i,jj}$ are the diagonal entries of the factor $L_i$ and we have used t
 - the determinant of the product of matrices is the product of the determinants,
 - the determinant of the transposed matrix is the same as the original one,
 - the determinant of a triangular matrix is the product of the diagonal.
+When weights $Z_i$ is specified, we have $\frac{1}{2}\log\det(\Omega) = \sum_{i=1}^{n}\sum_{j=1}^{m_i}\log(z_{i, j} l_{i, jj})$, where $z_{i, j}$ are the diagonal entries of the weight $Z_i$.
```
danielinteractive, line 241: similar here - not sure if we need this? e.g. if we explain above already how to calculate $L_i$ then this contains already everything needed here?

danielinteractive (commented):
thanks a lot Ya for the vignette start - I added some comments to simplify this further, please have a look whether it makes sense to you. basically I am thinking if we define the model including the weights matrix from the beginning it should be easier in the sense that we don't need to modify many formulas/paragraphs

danielinteractive (commented):
Vignette Looks good like this! So you can proceed to coding

danielinteractive:
@ywang-gilead looks good! very nice and simple. So next is the tests right?

danielinteractive (commented):
Hi @ywang-gilead , I just had an idea what might well be the problem: the weights vector is not necessarily aligned with the design matrix etc. if there are any rows with missing values in the required data frame columns then those will be omitted. However I see so far the weights vector would not take that into account. I would suggest :
1) check how other packages are aligning the weights vector if it comes from a separate vector. If that is easy let's do that: in `h_mmrm_tmb_data()`:
a) pass it to `model.frame()`
b) extract with `model.weights()`

2) otherwise it might still be worth considering to pass the weights specification diffently. E.g. one could imagine to have it in the same model formula with a `weights()` syntax giving the weights column to use from the data frame. 
`CHG ~ SEX + RACE + us(VISIT | SUBJID) + weights(HEIGHT)`

danielinteractive:
maybe we need to do something like this:

```r
m <- match(c("formula", "data", "subset", "weights", "na.action", 
        "offset"), names(mf), 0L)
    mf <- mf[c(1L, m)]
    mf$drop.unused.levels <- TRUE
    mf[[1L]] <- quote(stats::model.frame)
    mf <- eval(mf, parent.frame())
```

hopefully we don't need this

danielinteractive:
Thanks a lot @clarkliming for the tip! Indeed that would work. I wonder if instead of throwing away the environment it would be an alternative to assign the weights in that environment. @ywang-gilead see the commit I just pushed, this also works. so you can proceed from here as planned I think

R/tmb.R
```diff
@@ -166,7 +166,9 @@ h_mmrm_tmb_data <- function(formula_parts,
     )
   }
 
+  data$order_id <- c(1:nrow(data))
   data <- data[order(data[[formula_parts$subject_var]], data[[formula_parts$visit_var]]), ]
+  weights <- weights[data$order_id]
```
danielinteractive, line 171:
```suggestion
  data_order <- order(data[[formula_parts$subject_var]], data[[formula_parts$visit_var]])
  data <- data[data_order, ]
  weights <- weights[data_order]
```

R/fit.R
```diff
@@ -219,9 +227,18 @@ mmrm <- function(formula,
 
   attr(data, which = "dataname") <- toString(match.call()$data)
 
+  nobs <- nrow(data)
+  if (is.null(weights)){
+    weights <- rep(1, nobs)
+  }
+  if(!is.null(weights) && any(weights <= 0)){
+    stop("only positive weights are allowed")
+  }
```
danielinteractive, line 236: I would assert this in lower level function e.g. `h_mmrm_tmb()`, so not needed here
```suggestion
```

R/tmb.R
```diff
@@ -426,11 +442,13 @@ h_mmrm_tmb_fit <- function(tmb_object,
 #' system.time(result <- h_mmrm_tmb(formula, data))
 h_mmrm_tmb <- function(formula,
                        data,
+                       weights,
                        reml = TRUE,
                        control = h_mmrm_tmb_control()) {
   formula_parts <- h_mmrm_tmb_formula_parts(formula)
   assert_class(control, "mmrm_tmb_control")
-  tmb_data <- h_mmrm_tmb_data(formula_parts, data, reml, accept_singular = control$accept_singular)
+  assert_vector(weights)
```
danielinteractive, line 450: I would assert this here 
```suggestion
  assert_numeric(weights, any.missing = FALSE)
  assert_true(all(weights > 0))
```

R/tmb.R
```diff
@@ -158,10 +160,12 @@ h_mmrm_tmb_formula_parts <- function(formula) {
 #' @keywords internal
 h_mmrm_tmb_data <- function(formula_parts,
                             data,
+                            weights,
                             reml,
                             accept_singular) {
   assert_class(formula_parts, "mmrm_tmb_formula_parts")
   assert_data_frame(data)
+  assert_vector(weights)
```
danielinteractive, line 168:
```suggestion
  assert_numeric(weights, len = nrow(data))
```

design/SAS/sas_weighted_mmrm.sas
```diff
@@ -0,0 +1,28 @@
+* Weighted MMRM ----;
+
+** ML ----;
+
+ODS RTF FILE="&datpath\sas_weighted_mmrm_ml.rtf";
```
danielinteractive, line 5: thanks @ywang-gilead , one minor ask, can you please save the SAS results as txt files? because then it is much easier to view them during code review etc. thank you!

tests/testthat/test-tmb-methods.R
```diff
@@ -48,7 +48,7 @@ test_that("model.frame returns full model frame if requested", {
   object <- get_mmrm_tmb()
   result <- expect_silent(model.frame(object, full = TRUE))
   expect_data_frame(result, nrows = length(object$tmb_data$y_vector))
-  expect_named(result, c("FEV1", "RACE", "USUBJID", "AVISIT"))
+  expect_named(result, c("FEV1", "RACE", "USUBJID", "AVISIT", "(weights)"))
```
danielinteractive, line 51: that is good, can you please adapt the documentation accordingly? (explanation for function argument `full` to include mentioning of weights)

danielinteractive (commented):
thanks a lot @ywang-gilead ! that is great. I think only last step remaining is to add a subsection "Weighting" or so under "Common customizations" and then you are done!

danielinteractive (commented):
sorry one more thing: For the `print()` method - It would be good if we could include a line that mentions the weights there, so that the user is made aware of the fact that weights were used. If no weights were used can have "Weights: None" otherwise use e.g. "Weights: fev_data$weights" if the function was called like that

design/SAS/sas_weighted_mmrm.sas
chstock: Thanks for considering the implementation of the weight statement in PROC MIXED! Just as a remark, not sure it should become an issue, at least not a pressing one: I was also thinking about use cases for weights that are sampling weights, more in the sense of non-random survey samples, to model e.g. channeling into treatments when using RWD. This is implemented in SAS survey procedures, e.g. PROC SURVEYREG, and also in PROC GLIMMIX ("obsweight" in MODEL statement and "weight" in RANDOM statement, to weight observations and subjects (higher levels), respectively). 
See e.g. https://documentation.sas.com/doc/en/pgmsascdc/9.4_3.3/statug/statug_glimmix_details25.htm . Of note perhaps, PROC GLIMMIX does this with G-side random effects only, not R-side random effects.

danielinteractive: Thanks @chstock - let me move this to a separate issue and then we can discuss further there / in a dedicated meeting.

R/fit.R
```diff
@@ -230,9 +230,9 @@ mmrm <- function(formula,
   nobs <- nrow(data)
   if (is.null(weights)){
     weights <- rep(1, nobs)
-  }
-  if(!is.null(weights) && any(weights <= 0)){
-    stop("only positive weights are allowed")
+    attr(weights, which = "dataname") <- "None"
```
danielinteractive, line 233: do we even need this then? because I see from the print code below that if the weights were not supplied by the user in the first place then it is not even printed at all

R/tmb-methods.R
```diff
@@ -40,7 +40,7 @@ fitted.mmrm_tmb <- function(object, ...) {
 }
 
 #' @describeIn mmrm_tmb_methods obtains the model frame.
-#' @param full (`flag`)\cr whether to include subject and visit variable.
+#' @param full (`flag`)\cr whether to include subject, visit and weight variable.
```
danielinteractive, line 43:
```suggestion
#' @param full (`flag`)\cr whether to include subject, visit and weight variables.
```

R/tmb.R
```diff
@@ -484,6 +485,11 @@ h_mmrm_tmb <- function(formula,
     attr(data, which = "dataname"),
     toString(match.call()$data)
   )
+  fun_call$weights <- ifelse(
+    !is.null(attr(weights, which = "dataname")),
+    attr(weights, which = "dataname"),
+    deparse(match.call()$weights)
+  )
```
danielinteractive, line 492:
```suggestion
  weights_str <- attr(weights, which = "dataname")
  if (!is.null(weights_str)) {
    fun_call$weights <- weights_str
  }
```
because if it is NULL then we don't need to add it

danielinteractive (commented):
thanks a lot @ywang-gilead ! looks great. can you please add a snapshot test for the new print behavior? (currently I don't see a change there) then this should be finished!

R/fit.R
ywang-gilead: current code prints `Weights:     None` when weights is not specified, otherwise we will have `Weights:         `.

R/tmb.R
ywang-gilead: I changed it back for now as otherwise got 2 failed tests:
image

R/fit.R
danielinteractive: hm I am looking at it now, because currently we get 
`h_mmrm_tmb(formula = FEV1 ~ RACE + SEX + ARMCD * AVISIT + us(AVISIT | 
    USUBJID), data = "fev_data", weights = "None", reml = reml, 
    control = control)`
as the `fit$call` result which is misleading

R/tmb.R
danielinteractive: ok I changed the condition for the printing slightly (checking for `symbol` instead of not `NULL`) and then it works

danielinteractive (approved):

PR #96, opened 2022-08-31
95: fix typo and update covariance structure check

### LINKED ISSUE #95: Fix typo, improve error message for using wrong covariance structure string
Problem:
image

To do:

- [x] Search for `arh1` and replace with `ar1h` across the package
- [x] Improve error message for  using wrong covariance structure string
- [x] add test for this

clarkliming:
close #95 

the typo in [vignettes/covariance.Rmd](https://github.com/openpharma/mmrm/compare/95_typo?expand=1#diff-1a86968857585f8d0d042c84415044f4dd62dd4757fdd8175abc4c6a8e7e5ba0) file is updated

now there is a covariance structure checking include

1. no covariance matrix specified
2. covariance matrix specified multiple times
3. covariance matrix does not take the form  `cs(VISIT|ID)`

### DIFF
--- a/R/tmb.R
+++ b/R/tmb.R
@@ -95,7 +95,7 @@ h_mmrm_tmb_formula_parts <- function(formula) {
     stop(
       "Covariance structure must be of the form `",
       cov_term[[1]],
-      "(VISIT|ID)`"
+      "(time|subject)`"
     )
   }
   visit_term <- cov_content[[2L]]
--- a/R/tmb.R
+++ b/R/tmb.R
@@ -68,7 +68,7 @@ h_mmrm_tmb_formula_parts <- function(formula) {
   if (length(cov_selected) == 0) {
     stop(
       "Covariance structure must be specified in formula. ",
-      "Possible covariance structure include: ",
+      "Possible covariance structures include: ",
       paste0(cov_functions, collapse = ", ")
     )
   }
--- a/R/tmb.R
+++ b/R/tmb.R
@@ -64,11 +64,23 @@ h_mmrm_tmb_formula_parts <- function(formula) {
   cov_functions <- c("us", "toep", "toeph", "ar1", "ar1h", "ad", "adh", "cs", "csh")
   terms_object <- stats::terms(formula, specials = cov_functions)
   found_specials <- attr(terms_object, "specials")
-  cov_selected <- !sapply(found_specials, is.null)
-  assert_true(identical(sum(cov_selected), 1L))
-  cov_index <- found_specials[[which(cov_selected)]] + 1L # Subtract 1 for `list()`.
+  cov_selected <- Filter(Negate(is.null), found_specials)
+  if (length(cov_selected) == 0) {
+    stop(
+      "Covariance matrix must be specified in formula. ",
+      "Possible covariance matrix include: ",
+      paste0(cov_functions, collapse = ", ")
+    )
+  }
+  if (length(cov_selected) > 1) {
+    stop(
+      "Only one covariance matrix can be specified. ",
+      "Currently specified covariance matrix is:",
+      paste0(names(cov_selected), collapse = ", ")
+    )
+  }
   terms_list <- attr(terms_object, "variables")
-  cov_term <- terms_list[[cov_index]]
+  cov_term <- terms_list[[unlist(cov_selected) + 1L]]
 
   # Remove the covariance term to obtain the model formula.
   model_formula <- stats::update(
--- a/R/tmb.R
+++ b/R/tmb.R
@@ -69,14 +69,14 @@ h_mmrm_tmb_formula_parts <- function(formula) {
     stop(
       "Covariance structure must be specified in formula. ",
       "Possible covariance structures include: ",
-      paste0(cov_functions, collapse = ", ")
+      toString(cov_functions)
     )
   }
   if (length(cov_selected) > 1) {
     stop(
       "Only one covariance structure can be specified. ",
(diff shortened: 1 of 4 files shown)


R/tmb.R
```diff
@@ -64,11 +64,23 @@ h_mmrm_tmb_formula_parts <- function(formula) {
...
   terms_object <- stats::terms(formula, specials = cov_functions)
   found_specials <- attr(terms_object, "specials")
-  cov_selected <- !sapply(found_specials, is.null)
-  assert_true(identical(sum(cov_selected), 1L))
-  cov_index <- found_specials[[which(cov_selected)]] + 1L # Subtract 1 for `list()`.
+  cov_selected <- Filter(Negate(is.null), found_specials)
+  if (length(cov_selected) == 0) {
+    stop(
+      "Covariance structure must be specified in formula. ",
+      "Possible covariance structure include: ",
```
danielinteractive, line 71:
```suggestion
      "Possible covariance structures include: ",
```

R/tmb.R
```diff
@@ -64,11 +64,23 @@ h_mmrm_tmb_formula_parts <- function(formula) {
...
+      paste0(cov_functions, collapse = ", ")
```
danielinteractive, line 72: Use `toString`

R/tmb.R
```diff
@@ -64,11 +64,23 @@ h_mmrm_tmb_formula_parts <- function(formula) {
...
+    )
+  }
+  if (length(cov_selected) > 1) {
+    stop(
+      "Only one covariance structure can be specified. ",
+      "Currently specified covariance structure is: ",
```
danielinteractive, line 78:
```suggestion
      "Currently specified covariance structures are: ",
```

R/tmb.R
```diff
@@ -64,11 +64,23 @@ h_mmrm_tmb_formula_parts <- function(formula) {
...
+      paste0(names(cov_selected), collapse = ", ")
```
danielinteractive, line 79: Also here can use `toString`

R/tmb.R
```diff
@@ -79,8 +91,13 @@ h_mmrm_tmb_formula_parts <- function(formula) {
   # Parse the covariance term to get covariance type, visit and subject variables.
   assert_true(identical(length(cov_term), 2L)) # 2 because `fun (...)`.
   cov_content <- cov_term[[2L]]
-  assert_true(identical(length(cov_content), 3L)) # 3 because `y | x`.
-  assert_true(identical(as.character(cov_content[[1L]]), "|"))
+  if (!identical(length(cov_content), 3L) || !identical(as.character(cov_content[[1L]]), "|")) {
+    stop(
+      "Covariance structure must be of the form `",
+      cov_term[[1]],
+      "(VISIT|ID)`"
```
danielinteractive, line 98: Maybe use more generic placeholder names? E.g time and subject?

tests/testthat/test-tmb.R
```diff
@@ -36,6 +36,28 @@ test_that("h_mmrm_tmb_formula_parts works as expected", {
...
+    h_mmrm_tmb_formula_parts(FEV1 ~ RACE + AVISIT + USUBJID),
+    paste(
+      "Covariance structure must be specified in formula.",
+      "Possible covariance structure include: us, toep, toeph, ar1, ar1h, ad, adh, cs, csh"
+    )
+  )
+  expect_error(
+    h_mmrm_tmb_formula_parts(FEV1 ~ RACE + arh1(AVISIT | USUBJID)),
+    paste(
+      "Covariance structure must be specified in formula.",
```
danielinteractive, line 49: In this case maybe an error pointing out that a wrong structure was specified would be even more helpful?

danielinteractive (commented):
Thanks a lot Liming, great first PR! Please see inside

tests/testthat/test-tmb.R
clarkliming: I think this is a good idea, but I wonder if it will add some complexity here as function calls are possible in a formula, e.g. I have a function called `arh1`,  users may specify this term `arh1( a|b)`, where a and b are treated as logical.

In my opinion, this will add too much complexity if we are going to implement this

workflow:
1. check all terms that has a function call
2. check if these functions are available (e.g., offset, strata, these functions have special meanings in glm/cox models; arm in tern; or other built in functions like exp)
3. if functions are not available, check the similarity to our pre-defined covariance structure and report the most similar one

clarkliming: by the way, this may introduce additional dependencies like `stringdist`; or we rewrite it from scratch?

danielinteractive: hm ok that sounds a bit too complex for the value we get from it. Then let's keep as is on this point.

R/tmb.R
```diff
@@ -64,11 +64,23 @@ h_mmrm_tmb_formula_parts <- function(formula) {
...
-  cov_index <- found_specials[[which(cov_selected)]] + 1L # Subtract 1 for `list()`.
+  cov_selected <- Filter(Negate(is.null), found_specials)
+  if (length(cov_selected) == 0) {
+    stop(
+      "Covariance structure must be specified in formula. ",
+      "Possible covariance structures include: ",
+      toString(cov_functions)
+    )
+  }
+  if (length(cov_selected) > 1) {
```
danielinteractive, line 75:
```suggestion
  } else if (length(cov_selected) > 1) {
```

danielinteractive (approved):

Issue #98, opened 2022-09-01
refactor: code readability + package structure


kkmann:
Hey,

sorry that my first issue is not at all about extending the functionality of mmrm. I am proposing to refactor the code basis a bit to be closer to the tidyverse way of structuring a package. I would hope that this makes it easier to understand the code base and maybe also to contribute. As with all things related to taste, there is no right or wrong here :)

Key issues that I personally stumbled upon: 

- [ ] file names do not correspond to content, e.g. `satterthwaite.R` -> `df_md()`
- [ ] main functions defined at end of files (also `satterthwaite.R`) and start with detailed h_* functions
- [ ] reduce exported functions further - see https://github.com/openpharma/mmrm/blob/692ad6b0441056eece083a015a14296a2bde67ba/NAMESPACE#L16 for starting point
- [ ] h_* pattern is not really necessary (use @export, everything that's not exported is a helper); leads to bloated code and (if not explained) confusion --> we can try with a simple PR that really just substitutes "h_" with nothing and then see how it looks like
- [x] generic utility/helper function not related to package logic could go in `utils.R` to reduce the number of files and make it easier to figure out how the overall structure of the relevant package logic works (`output.R`)

Will start opening PRs with concrete suggestions, let's see whether you agree. Opinions welcome!

kkmann:
related: can someone explain how 

https://github.com/openpharma/mmrm/blob/b45fa724b5ace95d91cfee52edce517427939472/R/catch-routine-registration.R#L1-L8

works and what it is for? It seem to somehow be related to package skeletons but I was not able to figure it out. 

A candidate to be moved to utils.R as well...

kkmann:
currently 

https://github.com/openpharma/mmrm/blob/b45fa724b5ace95d91cfee52edce517427939472/R/fit.R#L23

is exported. 

My preference would be to stick with `mmrm()` as user-facing function (there is always `mmrm:::fit_single_optimizer)`) 

If there is a good use case for it, maybe rename to `fit_mmrm()`? 
a) I would favor no suffix names for the base case (i.e. if any have a `fit_multiple_optimizers`)
b) exporting a generic name as fit_* without the specific model or package name can lead to name space confusion (granted, low probability due to the _single_optimizer suffix, but still)

danielinteractive:
> related: can someone explain how
> 
[...]

PR #100, opened 2022-09-01
collect generic helper/utility functions in utils.R

kkmann:
# Description

suggesting to consolidate output.R and parallel.R in utils.R. I guess it makes sense to keep the helper function from satterthwaite.R in the same file since they are specific.


kkmann:
see #98

danielinteractive (approved):

kkmann:
looks like the gh actions run into issues with git checkout. not sure what's causing this. R CMD check runs perfectly fine locally.

danielinteractive:
@cicdguy any idea what is going on re: checks?

cicdguy:
> @cicdguy any idea what is going on re: checks?
Let me take a look.

cicdguy:
Ok I see the issue. Will need to create a separate PR for it.
The issue is that the GitHub token used for those workflows do not work for forks (for security reasons). 
I'll just replace the tokens with the generic ones.

cicdguy:
> Ok I see the issue. Will need to create a separate PR for it. The issue is that the GitHub token used for those workflows do not work for forks (for security reasons). I'll just replace the tokens with the generic ones.

Actually, never mind. I do see that the token is generic. It's got something to do with token permissions.

R/utils.R
```diff
@@ -1,3 +1,56 @@
...
+  list(
+    result = result,
+    warnings = setdiff(env$warning, remove$warnings),
+    errors = setdiff(env$error, remove$errors),
+    messages = setdiff(env$message, remove$messages)
+  )
+}
+
+
+
```
cicdguy, line 53:
```suggestion
```

cicdguy:
FYI @kkmann - feel free to revert/drop my commit. I ended up fixing the issue and the workflows should run fine now.

R/utils.R
```diff
@@ -1,3 +1,55 @@
...
+  )
+  list(
+    result = result,
+    warnings = setdiff(env$warning, remove$warnings),
+    errors = setdiff(env$error, remove$errors),
+    messages = setdiff(env$message, remove$messages)
+  )
+}
+
+
```
cicdguy, line 52:
```suggestion

```

danielinteractive:
Ah good point. @kkmann actually no need to fork - you can push feature branches etc on this repo. If that does not work please let us know

kkmann:
thanks, guess these are still toothing issues from seeting this free from you Roche internal dev ops.

PR #105, opened 2022-09-05
spatial covariance matrix

### LINKED ISSUE #82: Add a first spatial covariance structure: spatial exponential
To do:

- [x] Read a bit about spatial covariance structures, e.g. at https://documentation.sas.com/doc/en/pgmsascdc/9.4_3.3/statug/statug_mixed_syntax14.htm
- [x] Adapt algorithm vignette (write down theoretical part - e.g. in this case we don't start from a common covariance matrix for all patients but construct a new one for each patient based on the coordinates)
- [x] Adapt covariance structure vignette (write down theoretical part of the exponential cov structure)
- [x] Think about how to best receive the required numeric time point values from the user
	- [x] Note that so far we are just taking the levels of the time point (factor) variable in their order
	- [x] Note that also multiple numeric columns might represent together the coordinates of a single visit
- [x] Think about how to best pass on the numeric time coordinates to the C++ code
- [x] Add definition of spatial in general and spatial exponential covariance structure in particular to covariance vignette 
- [x] Adjust covariance structure choices in https://github.com/openpharma/mmrm/blob/709f392e966b12514525ea6f6118d8e9af32089c/R/tmb.R#L60
- [x] Initialize start parameters in https://github.com/openpharma/mmrm/blob/709f392e966b12514525ea6f6118d8e9af32089c/R/tmb.R#L208
	- [x] Amend tests in https://github.com/openpharma/mmrm/blob/709f392e966b12514525ea6f6118d8e9af32089c/tests/testthat/test-tmb.R#L80
- [x] Amend printing of covariance structure in https://github.com/openpharma/mmrm/blob/b1b612263623c9d8f9e972adf19d09e23558664c/R/mmrm-methods.R#L109
- [x] Add lower Cholesky factor function and include it in branching in 
https://github.com/openpharma/mmrm/blob/01ae7f936cd43ad8788b84a7a03467b217a9ccaa/src/covariance.h#L41
	- [x] Test it in https://github.com/openpharma/mmrm/blob/main/src/test-covariance.cpp
		- [x] Install and Restart package
		- [x] Build > More > Test package
- [x] Amend integration tests in https://github.com/openpharma/mmrm/blob/709f392e966b12514525ea6f6118d8e9af32089c/tests/testthat/test-tmb.R#L256
	- [x] Note that `r2stream` is not available outside of Roche, so just run SAS differently and save log as text file in the design folder as reference
- [x] Run Clean and Rebuild > Check to ensure that checks pass

clarkliming:
close #82 

currently, only spatial exponential is added in the vignette. Others to be added later.


vignettes/covariance.Rmd
```diff
@@ -200,4 +200,31 @@ homogeneous compound symmetry covariance structure with total only $k=2$
 variance parameters, otherwise we obtain the heterogeneous compound
 symmetry covariance structure with $k = m + 1$ variance parameters.
 
+# Spatial Covariance Structure
```
danielinteractive, line 203: Can we rename to "Euclidean" ?

src/mmrm_spatial.cpp
```diff
@@ -0,0 +1,134 @@
+#include "init.h"
```
danielinteractive, line 1: instead of starting a separate file, it would be better to only have if/else separation as needed for euclidean vs. ordered covariance structures within the old file/ program

danielinteractive (commented):
thanks @clarkliming , before going further into the coding, can you please add the proposal to the algorithm vignette? because this will be very helpful to see where this differs from other covariance structures

src/mmrm_spatial.cpp
```diff
@@ -0,0 +1,134 @@
...
+  // Go through all subjects and calculate quantities initialized above.
+  for (int i = 0; i < n_subjects; i++) {
+    // Start index and number of visits for this subject.
+    int start_i = subject_zero_inds(i);
+    int n_visits_i = subject_n_visits(i);
+    // Create the distance matrix.
+    matrix<Type> distance_i = distance.block(start_i, start_i, n_visits_i, n_visits_i);
+    // Obtain Cholesky factor Li.
+    vector<Type> theta_i = theta.segment(subject_groups(i) * theta_size, theta_size);
+    matrix<Type> cov_i = get_spatial_covariance_lower_chol(theta_i, distance_i);
```
kkmann, line 66: This should be the only structural difference from a non "spatial" (still don't like the term) covariance structure, right? Besides, passing in the precalculated distances. There might even be a case to be made for calculating the distances within TMB to keep the changes to the rest of the API small. I doubt that comes at a large performance hit.

vignettes/covariance.Rmd
kkmann: "Covariance structures induced by covariance functions" ?

"Gaussian process covariance structures"?

special case: stationary covariance functions based on Euclidean distance. I guess the best name depends a bit on whether there is any chance that we might want to support more covariance functions.

danielinteractive: I think we keep it to Euclidean for the time being. Anyway I don't see large demand for these.

danielinteractive:
> > thanks @clarkliming , before going further into the coding, can you please add the proposal to the algorithm vignette? because this will be very helpful to see where this differs from other covariance structures
> 
> hi @danielinteractive what proposal do you mean?

no worries, you have it already now in the algorithm vignette- I mean with proposal = spatial / euclidean/ non-ordered / GP cov structures.

src/mmrm.cpp
```diff
@@ -56,17 +57,18 @@ Type objective_function<Type>::operator() ()
   Type sum_log_det = 0.0;
   // Get theta_size
   int theta_size = theta.size() / n_groups;
-
+  // Get the spatial indicator
+  bool spatial = is_spatial(cov_type);
```
danielinteractive, line 61: you can directly pass this from R? then you don't need to define helper function

R/tmb.R
```diff
@@ -154,12 +154,14 @@ h_mmrm_tmb_formula_parts <- function(formula) {
...
+  cov_type <- deparse(cov_term[[1L]])
   structure(
-    list(
+    .Data = list(
       formula = formula,
       model_formula = model_formula,
       full_formula = full_formula,
-      cov_type = deparse(cov_term[[1L]]),
+      cov_type = cov_type,
+      spatial = cov_type %in% cov_type_spatial,
```
danielinteractive, line 164:
```suggestion
      is_spatial = cov_type %in% cov_type_spatial,
```
and adapt downstream

R/tmb.R
```diff
@@ -277,12 +277,28 @@ h_mmrm_tmb_data <- function(formula_parts,
     subject_groups <- factor(rep(0L, n_subjects))
     n_groups <- 1L
   }
-
-  structure(
+  coordinates <- full_frame[formula_parts$visit_var]
+  if (formula_parts$cov_type %in% cov_type_spatial) {
```
danielinteractive, line 281: we can use `formula_parts$is_spatial` here

R/tmb.R
```diff
@@ -277,12 +277,28 @@ h_mmrm_tmb_data <- function(formula_parts,
...
+    visits_zero_inds <- 0L
+    n_visits <- 2L
+  } else {
+    assert(identical(ncol(coordinates), 1L))
+    assert_factor(coordinates[[1L]])
+    visits_zero_inds <- as.integer(coordinates[[1L]]) - 1L
+    coordinates_matrix <- as.matrix(visits_zero_inds, ncol = 1)
+    n_visits <- nlevels(coordinates[[1L]])
+  }
+  ret <- structure(
```
danielinteractive, line 296:
```suggestion
  structure(
```

R/tmb.R
```diff
@@ -277,12 +277,28 @@ h_mmrm_tmb_data <- function(formula_parts,
...
+      assert_numeric
+    )
+    coordinates_matrix <- as.matrix(coordinates)
+    visits_zero_inds <- 0L
+    n_visits <- 2L
+  } else {
+    assert(identical(ncol(coordinates), 1L))
+    assert_factor(coordinates[[1L]])
+    visits_zero_inds <- as.integer(coordinates[[1L]]) - 1L
+    coordinates_matrix <- as.matrix(visits_zero_inds, ncol = 1)
```
danielinteractive, line 293: do we really then use this in C++? if not, I would suggest to just pass an empty matrix

R/tmb.R
```diff
@@ -277,12 +277,28 @@ h_mmrm_tmb_data <- function(formula_parts,
...
     n_groups <- 1L
   }
-
-  structure(
+  coordinates <- full_frame[formula_parts$visit_var]
+  if (formula_parts$cov_type %in% cov_type_spatial) {
+    lapply(
+      coordinates,
+      assert_numeric
+    )
```
danielinteractive, line 285:
```suggestion
    lapply(coordinates, assert_numeric)
```

R/tmb.R
```diff
@@ -328,7 +346,8 @@ h_mmrm_tmb_parameters <- function(formula_parts,
     ad = m,
     adh = 2 * m - 1,
     cs = 2,
-    csh = m + 1
+    csh = m + 1,
+    gp_exp = 2
```
danielinteractive, line 350:
```suggestion
    sp_exp = 2
```

R/tmb.R
```diff
@@ -392,7 +411,7 @@ h_mmrm_tmb_assert_opt <- function(tmb_object,
   }
 }
 
-#' Extract covariance matrix from `TMB` report and input data
+#' Extract covariance matrix from `TMB` report and tmb data
```
danielinteractive, line 414:
```suggestion
#' Extract covariance matrix from `TMB` report and input data
```
already says TMB upfront, point is that `tmb_data` is the input data

R/tmb.R
```diff
@@ -406,7 +425,12 @@ h_mmrm_tmb_assert_opt <- function(tmb_object,
 #' @keywords internal
 h_mmrm_tmb_extract_cov <- function(tmb_report, tmb_data, visit_var) {
   d <- dim(tmb_report$covariance_lower_chol)
-  visit_names <- levels(tmb_data$full_frame[[visit_var]])
+  visits <- tmb_data$full_frame[[visit_var]]
+  if (is.factor(visits)) {
+    visit_names <- levels(visits)
+  } else {
+    visit_names <- c(0, 1)
+  }
```
danielinteractive, line 433:
```suggestion
  visit_names <- if (is.factor(visits)) {
    levels(visits)
  } else {
    c(0, 1)
  }
```

R/tmb.R
```diff
@@ -406,7 +425,12 @@ h_mmrm_tmb_assert_opt <- function(tmb_object,
 #' @keywords internal
 h_mmrm_tmb_extract_cov <- function(tmb_report, tmb_data, visit_var) {
   d <- dim(tmb_report$covariance_lower_chol)
-  visit_names <- levels(tmb_data$full_frame[[visit_var]])
+  visits <- tmb_data$full_frame[[visit_var]]
```
danielinteractive, line 428: does this work for the spatial case? I thought then we might have `visit_var` being a non-scalar vector, i.e. more than one element?

R/utils.R
```diff
@@ -88,3 +88,21 @@ h_free_cores <- function() {
...
+#' ar1h: "heterogeneous auto-regressive order one"
+#' ad: "ante-dependence"
+#' adh: "heterogeneous ante-dependence"
+#' cs: "compound symmetry"
+#' csh: "heterogeneous compound symmetry"
+#' gp_exp: "spatial exponential"
+#' @rdname covariance_type.Rd
+cov_type <- c("us", "toep", "toeph", "ar1", "ar1h", "ad", "adh", "cs", "csh")
+#' @rdname covariance_type.Rd
+cov_type_spatial <- c("gp_exp")
```
danielinteractive, line 108:
```suggestion
cov_type_spatial <- c("sp_exp")
```

R/utils.R
```diff
@@ -88,3 +88,21 @@ h_free_cores <- function() {
...
+#' us: "unstructured"
+#' toep: "Toeplitz"
+#' toeph: "heterogeneous Toeplitz"
+#' ar1: "auto-regressive order one"
+#' ar1h: "heterogeneous auto-regressive order one"
+#' ad: "ante-dependence"
+#' adh: "heterogeneous ante-dependence"
+#' cs: "compound symmetry"
+#' csh: "heterogeneous compound symmetry"
+#' gp_exp: "spatial exponential"
```
danielinteractive, line 104:
```suggestion
#' sp_exp: "spatial exponential"
```

danielinteractive, line 104: move this down? just above the spatial?

R/utils.R
```diff
@@ -88,3 +88,21 @@ h_free_cores <- function() {
...
+#' @rdname covariance_type.Rd
```
danielinteractive, line 105:
```suggestion
#' @rdname covariance_type
```
(normally no Rd needed)
also is this a separate help page?

design/SAS/sas_gp_exp.R
```diff
@@ -0,0 +1,40 @@
+# Spatial exponential covariance structure ----
+formula <- FEV1 ~ cs(TIME | USUBJID)
```
danielinteractive, line 2:
```suggestion
formula <- FEV1 ~ sp_exp(TIME | USUBJID)
```
please adjust example R code here

src/covariance.h
```diff
@@ -127,6 +127,19 @@ matrix<T> get_compound_symmetry_heterogeneous(const vector<T>& theta, int n_visi
   return get_heterogeneous_cov(sd_values, fun);
 }
 
+//Spatial Exponential Cholesky factor.
```
danielinteractive, line 130:
```suggestion
// Spatial Exponential Cholesky factor.
```

src/covariance.h
```diff
@@ -127,6 +127,19 @@ matrix<T> get_compound_symmetry_heterogeneous(const vector<T>& theta, int n_visi
...
+template <class T>
+matrix<T> get_spatial_exponential(const vector<T>& theta, const matrix<T>& distance) {
+  T const_sd = exp(theta(0));
+  T logrho = exp(theta(1));
+  matrix<T> expdist = exp(-distance.array() *  logrho);
+  matrix<T> result = expdist * const_sd;
+  matrix<T> chol(distance.rows(), distance.cols());
```
danielinteractive, line 137: can you try without defining this first? I think you can just directly return the line 139 result.

src/covariance.h
```diff
@@ -157,15 +170,40 @@ matrix<T> get_covariance_lower_chol(const vector<T>& theta, int n_visits, std::s
   return result;
 }
 
+// Creates a new spatial covariance cholesky.
+template <class T>
+matrix<T> get_spatial_covariance_lower_chol(const vector<T>& theta, const matrix<T>& distance, std::string cov_type) {
+  matrix<T> result;
+  if (cov_type == "gp_exp") {
```
danielinteractive, line 177:
```suggestion
  if (cov_type == "sp_exp") {
```
probably easiest to search for `gp_` in the whole package and replace by `sp_` :-)

src/covariance.h
```diff
@@ -157,15 +170,40 @@ matrix<T> get_covariance_lower_chol(const vector<T>& theta, int n_visits, std::s
...
+    result = get_spatial_exponential<T>(theta, distance);
+  } else {
+    Rf_error(("Unknown covariance type '" + cov_type + "'.").c_str());
```
danielinteractive, line 180:
```suggestion
    Rf_error(("Unknown spatial covariance type '" + cov_type + "'.").c_str());
```

src/covariance.h
```diff
@@ -157,15 +170,40 @@ matrix<T> get_covariance_lower_chol(const vector<T>& theta, int n_visits, std::s
...
+  }
+  return result;
+}
+
 // Creates a grouped correlation object dynamically.
 template <class T>
-matrix<T> get_cov_lower_chol_grouped(const vector<T>& theta, int n_visits, std::string cov_type, int n_groups) {
+matrix<T> get_cov_lower_chol_grouped(const vector<T>& theta, int n_visits, std::string cov_type, int n_groups, int spatial) {
```
danielinteractive, line 187:
```suggestion
matrix<T> get_cov_lower_chol_grouped(const vector<T>& theta, int n_visits, std::string cov_type, int n_groups, bool spatial) {
```
should work? would be a bit nicer than int

src/covariance.h
```diff
@@ -157,15 +170,40 @@ matrix<T> get_covariance_lower_chol(const vector<T>& theta, int n_visits, std::s
...
+  if (spatial == 1) {
+    n_visits = 2;
```
danielinteractive, line 189: this is already set in R?

src/covariance.h
```diff
@@ -157,15 +170,40 @@ matrix<T> get_covariance_lower_chol(const vector<T>& theta, int n_visits, std::s
...
+  }
   matrix<T> result(n_visits * n_groups, n_visits);
   int covariance_size = theta.size() / n_groups;
-  for (int i = 0; i < n_groups; i++) {
-    matrix<T> lower_chol = get_covariance_lower_chol<T>(theta.segment(i * covariance_size, covariance_size), n_visits, cov_type);
-    result << result.block(0, 0, n_visits * i, n_visits), lower_chol;
+  if (spatial == 1) {
+    matrix<T> standard_dist(2, 2);
+    standard_dist << 0, 1, 1, 0;
```
danielinteractive, line 195: why do you take the 2x2 identity matrix here?

src/covariance.h
danielinteractive: also better consistently use `is_spatial` as argument and variable name

src/mmrm.cpp
```diff
@@ -33,11 +33,13 @@ Type objective_function<Type>::operator() ()
   DATA_VECTOR(y_vector);           // Response vector (length n).
   DATA_VECTOR(weights_vector);     // Weights vector (length n).
   DATA_IVECTOR(visits_zero_inds);  // Zero-based Visits vector (length n).
+  DATA_MATRIX(coordinates);        // Coordinates matrix.
   DATA_INTEGER(n_visits);          // Number of visits, which is the dimension of the covariance matrix.
   DATA_INTEGER(n_subjects);        // Number of subjects.
   DATA_IVECTOR(subject_zero_inds); // Starting indices for each subject (0-based) (length n_subjects).
   DATA_IVECTOR(subject_n_visits);  // Number of observed visits for each subject (length n_subjects).
   DATA_STRING(cov_type);           // Covariance type name.
+  DATA_INTEGER(spatial);           // Spatial covariance (1)? Otherwise non-spatial covariance.
```
danielinteractive, line 42:
```suggestion
  DATA_INTEGER(is_spatial_int);           // Spatial covariance (1)? Otherwise non-spatial covariance.
```
and then make it a bool below called `is_spatial`

src/mmrm.cpp
```diff
@@ -54,29 +56,39 @@ Type objective_function<Type>::operator() ()
   vector<Type> y_vec_tilde = vector<Type>::Zero(y_vector.rows());
   // Sum of the log determinant will be incrementally calculated here.
   Type sum_log_det = 0.0;
+  // Get theta_size
+  int theta_size = theta.size() / n_groups;
```
danielinteractive, line 60: wording is confusing, because `theta_size != theta.size()`.
please change, e.g.:
```suggestion
  // Get number of variance parameters for one group.
  int theta_one_group_size = theta.size() / n_groups;
```

vignettes/algorithm.Rmd
```diff
@@ -157,6 +157,22 @@ The parametrization of $\theta$ is similar to other non-grouped $\theta$.
 Assume that there are total number of $G$ groups, the length of $\theta$ is multiplied by $G$, and for each part, $\theta$ is parametrized in the same fashion.
 For example, for an unstructured covariance matrix, $\theta$ has $k = G * m(m+1)/2$ entries.
 
+#### Spatial covariance matrix
+
+Under spatial covariance structure, the covariance matrix is different for all subjects, if they have different visit timepoints.
+Following the notations in the previous section, for subject $i$ with total number of $m_i$ visits, we have
+
+\[
+\Sigma_{i}(m_{ij}, m_{ik}) = \sigma * f(dist(t_{m_j}, t_{m_k}))
```
danielinteractive, line 166: I find the notation can be simplified probably - e.g. you could write
```suggestion
\sigma_{ijk} = \sigma * f(dist(\boldsymbol{c}_{ij}, \boldsymbol{c}_{ik}))
```
and explain that this is the j, k element of the covariance matrix for patient i, and c are the corresponding coordinate vectors.

vignettes/covariance.Rmd
```diff
@@ -200,4 +200,35 @@ homogeneous compound symmetry covariance structure with total only $k=2$
...
+
+Spatial covariance structures, unlike other covariance structures,
+does not require that the timepoints be consistent between subjects.
+Instead, as long as the distance between visits can be quantified,
+the spatial covariance structure can be applied.
+Euclidean distance is the most common case.
+For each subject, the covariance structure can be different,
+as they can have totally different timepoints.
+Only homogeneous structures are allowed(the variance is constant in all timepoints).
```
danielinteractive, line 212:
```suggestion
does not require that the timepoints are consistent between subjects.
Instead, as long as the distance between visits can be quantified in terms of time and/or other coordinates,
the spatial covariance structure can be applied.
Euclidean distance is the most common case.
For each subject, the covariance structure can be different.
Only homogeneous structures are allowed (i.e. a common variance is used).
```

R/utils.R
clarkliming: I just wonder if we should keep a help page for supported covariance and their abbreviations

R/tmb.R
clarkliming: currently, I don't know how to consolidate the coordinates and visits_zero_inds; but they should all represent the same meaning; I wonder one day we shall use coordinates to replace visits_zero_inds

src/covariance.h
clarkliming: following SAS outputs, we have subject-level covariance matrix. I think it is enough to display the covariance matrix for the standard distance matrix c(0,1,1,0), it is not identity matrix.

src/covariance.h
clarkliming: could you please clarify what is set in R?

vignettes/covariance.Rmd
```diff
@@ -227,7 +227,7 @@ In previous examples, $\rho$ can take values from -1 to 1, but here we need to r
 Hence we have the following parametrization form.
 
 \[
-  \theta = (\log(\sigma), \log(-\log(\rho)))
+  \theta = (\log(\sigma), logit(\rho)))
```
danielinteractive, line 230: you can declare this as math operator then use the command here, then it renders upright (same as log)

vignettes/covariance.Rmd
```diff
@@ -200,4 +200,35 @@ homogeneous compound symmetry covariance structure with total only $k=2$
...
+where $d_{ij}$ is the distance between time point $i$ and time point $j$,
+
+A total number of parameters $k = 2$ is needed:
+
+The parameterization for $\theta$ is a little different from previous examples.
+In previous examples, $\rho$ can take values from -1 to 1, but here we need to restrict $\rho$ to (0, 1).
+Hence we have the following parametrization form.
+
+\[
+  \theta = (\log(\sigma), \log(-\log(\rho)))
```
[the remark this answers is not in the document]
danielinteractive: I think logit is just more natural to map (0,1) to real line, there should not be a measureable speed disadvantage

vignettes/algorithm.Rmd
```diff
@@ -157,6 +157,22 @@ The parametrization of $\theta$ is similar to other non-grouped $\theta$.
...
+\]
+
+The $(m_{ij}, m_{ik})$ element of $\Sigma_{i}$ is a function of the distance between $m_{ij}$ and $m_{ik}$ visit occurring on $t_{m_{ij}}$ and $t_{m_{ik}}$.
+$t_{m_{ij}}$ is the coordinate(time) of $m_{ij}$ visit for subject $i$.
+$\sigma$ is the constant variance.
+Usually we use Euclidean distance.
+
+Currently only spatial exponential covariance structure is implemented(with only one continuous time variable).
```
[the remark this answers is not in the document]
clarkliming: multiple coordinates supported now; at first I though it is not that necessary to support multiple coordinates because

1. the distance can be more complicated (there can be some changes to the euclidean distance, e.g, scale a specific variable, etc; but this can be done in R side)
2. extracting the terms can be complicated as there are multiple coordinates.

Anyway it is implemented now; the formula looks like  `dist(a,b)|c/d` ( I added `dist` here to avoid the confusion of adding a and b directly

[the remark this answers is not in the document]
danielinteractive: Thanks Liming, cool!
Re: formula - I think we could just say `sp_exp(a, b | c/d)` ?  Idea being everything before the `|` are the coordinates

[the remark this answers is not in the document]
clarkliming: updated now

src/covariance.h
clarkliming: oh n_visits is defined. updated

vignettes/algorithm.Rmd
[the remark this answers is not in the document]
danielinteractive: awesome thanks!

tests/testthat/_snaps/mmrm-methods.md
```diff
@@ -140,3 +140,66 @@
...
+      3859.1   3865.7  -1927.6   3855.1 
+    
+    Coefficients: 
+                Estimate Std. Error    df t value Pr(>|t|)    
+    (Intercept)     40.3        0.7 194.0      60   <2e-16 ***
+    ARMCDTRT         4.2        1.0 188.0       4    2e-05 ***
+    ---
+    Signif. codes:  0 '***' 0.001 '**' 0.01 '*' 0.05 '.' 0.1 ' ' 1
+    
+    Covariance estimate:
```
danielinteractive, line 200: maybe in addition to documentation (see my comment above) it would be good in this print statement to add the explanation for `spatial` cov estimate that it is for 2 timepoints with unit distance (and saying that I guess for more complicated spatial cov structures it will need to be more than that ?...)

danielinteractive (commented):
thanks Liming! almost there, last round of comments inside.

R/tmb.R
```diff
@@ -41,31 +41,32 @@ h_mmrm_tmb_control <- function(optimizer = stats::nlminb,
...
-      cov_term[[1]],
-      "(time|(group/)subject)`"
+      "Covariance structure must be of the form `time|(group/)subject`."
     )
   }
   visit_term <- cov_content[[2L]]
   if (!identical(length(visit_term), 1L)) {
     stop(
-      "`time` in `(time|(group/)subject)` must be specified as one single variable."
+      "`time` in `time|(group/)subject` must be specified as one single variable."
```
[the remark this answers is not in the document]
clarkliming: for spatial cov structure, e.g.,  `sp_exp(t1, t2, t3 | group / subject)`, the last part of this call, `t3 | group / subject` is used in this function. the `visit_term` is `t3`. The last part is consistent between spatial covariance structure and non-spatial. The `t3` should be a simple symbol instead of a composite like `(t3+t4)`.  So here I think it is fine as `time` indicate the visit term.

forgive my English vocabulary maybe I did not express myself clearly.

R/tmb.R
```diff
@@ -75,29 +76,68 @@ h_mmrm_tmb_extract_vars <- function(cov_term) {
...
+    # spatial can have sp_exp(time1, ... | group / subject)
+    visit_terms <- cov_term[c(-1L, -length(cov_term))]
+    term_lengths <- vapply(
+      visit_terms,
+      length,
+      1L
+    )
+    if (any(term_lengths != 1L)) {
+      stop(
+        "Spatial covariance term shoul have all `time` variables as single."
```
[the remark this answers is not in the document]
clarkliming: same as the above: Here I use `single` for very simple symbols instead of a composite. Could you please help me improve the wording?

tests/testthat/_snaps/mmrm-methods.md
clarkliming: according to https://documentation.sas.com/doc/en/pgmsascdc/9.4_3.3/statug/statug_mixed_syntax14.htm  most of the spatial covariance matrix are simple enough to be represented using `d_{ij}`, and we use `d_{ij} = 1` to represent the covariance matrix; it is difficult for users to create covariance matrix of a custom distance matrix though.

R/tmb.R
```diff
@@ -41,31 +41,32 @@ h_mmrm_tmb_control <- function(optimizer = stats::nlminb,
...
-#' @return Named list of with elements:
+#' Extract group/subject vars from covariance term
+#' @param cov_content (`call`)\cr covariance term
+#' @details a covariance term, which takes the following form:
+#' `us(time | group / subject)`, or `sp_exp(time1, time2, ... | group / subject)`.
+#' Only for spatial covariance structure can have multiple time coordinates.
+#' This function, take the call of `time | group / subject`, last element of the covariance term,
+#' extract the corresponding grouping var, time var and subject var.
+#' @return Named list with elements:
 #' - `visit_var`: `string` with the visit variable name.
```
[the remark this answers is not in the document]
clarkliming: use `character` instead

tests/testthat/_snaps/mmrm-methods.md
danielinteractive: yeah let's keep it for now like this - but then still good to add explanation to the print statement here so that the user understand what is this

clarkliming: added in covariance.md, Spatial Covariance Structure section and details of Methods for `mmrm` Objects.

danielinteractive (approved):
Thanks Liming looks great! Do we have a test for multiple coordinates ? Both unit tests and integration test comparing with SAS results? If that is in then good to merge

clarkliming:
> Thanks Liming looks great! Do we have a test for multiple coordinates ? Both unit tests and integration test comparing with SAS results? If that is in then good to merge

integration tests added now

clarkliming:
> thanks @clarkliming - however the example you added is a bit concerning, in the sense that should we really allow the same variable to be used twice as coordinates? just on first thought that looks like something that should not be done.
> 
> In addition, for the integration test it would be better to use the "usual" situation of two different variables used as coordinates I think

well since they are coordinates.... it is legal to use the same variable multiple times.  Or do you mind if I just create another variable with similar value(e.g, VISIT2 = VISITN + 1). The results will be the same but the formula is different.

danielinteractive:
But would it ever make sense to use the same dimension (variable) more than once as coordinate? I don't think so.

Therefore can you please:
1) check that during the formula parsing and give an error in case same variable is used twice

2) add more variables to the example dataset so that you can use them for a realistic example? For the integration test

Thanks!

clarkliming:
> But would it ever make sense to use the same dimension (variable) more than once as coordinate? I don't think so.
> 
> Therefore can you please:
> 
> 1. check that during the formula parsing and give an error in case same variable is used twice
> 2. add more variables to the example dataset so that you can use them for a realistic example? For the integration test
> 
> Thanks!

for 1, this can be done, but I don't think we should stop this, a warning should be sufficient. anyway this is not wrong, but not making sense, right?

for 2, VISITN2 = VISITN + 1; do you think it is fine to use the formula  `FEV1 ~ sp_exp(VISITN, VISITN2 | USUBJID)` ?

danielinteractive:
Ok for 1) let's do a warning (please add a unit test to check that) and 2) yes please random numbers in the dataset so that we have two different coordinates

danielinteractive (approved):
congratulations Liming!!

PR #112, opened 2022-09-26
renaming h_mmrm_tmb_control to mmrm_control

kkmann:
This PR is part of #98 and renames `h_mmrm_tmb_control` to `mmrm_control`. 

The current options in mmrm_tmb_control are not TMB specific and it would be best to hide as much of the implementation details (TMB) from the end user as possible. The new naming scheme is similar to other packages nloptr. The `_tmb` part does not add any information in the context of this package (all based on TMB anyway) but might be rather confusing to end users and unecessarily prolongs code lines. 

Since this is a fucntion that needs to be exported, I would drop the `h_` prefix.

The PR also creates a separate section for the marginal means topic (deserves it!)

- [x] should `mmrm_control()` be moved to `fit.R` since is not TMB specific? (see #113) (-> fit.R)


_pkgdown.yml
```diff
@@ -18,25 +18,27 @@ reference:
...
+      - free_cores
   - title: Methods
     contents:
-      - emmeans_support
       - mmrm_methods
       - mmrm_tmb_methods
   - title: Covariance Structures
     contents:
       - covariance_types
+  - title: Estimate Marginal Means
```
BrianLang, line 39: Seems a bit off-topic in this PR and without proper motivation

danielinteractive: Thanks @BrianLang for having a look!
In this case, since it is a small clean-up change, it is ok to include here.

danielinteractive (approved):
thanks @kkmann ! You can also add yourself to the authors of this package.

danielinteractive:
> * [ ]  should `mmrm_control()` be moved to `fit.R` since is not TMB specific? (see [mmrm vs mmrm_tmb abstraction #113](https://github.com/openpharma/mmrm/issues/113))

@kkmann Hm both locations are ok I think. For now maybe let's keep it where it is?

kkmann:
ok, since both are fine I am moving it to fit.R to reflect the current structure (not TMB specific).

danielinteractive:
@kkmann I see that you merged despite failed linter checks ... let me see that I can make those mandatory to pass before merging ...

PR #119, opened 2022-10-07
added test case for extreme scaling

### LINKED ISSUE #109: Add a test case where scales are leading to convergence problems
To do:

- [x] come up with extreme scaling case, e.g. one coefficient OLS estimate should be 100000, other one should be very low e.g. 10^-5, variance very large -> does it work?
- [x] if yes -> add as a test case, done.
- [ ] if no -> let's have a look

kkmann:
closes #109 

Scaling of the likelihood function does not seem to be a problem. I suggest that we keep the test case and close the issue for now.


danielinteractive:
@danleibovitz can you maybe review this given you wanted to work on this topic? then it fits well together

tests/testthat/test-fit.R
```diff
@@ -84,6 +84,33 @@ test_that("fit_single_optimizer gives error messages", {
...
+  # we generate an example data set with two variables on very different scales
+  # after rescaling to sd = 1 both would have unit effect on the outcome
+  theta <- 1e6 # the scaling factor between the two variables
+  # generate some data for 50 individuals with 2 time points each
+  set.seed(42L)
+  id <- factor(rep(1:50, each = 2))
+  visit <- factor(rep(c(1, 2), 50))
+  x1 <- rep(rnorm(50, sd = sqrt(theta)), each = 2) # large scale
+  x2 <- rep(rnorm(50, sd = 1 / sqrt(theta)), each = 2) # small scale
+  y <- x1 / sqrt(theta) + sqrt(theta) * x2 + + rnorm(100, sd = .5) # add some noise
```
danleibovitz, line 97: I think this is a totally harmless typo, but there's '+ +' instead of '+' in line 97

danleibovitz (approved):
The only issue is line 97, which I think is harmless.
Kevin beat me to this by several days, and his test is more elegant than mine would have been so everybody wins :) I'll self-assign a new issue.

tests/testthat/test-fit.R
kkmann: great spot, thx. Weird that it works in the first place... fixed.

danielinteractive (approved):
thanks Kevin! please add yourself to package authors

Issue #118, opened 2022-10-07
refactoring: do not export h_mmrm_tmb


kkmann:
Dear all,

after merging #112 the only remaining exported h_* helper would be `h_mmrm_tmb()`. If I interpret the discussion in #113 correctly this is not intended for the typical "end user" anyway. In this case, I suggest that we simply don't export it. 
It can still be used via `mmrm:::h_mmrm_tmb()` indicating that you really should know what you are doing. The public interface would then be free from any `_tmb` or `h_` names.

kkmann:
might be better to rename to fit_mmrm

PR #120, opened 2022-10-07
renaming h_mmrm_tmb to fit_mmrm

kkmann:
exactly that, see #118 

Renaming makes more sense here since the function should actually be exported for advanced use. fit_* is close to the naming conventions of other packages of providing more fine-grained control over the fitting process. 

With this PR no `h_` or `_tmb` objects are exported anymore and the reference documentation lists functions roughly in order of importance to people new to the package.


R/tmb-methods.R
```diff
@@ -13,7 +13,7 @@
 #'
 #' @examples
 #' formula <- FEV1 ~ RACE + SEX + ARMCD * AVISIT + us(AVISIT | USUBJID)
-#' object <- h_mmrm_tmb(formula, fev_data, weights = rep(1, nrow(fev_data)))
+#' object <- fit_mmrm(formula, fev_data, weights = rep(1, nrow(fev_data)))
```
danielinteractive, line 16: unrelated but same file therefore linter checks were triggered:
guess needs a `# nolint` at end of line 105

R/tmb.R
```diff
@@ -537,7 +537,7 @@ h_mmrm_tmb_fit <- function(tmb_object,
   )
 }
 
-#' Fitting an MMRM with `TMB`
+#' Low-Level Fitting Function for MMRM
```
danielinteractive, line 540: unrelated, but need to break line 435 to avoid line length excess

R/tmb.R
```diff
@@ -567,8 +567,8 @@ h_mmrm_tmb_fit <- function(tmb_object,
 #' @examples
 #' formula <- FEV1 ~ RACE + SEX + ARMCD * AVISIT + us(AVISIT | USUBJID)
 #' data <- fev_data
-#' system.time(result <- h_mmrm_tmb(formula, data, rep(1, nrow(fev_data))))
-h_mmrm_tmb <- function(formula,
+#' system.time(result <- fit_mmrm(formula, data, rep(1, nrow(fev_data))))
```
danielinteractive, line 570: since now this is called `fit_*` and does not need TMB understanding, would it make sense to move to `fit.R`?... not necessary just wondering

danielinteractive (commented):
thanks @kkmann ! good stuff and please see inside few comments

danielinteractive:
@kkmann can you maybe address the comments ? generally let's aim to have PRs closed quickly. thanks!

R/tmb.R
kkmann: moved.

R/tmb-methods.R
kkmann: ah, now I get it - we cant even comment on line 105 directly i.e. linting issues. x/

kkmann: Bit of a riddle to me. There is a random `# nolint` is in line 106. It is also inconsistent with other S3 method definitions (they are not cought by the lintr). See eg line 89.

kkmann: seems to be solved now

danielinteractive (approved):

PR #123, opened 2022-10-11
104 hessian nlminb

### LINKED ISSUE #104: Use Hessian for `nlminb()`
To do:

- [x] Improve the algorithm vignette details
- [x] Let's use the Hessian for `nlminb()`
	- [ ] maybe we need a flag to decide whether optimization function takes the `hessian` argument?
	- [ ] add tests specifically for `nlminb()` use

Original:

Hey,

the docs state 

https://github.com/openpharma/mmrm/blob/e6262cb35ad733aa5d45b0e853e666bffc3f74d4/vignettes/algorithm.Rmd#L167

However, looking at the code 

https://github.com/openpharma/mmrm/blob/e6262cb35ad733aa5d45b0e853e666bffc3f74d4/R/tmb.R#L443-L453

and 

https://github.com/openpharma/mmrm/blob/e6262cb35ad733aa5d45b0e853e666bffc3f74d4/R/fit.R#L23-L28

it would see that we only use gradient information (not Hessian!) and quasi Newton methods (Newton-Raphson for minimization would use exact Hession). Is that correct? I guess the Hessian comes into play for the standard errors?

clarkliming:
close #104

1. Hessian is always used for `nlminb`
2. for AR1 and AR1H, the starting value for `rho` is 0, and the Hessian can not be evaluated. The starting value is changed to 1 for `rho`.
3. add introduction that quasi-Newton Raphson is used
4. there should be already many test cases using `nlminb` (the default of `mmrm_control` is `nlminb`)


R/tmb.R
```diff
@@ -588,16 +587,20 @@ h_mmrm_tmb <- function(formula,
...
-      args = c(
-        list(par, fn, gr, control = control$optimizer_control),
-        control$optimizer_args
-      )
+    c(
+      list(par, fn, gr, hessian = he, control = control$optimizer_control),
+      control$optimizer_args
     )
   )
+  if (!identical(control$optimizer, stats::nlminb)) {
```
kkmann, line 597: Is this a stable way of checking optimizer class? Probably, just wanted to double check.

kkmann (approved):
Great, I didn't realize that the hessian was already computed as part of the tmb object...

R/tmb.R
clarkliming: `identical` should work in checking the functions if they are identical here

clarkliming: but we can save the `optimizer` in character and save it to `mmrm_control`

danielinteractive: hm I guess better compare the functions, as is

danielinteractive:
> 2. for AR1 and AR1H, the starting value for `rho` is 0, and the Hessian can not be evaluated. The starting value is changed to 1 for `rho`.

I would not set it to 1 since that is at the border of the parameter space, how about 0.5? 
and do we know why it cannot be evaluated at 0? I mean is this a technical problem or mathematical fact?

> 4. there should be already many test cases using `nlminb` (the default of `mmrm_control` is `nlminb`)

maybe just add one more test case where you explicitly set it, since the default might change over time potentially

clarkliming:
> > 2. for AR1 and AR1H, the starting value for `rho` is 0, and the Hessian can not be evaluated. The starting value is changed to 1 for `rho`.
> 
> I would not set it to 1 since that is at the border of the parameter space, how about 0.5? and do we know why it cannot be evaluated at 0? I mean is this a technical problem or mathematical fact?

set to 0.5 now; I believe this is related to mathematical fact; we first construct the covariance matrix and then obtain the Cholesky decomposition. If `rho` is 0, then all elements in that matrix are all 0, this matrix is not positive-definite, so the Cholesky decomposition is not feasible. any value other than 0 will give correct Cholesky decomposition.


> 
> > 4. there should be already many test cases using `nlminb` (the default of `mmrm_control` is `nlminb`)
> 
> maybe just add one more test case where you explicitly set it, since the default might change over time potentially

https://github.com/openpharma/mmrm/blob/9a97cb43c26369b0a2a5540a53e917645e3e577d/tests/testthat/test-fit.R#L25

there are already tests using nlminb or optim, add more tests here? or test in the inner function `h_mmrm_tmb`?

danielinteractive:
> > > 2. for AR1 and AR1H, the starting value for `rho` is 0, and the Hessian can not be evaluated. The starting value is changed to 1 for `rho`.
> > 
> > 
> > I would not set it to 1 since that is at the border of the parameter space, how about 0.5? and do we know why it cannot be evaluated at 0? I mean is this a technical problem or mathematical fact?
> 
> set to 0.5 now; I believe this is related to mathematical fact; we first construct the covariance matrix and then obtain the Cholesky decomposition. If `rho` is 0, then all elements in that matrix are all 0, and during the cholesky decomposition algorithm, devision by 0 could happen and this lead to incorrect result

Hm I don't understand this yet... if rho = 0 then we just have an identity matrix for the covariance, which is fine, Cholesky factor will be identity matrix as well. but ok we can leave as is because results are still the same as before

> there are already tests using nlminb or optim, add more tests here? or test in the inner function `h_mmrm_tmb`?

looks good now, thanks!

clarkliming:
> > > > 2. for AR1 and AR1H, the starting value for `rho` is 0, and the Hessian can not be evaluated. The starting value is changed to 1 for `rho`.
> > > 
> > > 
> > > I would not set it to 1 since that is at the border of the parameter space, how about 0.5? and do we know why it cannot be evaluated at 0? I mean is this a technical problem or mathematical fact?
> > 
> > 
> > set to 0.5 now; I believe this is related to mathematical fact; we first construct the covariance matrix and then obtain the Cholesky decomposition. If `rho` is 0, then all elements in that matrix are all 0, and during the cholesky decomposition algorithm, devision by 0 could happen and this lead to incorrect result
> 
> Hm I don't understand this yet... if rho = 0 then we just have an identity matrix for the covariance, which is fine, Cholesky factor will be identity matrix as well. but ok we can leave as is because results are still the same as before

My bad. should not be related to cholesky part but maybe there is a bug in the autoderivative part; the hessian is not correct when theta is 0. using other numeric methods the hessian can be obtained correctly. 

> 
> > there are already tests using nlminb or optim, add more tests here? or test in the inner function `h_mmrm_tmb`?
> 
> looks good now, thanks!

R/tmb.R
danielinteractive: sorry a minor thing here to improve, I would do it like first setting the args without the hessian, and then if we are in the special case of nlminb add the hessian

vignettes/introduction.Rmd
```diff
@@ -81,6 +81,9 @@ Users can specify which optimizer should be used, changing from the default of
 other choices if optimization fails to converge. Other choices are `BFGS`, `CG`, and
 `nlminb`.
 
+`L-BFGS-B`, `BFGS` and `CG` are all implemented with `stats::optim` and the Hessian is not used,
+while `nlminb` is using `stats::nlminb`, both the gradient and the Hessian.
```
danielinteractive, line 85:
```suggestion
`L-BFGS-B`, `BFGS` and `CG` are all implemented with `stats::optim()` and the Hessian is not used,
while `nlminb` is using `stats::nlminb() which in turn uses both the gradient and the Hessian for the optimization.
```

danielinteractive (approved):

PR #127, opened 2022-10-13
124: reduce lib size almost half

### LINKED ISSUE #124: Try to reduce size of shared library
Background: Currently we have a large shared library object from compiling the whole TMB into our package. This might cause problems for CRAN submission.

To do:
- [x] check current size of shared library - does it cause CRAN note during `R CMD check --as-cran`?
	- 133.8 MB currently. seems too big
	- `checking installed package size ... NOTE: installed size is 131.9Mb`
- [ ] is the idea in https://github.com/boennecd/survTMB/blob/master/src/_tmb.cpp feasible for our package?  They only precompile / include the necessary functions they need.
	- [ ] or not because we need too many functions?
	- [x] not tried yet, seems to complicated
- [x] other idea: avoid precompilation
	- [x] as explained here: https://rdrr.io/cran/TMB/man/precompile.html precompilation increases size of the binary
	- [x] we can get down to 73 Mb
	- [ ] however, `glmmTMB` only has 14 Mb ... see https://www.r-project.org/nosvn/R.check/r-release-windows-x86_64/glmmTMB-00check.html
- [x] check with TMB team on other possibilities to reduce size
	- [x] https://github.com/kaskr/adcomp/issues/369

danielinteractive:
closes #124 (for now...)

This reduces the lib size from 133 to 73 Mb.


danielinteractive:
@cicdguy bit off topic, but are you guys using the `--strip` option for `R CMD INSTALL` when installing packages into the docker images etc.? It seems that can save a lot of space. (We cannot do this in the package directly)

clarkliming (approved):
all looks good and I don't know how to reduce the size further.
but let's keep the issue open and see if there are any updates from tmb developer (and we can try following the survTMB or glmmTMB)

PR #128, opened 2022-10-14
124: reduce the size using strip

clarkliming:
following 

http://dirk.eddelbuettel.com/blog/2017/08/14/#009_compact_shared_libraries

it seems that the build package can be smaller if strip is enabled

the package is much smaller and the `mmrm.so` object is only 16 mb

### DIFF
--- a/src/Makevars
+++ b/src/Makevars
@@ -1,1 +1,4 @@
 PKG_CXXFLAGS = -D TMBAD_FRAMEWORK -D EIGEN_PERMANENTLY_DISABLE_STUPID_WARNINGS
+strippedLib: $(SHLIB)
+	if test -e "/usr/bin/strip"; then /usr/bin/strip --strip-debug $(SHLIB); fi
+.phony: strippedLib


danielinteractive (commented):
thanks @clarkliming for finding that information, really great!!

Can you try instead to work with the flags?
I guess some of these can be dropped in our case (e.g. the CXX version specific ones and Fortran one):

SHLIB_CXXLDFLAGS = -Wl,-S -shared
SHLIB_CXX11LDFLAGS = -Wl,-S -shared
SHLIB_CXX14LDFLAGS = -Wl,-S -shared
SHLIB_FCLDFLAGS = -Wl,-S -shared
SHLIB_LDFLAGS = -Wl,-S -shared

I would be curious to see if that works as well and then we don't need the strip tool

clarkliming:
> thanks @clarkliming for finding that information, really great!!
> 
> Can you try instead to work with the flags? I guess some of these can be dropped in our case (e.g. the CXX version specific ones and Fortran one):
> 
> SHLIB_CXXLDFLAGS = -Wl,-S -shared SHLIB_CXX11LDFLAGS = -Wl,-S -shared SHLIB_CXX14LDFLAGS = -Wl,-S -shared SHLIB_FCLDFLAGS = -Wl,-S -shared SHLIB_LDFLAGS = -Wl,-S -shared
> 
> I would be curious to see if that works as well and then we don't need the strip tool

I have tried, but it does not work on my computer; the compiled library `mmrm.so` is around 50 MB still

danielinteractive:
@clarkliming and if you do both? flags and strip? is it even smaller?

clarkliming:
> @clarkliming and if you do both? flags and strip? is it even smaller?

these flags just do not work, combined with /without strip on my side. Does it work for you ?

danielinteractive:
Hm I just looked at https://cran.r-project.org/doc/manuals/R-exts.html
and found this:

`The strip utility is platform-specific (and CRAN prohibits removing debug symbols). For example the options --strip-debug and --strip-unneeded of the GNU version are not supported on macOS nor Solaris[84](https://cran.r-project.org/doc/manuals/R-exts.html#FOOT84): the POSIX standard for strip does not mention any options, and what calling it without options does is platform-dependent. Stripping a .so file could even prevent it being dynamically loaded into R on an untested platform.
ld -S invokes strip --strip-debug on GNU ld but is not portable: in particular on Solaris it does something completely different and takes an argument.`

clarkliming:
oh oops

danielinteractive:
@clarkliming yeah I guess we better don't do that. I also read a bit on http://dirk.eddelbuettel.com/blog/2019/03/27/ and looked at the cited packages. They are on CRAN but they don't use this trick.

PR #131, opened 2022-10-14
117: test emmeans for weighted MMRM

### LINKED ISSUE #117: Add a test for emmeans from weighted MMRM
This is just to be sure we are not missing anything for weighted MMRMs there.

To do:

- [x] Add a new test to `test-emmeans.R`
  - [x] use a weighted MMRM
  - [x] run same analysis with SAS
  - [x] see that we match results up to numerical precision
- [x] Avoid any rounding in this file of results before comparing them to expectations
  - [x] if tolerance is too strict then reduce that slightly, e.g. `1e-3`

danielinteractive:
closes #117

### DIFF
--- a/R/data.R
+++ b/R/data.R
@@ -16,6 +16,7 @@
 #'   - `SEX`: sex.
 #'   - `FEV1_BL`: FEV1 at baseline (%).
 #'   - `FEV1`: FEV1 at study visits.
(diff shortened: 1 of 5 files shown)


clarkliming (approved):
all looks good and results are correct

PR #134, opened 2022-10-14
129: Avoid failing for convergence errors.

### LINKED ISSUE #129: Try to reproduce error and fix the cause
Error: was observed in downstream `tern.mmrm` test
image

Reprex:
```r
small_dat <- data.frame(
    FEV1 = c(1, 2, 3, 4, 5, 6),
    AVISIT = factor(c("V1", "V1", "V2", "V3", "V3", "V4")),
    USUBJID = c("A", "B", "A", "C", "D", "A")
  )
  vars <- list(
    response = "FEV1",
    id = "USUBJID",
    visit = "AVISIT"
  )
  fit <- mmrm::mmrm(
    formula = FEV1 ~ AVISIT + ar1(AVISIT | USUBJID),
    data = small_dat
  )
```

Problem: I cannot reproduce this error locally.

To do:
- [x] try to reproduce locally -> @BrianLang could reproduce it!
- [x] if locally cannot reproduce, use docker image http://ghcr.io/insightsengineering/rstudio_4.2.1_bioc_3.15 and try to reproduce there -> not reproducible
- [x] debug where it is coming from
- [x] add unit test
- [x] fix:
	- [x] `h_summarize_all_fits`: also include errors
	- [x] `fit_single_optimizer`: selectively exclude convergence failures from `stop()` since these should be included as `errors` in return struct
	- [x] `h_mmrm_tmb_assert_opt`: 
		- [x] always fail if convergence problems, don't just give warnings (it is an assertion function)
		- [x] use the same eigen threshold as `solve()` does or just directly use `solve()` since that is what is needed in the fit function next
- [x] new approach is needed, because we rely on the fact that `fit_single_optimizer` returns an `mmrm_fit` object even in the case of non-convergence - and it seems too much change now to throw this over.
	- [x] refactor `h_mmrm_tmb_assert_opt` into `h_mmrm_tmb_check_conv` which will be used after doing the `h_mmrm_tmb_fit`
	- [x] catch the `solve` error in `h_mmrm_tmb_fit`
	- [x] remove again the error handling stuff from before since now it will just be warnings

danielinteractive:
closes #129


clarkliming (approved):

danielinteractive:
@BrianLang if you are still looking, you could have a review look :-)

BrianLang:
It seems that our example in `helper-examples.R` doesn't actually converge, it gives: `convergence = 1, message = "new_X"`. And since we throw an error and don't return the fit we now have a broken `helper-examples.R` and testing suite. 
😅

danielinteractive:
> It seems that our example in `helper-examples.R` doesn't actually converge, it gives: `convergence = 1, message = "new_X"`. And since we throw an error and don't return the fit we now have a broken `helper-examples.R` and testing suite. sweat_smile

yeah I changed plans now, see the issue :-)

BrianLang:
`convergence = 1` is just that it reached max iterations, could maybe also adjust our `helper-examples.R` to make sure things converge.

PR #149, opened 2022-10-24
add basic explorations of KR df

clarkliming:
explore how we can implement Kenward-Roger df


design/KR/kenward.Rmd
```diff
@@ -0,0 +1,64 @@
...
+
+In current mmrm package, we provided Satterthwaite degree of freedom, through `df_1d` and `df_md`.
+
+# Design of Kenward-Roger degree of freedom
+
+## update to current functions
+
+* `df_1d` and `df_md` renamed to `df_1d_sat` and `df_md_sat`, do not export them
+* add a new function `df_1d` and `df_md` to include degree of freedom selection
+* implement `df_1d_kr(2)` and `df_md_kr(2)` (suffix 2 means that it is improved Kenward-Roger degree of freedom)
```
danielinteractive, line 23:
```suggestion
* implement `df_1d_kr2` and `df_md_kr2` (suffix 2 means that it is improved Kenward-Roger degree of freedom)
```
right because we don't want parentheses in the function names?

design/KR/kenward.Rmd
```diff
@@ -0,0 +1,64 @@
...
+* consider renaming the `df_1d` and `df_md` to something like `beta_test_1d` and `beta_test_md`?
```
danielinteractive, line 24: yeah maybe in a second step - in the first step we would need to be reverse-compatible / deprecate previous use

design/KR/kenward.Rmd
```diff
@@ -0,0 +1,64 @@
...
+
+## psedo code
```
danielinteractive, line 26:
```suggestion
## pseudo code
```

design/KR/kenward.Rmd
```diff
@@ -0,0 +1,64 @@
...
+
+* Current `mmrm` implementation is kind of complete, with the tests already implemented. If we are going to rely on `pbkrtest`
+package to implement Kenward-Roger degree of freedom, we are not able to include this method in `mmrm` otherwise we can introduce
+cyclic imports
+* `pbkrtest` is not well tested and we need effort to reliablely use that package
+* `pbkrtest` implements the original Kenward-Roger degree of freedom instead of the improved one(recommended)
+* SAS software provides both Kenward-Roger degree of freedom
+* `lmerTestR` suggests that "The culprit is the calculation of the scaling of the $F$-value for which **pbkrtest** does not appear to export a low-level (direct) method."
+which leads to slow speed which we can improve
+
```
danielinteractive, line 57: thanks @clarkliming good start! can you please continue then with one section that includes mathematical details of the KR2 calculations, plus another section with prototype code? (can be pseudo code or C++ as well)

danielinteractive (commented):
thanks Liming, good start, please see inside

design/KR/kenward.Rmd
clarkliming: yes we don't want parentheses, but here I mean `df_1d_kr` and `df_1d_kr2` for original and improved KR df

danielinteractive: Ah ok both versions. Ok or could do via an argument

design/KR/kenward.Rmd
```diff
@@ -194,7 +194,40 @@ to be added.
...
+  c1 <- g / denom
+  c2 <- (nl - g) / denom
+  c3 <- (nl + 2 - g) / denom
+  v_star <- 2 / nl * (1 + c1 * b) / (1 - c2 * b)^2 / (1 - c3 * b)
+  rho <- v_star / e_star^2
+  m <- 4 + (nl + 2)  / (nl * rho - 1)
+  lambda <- m / (e * (m - 2))
+  return(list(m = m, lambda = lambda))
+}
+```
```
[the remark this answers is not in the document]
danielinteractive: yeah also works - just thinking good to double check along the way that we are on the right path :-)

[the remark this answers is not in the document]
clarkliming: I wonder, instead of making it a "design" issue, why not implement the algorithm directly? the mathematical details are already there, and after the implementation, I can add integration tests to compare results between SAS and R.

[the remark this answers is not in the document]
danielinteractive: because it is a big change in the code base. those things we always do in 2 steps. design first, then implementation. saves time in the end, because design can change and then you don't need to update all documentation and tests all the time.

design/KR/kenward.Rmd
kkmann: I wonder, these function names get a bit clunky. Do we need to export them in the first place and could those be turned into S3 generics, like `degrees_of_freedom()`? I think the tidyverse style guideline discourages abbreviations although in this case given the audience, it probably makes sense to use `df()`.

danielinteractive: yeah `df()` could be nice. not sure about method dispatch here.

design/KR/kenward.Rmd
```diff
@@ -0,0 +1,402 @@
...
+## Further updates to the naming
+* consider renaming the `df_1d` and `df_md` to something like `beta_test_1d` and `beta_test_md`?(the names need to be discussed)
+* add deprecation notes on `df_1d` and `df_md`
+
+## pseudo code
+
+```{r, eval=FALSE}
+# df_1d
+formula <- FEV1 ~ ARMCD + ad(AVISIT | USUBJID)
+result <- mmrm(formula, fev_data, reml = FALSE)
```
danielinteractive, line 34: I think we will already need to have an argument here to signal `mmrm()` and inside C++ code that we also want the adjusted variance-covariance matrix of the fixed effects to be returned.

design/KR/kenward.Rmd
```diff
@@ -0,0 +1,402 @@
...
+df_1d(result, contrast = c(0, 1), df = "SAT") # test the armcd using SAT
+# df_1d_sat is called
+df_1d(result, contrast = c(0, 1), df = "KR") # KR defaults to KR2 (improved version)
```
danielinteractive, line 37: and then here if we ask for `KR` but it has not been calculated in `mmrm()` we fail

design/KR/kenward.Rmd
```diff
@@ -0,0 +1,402 @@
...
+* `pbkrtest` implements the original Kenward-Roger degree of freedom instead of the improved one(recommended)
+* SAS software provides both Kenward-Roger degree of freedom
+* `lmerTestR` suggests that "The culprit is the calculation of the scaling of the $F$-value for which **pbkrtest** does not appear to export a low-level (direct) method."
+which leads to slow speed which we can improve
+
+# Prototypes
+
+## mathmatical details for KR1
+
+### adjusted variance-covariance matrix
```
danielinteractive, line 65: So today I started reading the 1997 paper. I stopped after finishing Section 2 which describes this. I think we first need to get this one right. Afterwards we can proceed to the actual d.f. calculations further (Section 3 of the paper)

design/KR/kenward.Rmd
```diff
@@ -0,0 +1,402 @@
...
+
+Using previous notations,
+
+\[
+  \frac{\partial^2{\Phi}}{\partial{\sigma_i}\partial{\sigma_j}} = \Phi (P_i \Phi P_j + P_j \Phi P_i - Q_{ij} - Q_{ji} + R_{ij}) \Phi
+\]
+where
+\[
+  R_{ij} = X^\top\Sigma^{-1}\frac{\partial^2\Sigma}{\partial{\sigma_i}\partial{\sigma_j}} \Sigma^{-1} X
+\]
```
danielinteractive, line 120: so I think this part is not needed to repeat from the paper, unless we have specific parts where we want to adapt the notation to what we use in the model fitting vignette

design/KR/kenward.Rmd
```diff
@@ -0,0 +1,402 @@
...
+
+(Note: in some implementations of KR, $R_{ij}$ is ignored and leads to a variant.)
```
danielinteractive, line 122: this is interesting. So in SAS is it ignored or not? do we know?

design/KR/kenward.Rmd
```diff
@@ -0,0 +1,402 @@
...
+
+\[
+  \hat\Phi_A = \hat\Phi + 2\hat\Phi \left\{\sum_{i=1}^r{\sum_{j=1}^r{W_{ij}(Q_{ij} - P_i \hat\Phi P_j - \frac{1}{4}R_{ij})} }\right\} \hat\Phi
+\]
+
+$W$ can be estimated using the inverse of information matrix by substituting the observed values
+
+\[
+  2W_{ij} = tr(\frac{\partial{\Sigma^{-1}}}{\partial \sigma_i} \Sigma \frac{\partial{\Sigma^{-1}}}{\partial \sigma_j} \Sigma) - tr(2\Phi Q_{ij} - \Phi P_i\Phi P_j)
+\]
```
danielinteractive, line 134: this approximation we actually don't need - since we get the numeric Hessian directly from TMB, so we have a good W matrix already for the variance parameter covariance

design/KR/kenward.Rmd
```diff
@@ -0,0 +1,402 @@
...
+  R_{ij} = X^\top\Sigma^{-1}\frac{\partial^2\Sigma}{\partial{\sigma_i}\partial{\sigma_j}} \Sigma^{-1} X
+\]
+
+(Note: in some implementations of KR, $R_{ij}$ is ignored and leads to a variant.)
+
+So the adjusted estimator is
+
+\[
+  \hat\Phi_A = \hat\Phi + 2\hat\Phi \left\{\sum_{i=1}^r{\sum_{j=1}^r{W_{ij}(Q_{ij} - P_i \hat\Phi P_j - \frac{1}{4}R_{ij})} }\right\} \hat\Phi
+\]
```
danielinteractive, line 128: I would start with this and then explain the ingredients (W, Q, P, R)

design/KR/kenward.Rmd
```diff
@@ -0,0 +1,402 @@
...
+
+$W$ can be estimated using the inverse of information matrix by substituting the observed values
+
+\[
+  2W_{ij} = tr(\frac{\partial{\Sigma^{-1}}}{\partial \sigma_i} \Sigma \frac{\partial{\Sigma^{-1}}}{\partial \sigma_j} \Sigma) - tr(2\Phi Q_{ij} - \Phi P_i\Phi P_j)
+\]
+
```
danielinteractive, line 135: So before jumping into coding I would like to see here more simplifications and explanations / thoughts on the adjusted cov estimator, e.g.
- explain how Sigma is block diagonal into subjects
- explain that we have the derivatives of Sigma blocks exactly based on the chosen covariance structure
- think about how we will get these exact derivatives, I think we need to do that in C++, question is whether we can use the automatic differentiation engine within TMB to avoid manual coding of this?
- explain we have W
- break down the P, Q, R calculations into sums of subject specific terms where possible (hopefully possible for all)

design/KR/kenward.Rmd
```diff
@@ -0,0 +1,402 @@
...
+### Inference and degree of freedom
```
danielinteractive, line 136: and then after we have the formulas simplified/detailed as above, then we can dive here into some prototyping. or if needed we do part of that in C++ in TMB on this branch.

design/KR/kenward.Rmd
```diff
@@ -0,0 +1,402 @@
...
+
+Suppose we are testing the $l$ linear combination of $\beta$, $L\beta$, we can use the following Wald-type statistic
```
danielinteractive, line 138: let's park the d.f. section completely until we have nailed the adjusted var-cov

danielinteractive (commented):
thanks @clarkliming for starting this, quite complex stuff. let's do step by step, see my first thoughts inside. I hope we can get the adjusted var-cov matrix from SAS e.g. to compare?

design/KR/kenward.Rmd
clarkliming: I wonder if we should include these P, Q, R, W in the C++ part, as the `df_1d` part should be separated with the `mmrm` part. I think it will be complex for `df_1d` with KR to really rely on some of the calculations in C++ side of the model fitting while `df_1d` with SAT does not; We can, somehow use C++ to do the calculations instead of R side for performance considerations. The adjusted variance-covariance should not be implemented while fitting the object, as you may change the `df_1d` methods after you get the raw fit. Instead of fitting a new model with new adjusted variance covariance matrix, having them separately should be fine

design/KR/kenward.Rmd
clarkliming: This can be moved to the vignettes part

design/KR/kenward.Rmd
clarkliming: SAS has both, KR<linear>(sub option linear) and KR, and also KR2

design/KR/kenward.Rmd
clarkliming: ok we can change the order

design/KR/kenward.Rmd
danielinteractive: Ok separate is also fine, but I think we need to do the derivative calculations with automatic differentiation in C++ because for the inverse Sigma it won't be feasible manually. Plus we need the speed

clarkliming: you are right the inverse can be slow; we can report some kind of sigma inverse in the C++ side for each subject(because subjects can have different visits); for other part, like derivatives, I think we have closed form why we need the automatic differentiation? we can complete it in C++ side (but not with AD)

design/KR/kenward.Rmd
clarkliming: yes we can use the hessian part; at first I would like some kind of "re-parameterization" to make things simpler but somehow ran into issues. Using the raw `theta` vector solve the issue. We can rely on this calculation to double check if everything is correct.

design/KR/kenward.Rmd
danielinteractive: I don't think there is a closed form for the partial derivatives of Sigma^-1, so I think we need AD there.
Also AD will be less work for future additions of covariance structures. In terms of coding this is a hard part, but I think it will be worth the effort. Might need require separate experimentation with TMB's AD mechanism to figure this out

clarkliming:
```{r}
inverse_derivative <- function(sigma_inv, dsigma) {
  ret <- lapply(dsigma, function(x) {
    -sigma_inv %*% x %*% sigma_inv
  })
  return(ret)
}
```

if sigma is invertible, we have  partial( sigma * sigma_inverse)/partial(theta) = partial(identity matrix) / partial(theta)

and partial( sigma * sigma_inverse)/partial(theta) =  partial(sigma) / partial(theta)  * sigma_inverse + sigma * partial(sigma_inverse) / partial(theta)

so partial(sigma_inverse) / partial(theta) = - sigma_inverse * partial(sigma) / partial(theta) * sigma_inverse

partial(sigma) / partial(theta) here we have closed form and as long as we get the inverse of sigma everything is fine

danielinteractive: ah cool! I did not see that. Very good. Still I wonder if for the partial derivatives we can employ AD to save writing code for each structure. Would be worth trying

clarkliming: there is auto jacobian and hessian in https://kaskr.github.io/adcomp/namespaceautodiff.html#aece87159d1d7bc9099eff6b751143db4 

it seems that we can flat the covariance matrix and the take the jacobian of it (and loop its values to take hessian which is used by R matrice)

design/KR/kenward.Rmd
clarkliming: currently the hessian and the approximation does not match. Possibly there is an error. I checked but fail to find any error. could you please have a look @danielinteractive

design/KR/kenward.Rmd
danielinteractive: @clarkliming I would start with this steps here

design/KR/kenward.Rmd
danielinteractive: But it is just an approximation. I think we don't need to dig more into it because we have a nice W anyway

clarkliming: it could somehow reveal errors in the derivatives. In the approximated W, only the 2,2 element does not match largely.

danielinteractive: Yeah but I trust TMB there , no worries

clarkliming:
note: `devtools::load_all` will overwrite the current Rcppexports file (which we don't want it to happen because we still have TMB exports).

we have to use `pkgbuild::compile_dll(compile_attributes = F)` first and then call `devtools::load_all()`.

danielinteractive:
thanks @clarkliming , so I changed now a bit the file names and we can take the manual route of exporting C++ functions now, so we can still use devtools etc. in RStudio as well now.

danielinteractive (commented):
thanks @clarkliming , I just ran that small part of the design doc and nice to see that the results are the same between the C++ and the R parts.
I think now is a good time to clean up the PR as you describe above as well.
I would highly recommend in parallel to write nicely the KR vignette following the algorithm vignette notation (and extending the notation where needed), and clean up the code. That will make review and maintenance possible

R/satterthwaite.R
```diff
@@ -296,7 +296,7 @@ h_df_md_from_1d <- function(object, contrast) {
 #'   elements for singular coefficient estimates, i.e. only refer to the
 #'   actually estimated coefficients.
 #'
-#' @return List with `est`, `se`, `df`, `t_stat` and `p_val`.
+#' @return List with `num_df`, `denom_df`, `f_stat` and `p_val` (2-sided p-value).
```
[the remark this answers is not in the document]
clarkliming: no; but it is a minor issue spotted (df_md is the f test not the t test statistic)

clarkliming:
@danielinteractive vignette updated; for the wrapping, as we have discussed, we can put it in `mmrm` function (and we can skip the parameter checking this pqr function because in mmrm the arguments should be valid already)

shall we make another issue of "implementation of KR"?

Currently, grouped and weighted KR are not there yet (grouped need minor modification, while weighted takes more changes)

danielinteractive:
@clarkliming thanks for the vignette work, that is great. 
So we have 2 options:
1) merge this PR as a design work first. In that case, please move the vignette out of vignettes into design folder for now. Since vignettes should only contain actual features of the package that already work.
2) further work on this PR for implementation in the package.

I would probably go for option 1). In that case, it would be good to describe how the implementation would handle the grouped and the weighted covariance matrices. My thinking would naively be that for the weighted version, the weights can be multiplied relatively easily on the current Cholesky derivatives?

danielinteractive (commented):
Thanks @clarkliming for the important and complex work on the KR stuff. Really amazing.
So there is a lot of files touched in this PR, and I would not like to merge all of this just for the design step. Can we do like this:
1) make a new PR with just the one theoretical design doc `kenward.Rmd` then I will just review that (maybe I find some time over the weekend too)
2) keep this branch, keep up to date via merging main into it, and use this branch and PR to implement the KR - then it does not hurt that we have all the additional files touched etc.

danielinteractive:
@clarkliming Do we still need this PR?

clarkliming:
> @clarkliming Do we still need this PR?

I think we can close this PR for now; but let's keep the branch and after implementation is merged we are fine to remove the branch

PR #155, opened 2022-10-29
deprecate free_cores()

### LINKED ISSUE #146: Deprecate `free_cores()`
Note that this might likely also lead to improvements of the function itself.

To do:

- [ ] Evaluate https://cran.r-project.org/web/packages/parallelly/index.html in particular its function `freeCores()`
- [ ] if that looks ok we can just drop our `free_cores()` from this package completely
- [ ] which means to first deprecate it and later remove it
- [ ] in which case just need to refactor downstream dependencies `tern.mmrm` and `teal.modules.clinical`

kkmann:
closes #146 

Our `free_cores()` function is not really necessary and can be replaced by `parallelly::availableCores(omit = 1)`. `parallelly` is widely used, well maintained, hence no problem to depend on (optionally!). Removing non-core functionality makes the package easier to maintain in the long term.


NEWS.md
```diff
@@ -1,5 +1,9 @@
 # mmrm 0.1.5.9000
 
+- deprecated `free_cores()` in favor of `parallelly::availableCores(omit = 1)`
```
danielinteractive, line 3: Please use sentence style (see entries below).

tests/testthat/test-parallel.R
```diff
@@ -1,5 +0,0 @@
-test_that("free_cores works as expected", {
-  skip_on_cran()
-
-  expect_silent(free_cores())
```
danielinteractive, line 4: Please don't delete already, instead please check that the deprecation warning works.
(because the function is still in the package we still need to test it)

R/utils.R
```diff
@@ -68,6 +69,7 @@ h_record_all_output <- function(expr, remove = list()) {
 #'
 #' @export
 free_cores <- function() {
+  .Deprecated("parallelly::availableCores(omit = 1)")
```
danielinteractive, line 72: suggest to follow https://lifecycle.r-lib.org/articles/manage.html so `lifecyle::deprecate_warn()` here instead

danielinteractive (commented):
thanks @kkmann ! please see comments inside, in addition, can you please adjust the `pkgdown` yaml such that we have a "Deprecated" section now?

R/utils.R
kkmann: ok, I think I got it in and adjusted the test

tests/testthat/test-parallel.R
kkmann: replaced with deprecation test

NEWS.md
kkmann: x) done

R/utils.R
kkmann: btw, I am prospectively setting the depreciation version to 0.1.6 since we are in 0.1.5.9xxx now, correct?

kkmann:
@danielinteractive should be good to go now, hope I cought everything / see comments.

NAMESPACE
```diff
@@ -25,6 +25,7 @@ export(mmrm_control)
 export(refit_multiple_optimizers)
 import(checkmate)
 importFrom(Rdpack,reprompt)
+importFrom(lifecycle,deprecate_warn)
```
danielinteractive, line 28: no need to import this - just use the `::` operator. we just always import a single function (I think that is also in the dev guide)

tests/testthat/test-free_cores.R
```diff
@@ -0,0 +1,7 @@
+test_that("free_cores throws deprecation warning", {
+  skip_on_cran()
+  expect_warning(
+    free_cores(),
+    regexp = "`free_cores\\(\\)` was deprecated in mmrm 0\\.1\\.6."
```
danielinteractive, line 5: consider `fixed = TRUE` then you don't need the escapes

NAMESPACE
kkmann: If so I didn't see it. What do you mean with importing a single function? My reference would be https://kbroman.org/pkg_primer/pages/depends.html since we do not use a lot of the lifecycle package @importFrom is clearer and makes it easier to clean up the namespace after removing the deprecation has been removed. Anyway, not really important to me - @import or @importFrom?

tests/testthat/test-free_cores.R
kkmann: I am happy with escapes.

kkmann:
@danielinteractive can you indicate which points are resolved and review please? Don't want to close comments I didn't open.

tests/testthat/test-free_cores.R
danielinteractive: hm I would prefer without :-)

NAMESPACE
danielinteractive: Hm ok we need to add that to the contributing guide.
Basically no need to import any more here. Just use `lifecycle::deprecate_warn()` in the code.

R/utils.R
danielinteractive: perfect

tests/testthat/test-free_cores.R
BrianLang: It's definitely more of a fixed string match rather than a regex, isn't it?

kkmann: I just find it weird to use an argument calle "regex" that is not interpeted as a regex but all fine by me, changed it.

NAMESPACE
BrianLang: @danielinteractive, Can we have a description / comment on this choice? We use `lifecycle::` in the header but don't mention this? Is this OK mainly because `.Rd` are compiled by hand, separately from the functioning of the package?

tests/testthat/test-free_cores.R
BrianLang: do a PR to `testthat` for a rename of the argument ;-)

danielinteractive: @BrianLang haha :-) that will be difficult!

NAMESPACE
danielinteractive: @BrianLang yeah we will add this to the dev guide

danielinteractive (approved):

kkmann:
@danielinteractive I think the CI/CD is waiting for something

Issue #157, opened 2022-10-29
Some improvements to the documentation


kkmann:
- [x] CRAN badge 
- [x] Cumulative CRAN downloads badge / monthly
- [x] ~lifecycle badge~ overlaps too much with repo status. Not the biggest fan of either but ok.
- [x] build/check status badge
- [ ] test coverage badge
- [x] link detailed info directly in the readme, no need to show code to build vignettes

kkmann:
How do we get a test coverage badge for main? Could switch to codeov or covralls?

danielinteractive:
@cicdguy any idea for the test coverage badge that @kkmann refers to ?

kkmann:
@cicdguy the coverage badge does not display correclty yet for me but if I follow the link it displays the svg correctly - any ideas? Does it work for yo? https://raw.githubusercontent.com/openpharma/mmrm/_xml_coverage_reports/data/main/badge.svg

kkmann:
sry, too impatient - working now, great and thx again!
![image](https://user-images.githubusercontent.com/2482236/199489249-771c472f-60cf-4708-9924-dcdb00afaf03.png)

PR #159, opened 2022-10-31
Produce coverage info for main

cicdguy:
Coverage info for main will include:

- [x] Diff reports against main
- [x] A code coverage badge

Relates to #157


kkmann:
@cicdguy a) cool name b) thx for looking into this so quickly c) how do we then display the badge in the README.md, as in where is the badge put? Can you point me to the relevant bit in the GH action?

danielinteractive (approved):

PR #176, opened 2022-11-18
fix weights overwrite

### LINKED ISSUE #171: Avoid assignment of `weights` in environment as it can overwrite global environment variable
there is a side effect that `weights` is assigned to the global environment. If this argument is used by the user, the "weights" in global environment can be changed accidentally.

image

The reason is that https://github.com/openpharma/mmrm/blob/88836158c2a71826ee98fbe997a80654ad961f49/R/tmb.R#L258

Usually the formula's environment is "global env" so can cause issues.

I wonder if we should, instead of assigning weights to the environment, we should remove any environment of that formula?

To do:

- [x] Add unit test that reproduces overwriting of `weights` in global environment
- [x] Adapt code that this does not happen anymore
- [x] Make sure all tests pass
- [x] Add `NEWS` entry under Bug fixes

clarkliming:
close #171

by the way, we should skip the "parallelly" tests if not installed, since parallelly is in suggest


NEWS.md
```diff
@@ -7,7 +7,9 @@
   unique time points were available for any subject. Now it is checked that there are
   no duplicate time points per subject, and this is explained also in the
   function documentation and the introduction vignette.
-
+- Previously in `mmrm` calls, the `weights` object in the environment where formula is
+  defined will be replaced by the weight used. Now this behaviour is removed and your variable
```
clarkliming, line 11: is `behaviour` misspelled? I saw spell check is complaining...

BrianLang: if we're using American English spellings, it'd be behavior. Otherwise, behaviour is fine (British English). Typically we'd just strive for consistency across the documentation. @danielinteractive, have we been having this is mind?

clarkliming: spelling updated

danielinteractive: I think we should use American English. Not sure if we can define that for the spellcheck @cicdguy any tips?

tests/testthat/test-tmb.R
```diff
@@ -376,6 +376,21 @@ test_that("h_mmrm_tmb_data catches case with multiple time points per subject ea
   )
 })
 
+
+test_that("h_mmrm_tmb_data has no side effect of overwrite the weights in global env", {
+  weights <- c("this is global weights")
```
danielinteractive, line 381:
```suggestion
  weights <- "this is global weights"
```

tests/testthat/test-tmb.R
```diff
@@ -376,6 +376,21 @@ test_that("h_mmrm_tmb_data catches case with multiple time points per subject ea
...
+  res <- h_mmrm_tmb_data(
+    formula_parts,
+    data = fev_data,
+    weights = rep_len(1, nrow(fev_data)),
+    reml = TRUE,
+    accept_singular = TRUE
+  )
+  expect_identical(weights, "this is global weights")
+})
+
```
danielinteractive, line 393: if sticking to current solution, please add test that checks that
1) `weights` in incoming data is not allowed
2) unless `weights` in incoming data is the same as specified as weights

R/tmb.R
```diff
@@ -262,8 +262,7 @@ h_mmrm_tmb_data <- function(formula_parts,
   }
   data <- data[data_order, ]
   weights <- weights[data_order]
-
-  assign("weights", weights, envir = environment(formula_parts$full_formula))
+  data$weights <- weights
```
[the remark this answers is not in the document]
clarkliming: here I updated a little bit to make sure the weights will not be overwrite. Now I think we do not need more documentations: `data` is not used after here.

R/tmb.R
```diff
@@ -262,8 +262,12 @@ h_mmrm_tmb_data <- function(formula_parts,
   }
   data <- data[data_order, ]
   weights <- weights[data_order]
-  data$weights <- weights
-  full_frame <- droplevels(stats::model.frame(formula_parts$full_formula, data = data, weights = weights))
+  data <- data.frame(data, weights)
+  # weights is always the last column
```
danielinteractive, line 266: nice! I would still add a test to show that this does not meddle an existing `weights` column in the `data`

clarkliming: added now

danielinteractive (approved):

BrianLang:
Seems like the update to staged.dependencies (recently merged pr) has led the build check to fail.

NEWS.md
danielinteractive: perfect thx @cicdguy

danielinteractive: hm but we have that setting already. not sure why it was not caught by spellcheck then

PR #182, opened 2022-11-22
implementations of KR

### LINKED ISSUE #180: implement KR1 method for covariance calculation/testing
implement KR1 method

- [x] read through KR design doc
- [x] implement C++ side P,Q, R calculation
- [x] add support for weighting/groups
- [x] move sat/kr under mmrm
- [x] add argument in mmrm to conduct testing
- [x]  add unit testings
- [x] use fit as parameter for get_PQR(get_KR_component)
- [x] add integration tests compared with SAS
- [x] Add `NEWS` entry (under "New Features")
- [x] Adjust style for Roxygen documentation
	- [x] title style for titles, 
	- [x] blank lines for readability between sections
	- [x] parentheses for type with `\cr` afterwards
	- [x] At least one sentence description before arguments
- [x] add minimum assertions for inputs also for internal helper functions to make code defensive
- [x] make sure to keep lifecycle badges, and have them for new exported functions too.
- [x] prefix internal functions with `h_`
- [x] have unit tests for all functions (incl. internal, helper functions)

clarkliming:
close #180 

currently completed task

1. group and weighted mmrm. (for group, the autodiff is use for all parameters. the reason is that it is simpler here. We can refine it later
2. wrapper around c++ side function

others to be implemented soon


danielinteractive (commented):
Nice start thanks Liming! I would recommend including the C++ level unit tests already now to be sure that it works as expected

danielinteractive:
@clarkliming just heard from James Roger (because he will talk in the same session soon) :-) we need to be careful to require REML to be used when asking for Kenward-Roger, that can be an assertion in `mmrm()` and documented clearly for the user.

clarkliming:
currently, CS and AR1 matches with SAS well, however AR1H and CSH does not. investigating

danielinteractive:
Thanks @clarkliming ! Sounds great. Regarding integration tests vs SAS as next step please use similar code format as in other files and compare with tolerance vs SAS values

src/kr_comp.cpp
```diff
@@ -0,0 +1,104 @@
...
+  bool is_spatial = is_spatial_int == 1;
+  int n_groups = mmrm_fit["n_groups"];
+  IntegerVector subject_groups = mmrm_fit["subject_groups"];
+  NumericVector weights_vector = mmrm_fit["weights_vector"];
+  if (is_spatial) {
+    Rf_error("Spatial Kenward-Roger not implemented yet!");
+  }
+  auto theta_v = as_vector(theta);
+  std::string covtype_str = string(cov_type);
+  auto G_sqrt = as_vector(sqrt(weights_vector));
```
clarkliming, line 60: the square root issue is here. at first I use:
```{cpp}
auto G = as_vector(weights_vector);
auto Gsqrt = sqrt(G);
// auto Gsqrt = G.sqrt();
// auto Gsqrt = G.cwiseSqrt();
```

it gives seems uninitialized value on my computer, instead of the square root.

danielinteractive: hm interesting. maybe worth trying to isolate this in a reproducible toy example

clarkliming: I tested a bit on total isolation and it seems working fine 
```{cpp}
List my_sqrt(NumericVector x) {
  auto x1 = as_vector(x);
  auto x2 = x1.sqrt();
  auto x3 = x1.cwiseSqrt();
  auto x4 = sqrt(x1);
  return List::create(as_nv(x2), as_nv(x3), as_nv(x4));
}
```

need more investigation here

danielinteractive: maybe the `as_vector` needs to copy / allocate memory or something like that?

clarkliming:
minor update:  `ANTE(1)` in SAS corresponds to `adh` in mmrm. I made a mistake to compare `ad` with `ANTE(1)` which lead to issues. Currently it just works fine with degree of freedom, minor discrepancy with covariancem

R/fit.R
```diff
@@ -230,6 +230,7 @@ mmrm_control <- function(optimizer = stats::nlminb,
 #' @param optimizer (`string`)\cr optimizer to be used to generate the model.
 #' @param n_cores (`count`)\cr number of cores which could in principle be used for
 #'   parallel computations on Linux or Mac machines.
+#' @param ddfm (`string`)\cr denominator of degree of freedom method.
```
danielinteractive, line 233:
```suggestion
#' @param method (`string`)\cr method used for adjusting degrees of freedom and covariance estimates.
```
or similar (easier name of argument, and it is not just for degrees of freedom)

R/fit.R
```diff
@@ -273,10 +274,14 @@ mmrm <- function(formula,
...
-                 accept_singular = TRUE) {
+                 accept_singular = TRUE,
+                 ddfm = c("Satterthwaite", "Kenward-Roger")) {
   assert_string(optimizer)
   use_automatic <- identical(optimizer, "automatic")
-
+  ddfm <- match.arg(ddfm)
+  if (ddfm == "Kenward-Roger" && reml != TRUE) {
+    stop("Kenward-Roger only works for REML!")
+  }
```
danielinteractive, line 284:
```suggestion
  if (method == "Kenward-Roger" && !reml) {
    stop("Kenward-Roger only works for REML!")
  }
```

R/get_kr_comp.R
```diff
@@ -23,27 +23,28 @@ h_get_kr_comp <- function(tmb_data, theta) {
 #' fit <- mmrm(FEV1 ~ ARMCD + ar1(AVISIT | USUBJID), data = fev_data)
 #' contrast <- matrix(c(0,1), ncol = 1)
 #' kr(fit, contrast)
-kr <- function(fit, contrast) {
+kr <- function(fit, contrast, order = 2L) {
```
danielinteractive, line 26: what is `order`? do we support another order than 2?

R/testing.R
```diff
@@ -0,0 +1,66 @@
...
+#' contrast[3] <- 1
+#' df_1d(object, contrast)
+df_1d <- function(object, contrast) {
+  assert_class(object, "mmrm")
+  assert_numeric(contrast, any.missing = FALSE)
+  contrast <- as.vector(contrast)
+  if (object$ddfm == "Satterthwaite") {
+    return(df_1d_sat(object, contrast))
+  } else if (object$ddfm == "Kenward-Roger") {
+    return(kr(object, matrix(contrast, nrow = 1)))
```
danielinteractive, line 28:
```suggestion
    df_1d_sat(object, contrast)
  } else if (object$ddfm == "Kenward-Roger") {
    kr(object, matrix(contrast, nrow = 1))
```

R/testing.R
```diff
@@ -0,0 +1,66 @@
...
+  assert_class(object, "mmrm")
+  assert_numeric(contrast, any.missing = FALSE)
+  if (!is.matrix(contrast)) {
+    contrast <- matrix(contrast, ncol = length(contrast))
+  }
+  assert_matrix(contrast, ncols = length(component(object, "beta_est")))
+  if (object$ddfm == "Satterthwaite") {
+    return(df_md_sat(object, contrast))
+  } else if (object$ddfm == "Kenward-Roger") {
+    return(kr(object, matrix(contrast, nrow = 1)))
```
danielinteractive, line 64: same here

R/get_kr_comp.R
```diff
@@ -0,0 +1,115 @@
+#' obtain kr component
```
danielinteractive, line 1: Please use conventional formatting for documentation, in particular title style for title, see other roxygen chunks for reference (also blank lines to make it easier to read)

R/get_kr_comp.R
```diff
@@ -0,0 +1,115 @@
...
+#' @param tmb_data (`mmrm_tmb_data`)\cr produced by [h_mmrm_tmb_data()].
+#' @param theta numeric vector of the theta estimate.
+#' @details the function retuns a named list, $P$, $Q$ and $R$, which corresponds to the
```
danielinteractive, line 4: maybe good to double check that math can be included like this or whether we need `\eqn` or something like that

NAMESPACE
```diff
@@ -20,10 +20,13 @@ export(df_md)
 export(fit_mmrm)
 export(fit_single_optimizer)
 export(free_cores)
+export(kr)
 export(mmrm)
 export(mmrm_control)
 export(refit_multiple_optimizers)
+export(tr)
```
danielinteractive, line 27: would recommend to not export `tr` as this is internal helper, also would prefix with `h_`

NAMESPACE
```diff
@@ -20,10 +20,13 @@ export(df_md)
 export(fit_mmrm)
 export(fit_single_optimizer)
 export(free_cores)
+export(kr)
```
danielinteractive, line 23: since this is exported, have slightly longer / more expressive function name?

R/get_kr_comp.R
```diff
@@ -0,0 +1,115 @@
...
+  if (fit$tmb_data$reml != 1) {
+    stop("Kenward-Roger is only for REML!")
+  }
+  kr_comp <- fit$kr_comp
+  w <- solve(fit$tmb_obj$he(fit$theta_est))
+  v_adj <- v_a(fit$beta_vcov, w, kr_comp$P, kr_comp$Q, kr_comp$R, order = order)
+  df <- h_kr_df(fit$beta_vcov, contrast, w, kr_comp$P)
+  f_statistic <- 1 / nrow(contrast) * fit$beta_est %*% t(contrast) %*% solve(contrast %*% v_adj %*% t(contrast)) %*% contrast %*% fit$beta_est
+  f_star <- f_statistic * df$m / (df$m - 1 + nrow(contrast))
+  return(list(df = df, v_adj = v_adj, f_star = f_star))
```
danielinteractive, line 36:
```suggestion
  list(df = df, v_adj = v_adj, f_star = f_star)
```

R/get_kr_comp.R
```diff
@@ -0,0 +1,115 @@
...
+}
+
+#' obtain the adjusted Kenward-Roger degree of freedom
+#' @param v0 unadjusted covariance matrix
+#' @param l linear combination matrix
+#' @param w hessian matrix
+#' @param p P matrix from `h_get_kr_comp`
```
danielinteractive, line 43: even for internal functions please have `@return` documented and follow conventions for roxygen, thanks

R/get_kr_comp.R
```diff
@@ -0,0 +1,115 @@
...
+  g <- ((nl + 1) * a1 - (nl + 4) * a2) / ((nl + 2) * a2)
+  denom <- (3 * nl + 2 - 2 * g)
+  c1 <- g / denom
+  c2 <- (nl - g) / denom
+  c3 <- (nl + 2 - g) / denom
+  v_star <- 2 / nl * (1 + c1 * b) / (1 - c2 * b)^2 / (1 - c3 * b)
+  rho <- v_star / (2 * e_star^2)
+  m <- 4 + (nl + 2)  / (nl * rho - 1)
+  lambda <- m / (e_star * (m - 2))
+  return(list(m = m, lambda = lambda))
```
danielinteractive, line 75: generally no explicit return

R/get_kr_comp.R
```diff
@@ -0,0 +1,115 @@
...
+
+#' obtain the adjusted covariance matrix
+#' @param v unadjusted covariance matrix
+#' @param w hessian matrix
+#' @param p P matrix from `h_get_kr_comp`
+#' @param q Q matrix from `h_get_kr_comp`
+#' @param r R matrix from `h_get_kr_comp`
+#' @param order the order of the Kenward-Roger approximation. Only 1L or 2L supported.
+#' @keywords internal
+v_a <- function(v, w, p, q, r, order = 2L) {
```
danielinteractive, line 86: as this is internal please prefix with `h_`

R/get_kr_comp.R
```diff
@@ -0,0 +1,115 @@
...
+    r[] <- 0
+  }
+  dr <- ncol(v)
+  n_theta <- ncol(w)
+  n_groups <- n_theta / dr
+  ret <- v
+  for (gi in seq_len(n_groups)) {
+    for (gj in seq_len(n_groups)) {
+      for (i in seq_len(dr)) {
+        for (j in seq_len(dr)) {
```
danielinteractive, line 97: wow 4 levels nested loop! 
ok is this fast enough in R? would it be better to do this in Rcpp?

R/get_kr_comp.R
```diff
@@ -0,0 +1,115 @@
...
+              q[ijid:(ijid + dr - 1),] -
+              p[iid:(iid + dr - 1), ] %*% v %*% p[jid:(jid + dr - 1), ] -
+              1 / 4 * r[ijid:(ijid + dr - 1), ]
+            ) %*% v
+          }
+        }
+      }
+    }
+  }
+  return(ret)
```
danielinteractive, line 114:
```suggestion
  ret
```

R/satterthwaite.R
```diff
@@ -288,33 +273,15 @@ h_df_md_from_1d <- function(object, contrast) {
 
 #' Calculation of Satterthwaite Degrees of Freedom for Multi-Dimensional Contrast
 #'
-#' @description `r lifecycle::badge("experimental")`
```
danielinteractive, line 291: please keep lifecycle badges, and make sure to have them for all new exported functions too

R/testing.R
```diff
@@ -0,0 +1,66 @@
+#' Calculation of Satterthwaite Degrees of Freedom for One-Dimensional Contrast
```
danielinteractive, line 1: I guess this file will be removed later?

src/chol_derivative.h
```diff
@@ -0,0 +1,183 @@
+#include <Rcpp.h>
+#define INCLUDE_RCPP
+#include "utils.h"
+#include "covariance.h"
+#include "tmb_includes.h"
+
+using namespace Rcpp;
+using std::string;
+
+struct chol {
```
danielinteractive, line 10: so in C++ here would be nice to have some comments on top of each class and function to explain why it is there and approximately what it is doing.

src/chol_derivative.h
```diff
@@ -0,0 +1,183 @@
...
+  std::map<std::vector<int>, Eigen::SparseMatrix<Type>> sel_mat_cache;
+  std::string cov_type;
+  std::vector<int> full_visit;
+  int n_visits;
+  int n_theta; // theta.size()
+  vector<Type> theta;
+  matrix<Type> chol_full;
+  chols() {
+    
+  }
```
danielinteractive, line 88: correct?

src/chol_derivative.h
```diff
@@ -0,0 +1,183 @@
...
+  chols(vector<Type> theta, int n_visits, std::string cov_type): n_visits(n_visits), cov_type(cov_type), full_visit(std::vector<int>(n_visits)) {
+    this->theta = theta;
+    //this->full_visit = vector<int>(n_visits);
```
danielinteractive, line 91: please remove commented code

src/init.h
```diff
@@ -5,27 +5,4 @@
 #include <R_ext/Rdynload.h>
```
danielinteractive, line 5: do we still need this file then?

src/kr_comp.cpp
```diff
@@ -0,0 +1,104 @@
+#include <Rcpp.h>
+#define INCLUDE_RCPP
+#include "utils.h"
+#include "covariance.h"
+#include "chol_derivative.h"
+
+using namespace Rcpp;
+using std::string;
+
+vector<double> as_vector(NumericVector input) {
```
danielinteractive, line 10: this whole business of converting Rcpp types to std types would fit better in utils file I think

src/test-derivatives.cpp
```diff
@@ -0,0 +1,94 @@
...
+#include "testthat-helpers.h"
+#include <iostream>
+
+context("cho_jacobian") {
+  test_that("cho_jacobian works as expected") {
+    chol_jacobian chol_jac_obj(2, "ar1");
+    vector<double> theta {{1.0, 1.0}};
+    vector<double> result = chol_jac_obj(theta);
+    vector<double> expected(8);
+    expected << 2.718282, 1.922116, 0, 1.922116, 0.0, 0.9610578, 0.0, -0.9610578;
```
danielinteractive, line 15: do we know these values from R calculations? or is it a snapshot test?

src/test-derivatives.cpp
```diff
@@ -0,0 +1,94 @@
...
+    chol_jacobian chol_jac_obj(2, "ar1");
+    vector<double> theta {{1.0, 1.0}};
+    vector<double> result = autodiff::jacobian(chol_jac_obj,theta).vec();
+    vector<double> expected(16);
+    expected << 2.718282, 1.9221164, 0, 1.9221167, 0.0, 0.9610586, 0.0, -0.9610586, 0.0, 0.9610586, 0.0, -0.9610586, 0.0, -1.4415871, 0.0, 0.4805284;
+    expect_equal_vector(result, expected);
+  }
+}
+
+context("Rcpp and eigen conversion") {
```
danielinteractive, line 28: also then move this to utils test file

src/tmb_includes.h
```diff
@@ -1,5 +1,55 @@
...
+#ifdef R_Q_P01_boundaries
+#undef R_Q_P01_boundaries
+#endif
+#ifdef R_D_negInonint
+#undef R_D_negInonint
+#endif
+#ifdef R_D_nonint_check
+#undef R_D_nonint_check
+#endif
+#endif // ifdef WITH_RCPP
```
danielinteractive, line 52:
```suggestion
#endif // ifdef INCLUDE_RCPP
```
?

src/tmb_includes.h
```diff
@@ -1,5 +1,55 @@
 #ifndef TMB_INCLUDES_H
 #define TMB_INCLUDES_H
+#ifdef INCLUDE_RCPP
+/* Rcpp defines `R_NO_REMAP` which causes errors when using TMB. Thus, we
```
danielinteractive, line 4: hm all of this is not really nice. is there a way we could instead of doing this here contribute to `TMB` project to have this managed there?

danielinteractive (commented):
thanks a lot @clarkliming , looks great! see first comments inside

R/fit.R
clarkliming: ok; I am just thinking we can make it consistent with SAS(sas use ddfm here)

R/get_kr_comp.R
clarkliming: KR (linear); I am adding these tests currently

src/chol_derivative.h
clarkliming: this is created to enable [] methods (it require default constructor), we can remove it and instead of using `[]` we use `insert` and `get` functionalities. This is added just to create a default initializer without initializing anything

src/kr_comp.cpp
clarkliming: utils.h is a header file and I am currently putting the declaration in chol_derivative.h file. The definition need to be in cpp files (or we need another file called utility.cpp? this is different from utils.h).

src/test-derivatives.cpp
clarkliming: this is obtained from the previous design code using R. Shall we define cpp functions to obtain the derivatives, or just use something like this ? (using R to get the derivatives is easy and function is almost ready in the doc)

R/get_kr_comp.R
clarkliming: groups * number of theta iterations; it can be optimized later

R/fit.R
danielinteractive: No need to be consistent with sas here

R/get_kr_comp.R
danielinteractive: But then we just have order one or two right? Then I would say better make it a flag instead of a numeric argument

src/chol_derivative.h
danielinteractive: Ok better add a comment to explain for developers later

src/kr_comp.cpp
danielinteractive: Yeah we can have new utility.cpp

src/test-derivatives.cpp
danielinteractive: Ok this is fine then, just add a comment where the numbers come from

src/tmb_includes.h
clarkliming: need further investigation here; the Rcpp and TMB stuffs are quite complicated

danielinteractive: yeah you can file a separate issue for that then we can follow up later.

clarkliming:
hi @danielinteractive when implementing this, I really find that putting the `method` in `mmrm` is quite confusing, and we can't avoid the arguments for `df_1d` things because of `linear` option in KR ( or we can make "KR_linear" a valid option in mmrm method?)

Anyway these values can be called through other functions, the benefit to have these included in `mmrm` is minimal (users do not need to specify the method argument in `df_1d` but they need it in `mmrm`, but the `fit$beta_vcov` is still the original one instead of adjusted one.) I agree that for follow up steps having these arguments for emmeans things can be redundant, but I don't know if there is a better way to handle these things

src/test-derivatives.cpp
clarkliming: updated

src/kr_comp.cpp
clarkliming: updated

R/get_kr_comp.R
clarkliming: user `linear` as a flag argument

danielinteractive:
> hi @danielinteractive when implementing this, I really find that putting the `method` in `mmrm` is quite confusing, and we can't avoid the arguments for `df_1d` things because of `linear` option in KR ( or we can make "KR_linear" a valid option in mmrm method?)
> 
> Anyway these values can be called through other functions, the benefit to have these included in `mmrm` is minimal (users do not need to specify the method argument in `df_1d` but they need it in `mmrm`, but the `fit$beta_vcov` is still the original one instead of adjusted one.) I agree that for follow up steps having these arguments for emmeans things can be redundant, but I don't know if there is a better way to handle these things

thanks @clarkliming - so I would recommend 
1) allowing `method` to be `Satterthwaite` (default), `Kenward-Roger` or `Kenward-Roger-Linear` e.g. (or consider abbrevations for all of these, e.g. `ST`, `KR`, `KRLin`)
2) When using Kenward-Roger, we should make sure we always use the adjusted covariance matrix downstream. You can see what is the best way for that. But it would be important that when just calling `summary(fit)` that the adjusted one is used and not the unadjusted one.

clarkliming:
there are major changes in tmb, mmrm methods

1. component(x, "beta_vcov") will return x$beta_vcov for tmb or ST, return x$beta_vcov_adj for KR, KRLin
2. component(x, "method") not implemented as this component only requires "mmrm_tmb".
3. summary print will also print adjust method
4. mmrm_tmb do not have `method` attribute so printing will not print adjust
5. print.mmrm is not implemented and it will use print.mmrm_tmb so the adjust method will not be displayed
6. replace x$beta_vcov with component in nearly all places
7. checking in df_1d (now df_1d_sat) removed, and added in new df_1d (similar for df_md)
8. P, Q and R are still long matrix; we can split it into list (of list) of matrices, or it can be just fine to use it for now

R/component.R
```diff
@@ -21,7 +21,9 @@
 #' - `conv_message`: message accompanying the convergence code.
 #' - `evaluations`: number of function evaluations for optimization.
 #' - `beta_vcov`: estimated variance-covariance matrix of coefficients
-#'      (excluding aliased coefficients).
+#'      (excluding aliased coefficients). For Kenward-Roger and Kenward-Roger-Linear
```
danielinteractive, line 24:
```suggestion
#'      (excluding aliased coefficients). For Kenward-Roger
```
since Kenward-Roger-Linear is also one specific Kenward-Roger method.

R/fit.R
```diff
@@ -230,6 +230,7 @@ mmrm_control <- function(optimizer = stats::nlminb,
 #' @param optimizer (`string`)\cr optimizer to be used to generate the model.
 #' @param n_cores (`count`)\cr number of cores which could in principle be used for
 #'   parallel computations on Linux or Mac machines.
+#' @param method (`string`)\cr method used for adjusting degrees of freedom and covariance estimates.
```
danielinteractive, line 233:
```suggestion
#' @param method (`string`)\cr method used for adjusting the degrees of freedom and 
#'    potentially the coefficients variance-covariance matrix.
```

R/fit.R
```diff
@@ -273,10 +274,15 @@ mmrm <- function(formula,
                  reml = TRUE,
                  optimizer = "automatic",
                  n_cores = 1L,
-                 accept_singular = TRUE) {
+                 accept_singular = TRUE,
+                 method = c("Satterthwaite", "Kenward-Roger", "Kenward-Roger-Linear", "ST", "KR", "KRLin")) {
```
danielinteractive, line 278:
```suggestion
                 method = c("Satterthwaite", "Kenward-Roger", "Kenward-Roger-Linear")) {
```
let's go for the long method names - or for the short names. see below

R/fit.R
```diff
@@ -273,10 +274,15 @@ mmrm <- function(formula,
...
   assert_string(optimizer)
   use_automatic <- identical(optimizer, "automatic")
-
+  method <- match.arg(method)
+  method <- h_get_method(method)
```
danielinteractive, line 282: I would avoid this extra logic - either go for long names or for short names, but not allow both
```suggestion
```

R/fit.R
```diff
@@ -273,10 +274,15 @@ mmrm <- function(formula,
...
+  if (method %in% c("Kenward-Roger", "Kenward-Roger-Linear") && !reml) {
+    stop("Kenward-Roger only works for REML!")
```
danielinteractive, line 284:
```suggestion
    stop("Kenward-Roger only works for REML")
```

R/fit.R
```diff
@@ -312,10 +318,22 @@ mmrm <- function(formula,
...
+    covbeta_fun <- h_covbeta_fun(fit)
+    fit$jac_list <- h_jac_list(covbeta_fun, fit$theta_est)
+  } else {
+    kr_comp <- h_get_kr_comp(fit$tmb_data, fit$theta_est)
+    fit$kr_comp <- kr_comp
+    if (method == "Kenward-Roger-Linear") {
+      linear <- TRUE
+    } else {
+      linear <- FALSE
+    }
```
danielinteractive, line 332:
```suggestion
    linear <- (method == "Kenward-Roger-Linear")
```

R/fit.R
```diff
@@ -312,10 +318,22 @@ mmrm <- function(formula,
...
+    w <- solve(fit$tmb_obj$he(fit$theta_est))
```
danielinteractive, line 333:
```suggestion
    w <- component(fit, "theta_vcov")
```
(no need to recalculate this)

R/fit.R
```diff
@@ -312,10 +318,22 @@ mmrm <- function(formula,
...
+    v_adj <- h_var_adj(fit$beta_vcov, w, kr_comp$P, kr_comp$Q, kr_comp$R, linear = linear)
```
danielinteractive, line 334: for readability better name the arguments of `h_var_adj()` here since there are quite a few

R/fit.R
```diff
@@ -312,10 +318,22 @@ mmrm <- function(formula,
...
-  covbeta_fun <- h_covbeta_fun(fit)
-  fit$jac_list <- h_jac_list(covbeta_fun, fit$theta_est)
-
+  fit$method <- method
+  if (method == "Satterthwaite") {
+    covbeta_fun <- h_covbeta_fun(fit)
+    fit$jac_list <- h_jac_list(covbeta_fun, fit$theta_est)
+  } else {
+    kr_comp <- h_get_kr_comp(fit$tmb_data, fit$theta_est)
+    fit$kr_comp <- kr_comp
```
danielinteractive, line 327: would avoid temporary `kr_comp` here, just do this in one line and use `fit$kr_comp` in below code

R/fit.R
```diff
@@ -312,10 +318,22 @@ mmrm <- function(formula,
...
+    if (method == "Kenward-Roger-Linear") {
+      linear <- TRUE
+    } else {
+      linear <- FALSE
+    }
+    w <- solve(fit$tmb_obj$he(fit$theta_est))
+    v_adj <- h_var_adj(fit$beta_vcov, w, kr_comp$P, kr_comp$Q, kr_comp$R, linear = linear)
+    fit$beta_vcov_adj <- v_adj
```
danielinteractive, line 335: also here do this above already and avoid temporary `v_adj` variable

R/kenwardroger.R
```diff
@@ -0,0 +1,160 @@
+#' obtain Kenward-Roger adjustment component
```
danielinteractive, line 1: Please adjust style for roxygen (see previous review's comment)

R/kenwardroger.R
```diff
@@ -0,0 +1,160 @@
...
+#' @param tmb_data (`mmrm_tmb_data`)\cr produced by [h_mmrm_tmb_data()].
+#' @param theta (`numeric`)\cr numeric vector of the theta estimate.
```
danielinteractive, line 3:
```suggestion
#' @param theta (`numeric`)\cr theta estimate.
```
(no need to repeat type since it is in parentheses already)

R/kenwardroger.R
```diff
@@ -0,0 +1,160 @@
...
+#' @details the function returns a named list, \eqn{P}, \eqn{Q} and \eqn{R}, which corresponds to the
+#' paper in 1997. The matrices are stacked in columns so that \eqn{P}, \eqn{Q} and \eqn{R} has the same
+#' column number(number of beta parameters). The number of rows, is dependent on
+#' the total number of theta and number of groups, if the fit is a grouped mmrm.
+#' For \eqn{P} matrix, it is stacked sequentially. For \eqn{Q} and \eqn{R} matrix, it is stacked so
+#' that the \eqn{Q_{ij}} and \eqn{R_{ij}} is stacked from \eqn{j} then to \eqn{i}, i.e. \eqn{R_{i1}}, \eqn{R_{i2}}, etc.
+#' \eqn{Q} and \eqn{R} only contains intra-group results and inter-group results should be all zero matrices
+#' so they are not stacked in the result.
+#' @return Named list with elements:
+#' - `P`: `matrix` of P component.
```
danielinteractive, line 13:
```suggestion
#' - `P`: `matrix` of \eqn{P} component.
```
etc below to use same style as above

R/kenwardroger.R
```diff
@@ -0,0 +1,160 @@
...
+
+
+#' Calculation of Kenward-Roger Degrees of Freedom for Multi-Dimensional Contrast
+#' @param object `mmrm` object
+#' @param contrast `matrix` contrast matrix
+#' @param linear `logical`\cr whether to use linear Kenward-Roger approximation
+#' @return List with `num_df`, `denom_df`, `f_stat` and `p_val` (2-sided p-value).
+#' @keywords internal
+df_md_kr <- function(object, contrast, linear = FALSE) {
+  if (object$tmb_data$reml != 1) {
```
danielinteractive, line 29: would use `component()` here

R/kenwardroger.R
```diff
@@ -0,0 +1,160 @@
...
+    stop("Kenward-Roger is only for REML!")
```
danielinteractive, line 30:
```suggestion
    stop("Kenward-Roger is only for REML")
```

R/kenwardroger.R
```diff
@@ -0,0 +1,160 @@
...
+  }
+  kr_comp <- object$kr_comp
+  w <- solve(object$tmb_obj$he(object$theta_est))
```
danielinteractive, line 33: please don't recompute

R/kenwardroger.R
```diff
@@ -0,0 +1,160 @@
...
+  v_adj <- object$beta_vcov_adj
```
danielinteractive, line 34: use `component()` where possible, consider adding `component()` keywords where not yet possible

R/kenwardroger.R
```diff
@@ -0,0 +1,160 @@
...
+  df <- h_kr_df(object$beta_vcov, contrast, w, kr_comp$P)
```
danielinteractive, line 35: please name arguments here (for readability)

R/kenwardroger.R
```diff
@@ -0,0 +1,160 @@
...
+  vsolve <- solve(contrast %*% v_adj %*% t(contrast))
```
danielinteractive, line 36: would use our `h_quad_form_mat()` helper here

R/kenwardroger.R
```diff
@@ -0,0 +1,160 @@
...
+  f_statistic <- 1 / nrow(contrast) * object$beta_est %*% t(contrast) %*% vsolve %*% contrast %*% object$beta_est
```
danielinteractive, line 37: do in 2 steps: first contrast * beta_est, then `h_quad_form_mat()`

R/kenwardroger.R
```diff
@@ -0,0 +1,160 @@
...
+  f_star <- f_statistic * df$lambda
+  ret <- list(
+    num_df = nrow(contrast),
+    num_df = nrow(contrast),
+    denom_df = df$m,
+    f_stat = f_star[1, 1],
+    p_val = pf(f_star[1, 1], nrow(contrast), df$m, lower.tail = FALSE)
+  )
+  ret$sqrt <- sqrt(contrast %*% v_adj %*% t(contrast))
+  ret
```
danielinteractive, line 47: do in one step without defining temporary `ret`

R/kenwardroger.R
```diff
@@ -0,0 +1,160 @@
...
+}
+
+#' Calculation of Kenward-Roger Degrees of Freedom for One-Dimensional Contrast
+#' @param object `mmrm` object
+#' @param contrast contrast vector
+#' @param linear `logical`\cr whether to use linear Kenward-Roger approximation
+#' @return List with `est`, `se`, `df`, `t_stat` and `p_val`.
+#' @keywords internal
+df_1d_kr <- function(object, contrast, linear = FALSE) {
+  if (object$tmb_data$reml != 1) {
```
danielinteractive, line 57: similar comments here as in above df function

R/kenwardroger.R
```diff
@@ -0,0 +1,160 @@
...
+#' @param l linear combination matrix
+#' @param w hessian matrix
+#' @param p P matrix from `h_get_kr_comp`
+#' @return List with `m` and `lambda`
+#' @keywords internal
+h_kr_df <- function(v0, l, w, p) {
+  theta <- t(l) %*% solve(l %*% v0 %*% t(l)) %*% l
+  nl <- nrow(l)
+  thetav0 <- theta %*% v0
+  pl <- lapply(seq_len(nrow(p) / ncol(p)), function(x) {
```
danielinteractive, line 90: please add a comment line above this to help developers understand this, maybe link to formula in vignette.

R/kenwardroger.R
```diff
@@ -0,0 +1,160 @@
...
+    thetav0 %*% x %*% v0
+  })
+  a1 <- 0
+  a2 <- 0
+  for (i in seq_len(length(pl))) {
+    for (j in seq_len(length(pl))) {
+      a1 <- a1 + w[i, j] * h_tr(thetav0pv0[[i]]) * h_tr(thetav0pv0[[j]])
+      a2 <- a2 + w[i, j] * h_tr(thetav0pv0[[i]] %*% thetav0pv0[[j]])
+    }
+  }
```
danielinteractive, line 105: also here would  add a comment line above this to help developers understand this, maybe link to formula in vignette.

R/kenwardroger.R
```diff
@@ -0,0 +1,160 @@
...
+
+#' obtain the adjusted Kenward-Roger degree of freedom
+#' @param v0 unadjusted covariance matrix
+#' @param l linear combination matrix
+#' @param w hessian matrix
+#' @param p P matrix from `h_get_kr_comp`
+#' @return List with `m` and `lambda`
+#' @keywords internal
+h_kr_df <- function(v0, l, w, p) {
+  theta <- t(l) %*% solve(l %*% v0 %*% t(l)) %*% l
```
danielinteractive, line 87: please add minimum assertions here for inputs

R/kenwardroger.R
```diff
@@ -0,0 +1,160 @@
...
+#' @keywords internal
+h_var_adj <- function(v, w, p, q, r, linear = FALSE) {
+  if (linear) {
+    r <- matrix(0, nrow = nrow(r), ncol = ncol(r))
+  }
+  dr <- ncol(v)
+  n_theta <- ncol(w)
+  theta_per_group <- nrow(q) / nrow(p)
+  n_groups <- n_theta / theta_per_group
+  ret <- v
```
danielinteractive, line 138: also here a comment to link back to vignette might be helpful. or do that in roxygen chunk already

R/mmrm-methods.R
```diff
@@ -192,6 +193,7 @@ print.summary.mmrm <- function(x,
   cat("mmrm fit\n\n")
   h_print_call(x$call, x$n_obs, x$n_subjects, x$n_timepoints)
   h_print_cov(x$cov_type, x$n_theta, x$n_groups)
+  cat("Adjust:      ", x$method, "\n", sep = "")
   cat("Method:      ")
```
danielinteractive, line 197: hm so maybe we should replace `Method` here by `Inference` ? and use `Method` instead of `Adjust` above?

R/satterthwaite.R
```diff
@@ -173,29 +173,14 @@ h_df_1d_list <- function(est,
...
-#' contrast <- numeric(length(object$beta_est))
-#' contrast[3] <- 1
-#' df_1d(object, contrast)
-df_1d <- function(object, contrast) {
-  assert_class(object, "mmrm")
-  assert_numeric(contrast, any.missing = FALSE)
-
-  contrast <- as.vector(contrast)
+#' @keywords internal
+df_1d_sat <- function(object, contrast) {
```
danielinteractive, line 183: ok if this is internal then no lifecycle badge is good, but then prefix with `h_`

R/testing.R
```diff
@@ -0,0 +1,71 @@
...
+#'   data = fev_data
+#' )
+#' contrast <- numeric(length(object$beta_est))
+#' contrast[3] <- 1
+#' df_1d(object, contrast)
+df_1d <- function(object, contrast) {
+  assert_class(object, "mmrm")
+  assert_numeric(contrast, len = length(component(object, "beta_est")), any.missing = FALSE)
+  contrast <- as.vector(contrast)
+  method <- h_get_method(object$method)
```
danielinteractive, line 24: also here remove this
```suggestion
```

R/utils.R
```diff
@@ -146,3 +146,31 @@ cov_type <- c("us", "toep", "toeph", "ar1", "ar1h", "ad", "adh", "cs", "csh")
...
+  if (method %in% c("Satterthwaite", "ST")) {
+    "Satterthwaite"
+  } else if (method %in% c("Kenward-Roger", "KR")) {
+    "Kenward-Roger"
+  } else if (method %in% c("Kenward-Roger-Linear", "KRLin")) {
+    "Kenward-Roger-Linear"
+  } else {
+    stop("Unknown method!")
+  }
+}
```
danielinteractive, line 176:
```suggestion
```

src/chol_derivative.h
```diff
@@ -0,0 +1,181 @@
...
+  }
+  Eigen::SparseMatrix<Type> get_sel_mat(std::vector<int> visits) {
+    if (this->sel_mat_cache.count(visits) > 0) {
+      return this->sel_mat_cache[visits];
+    } else {
+      this->sel_mat_cache[visits] = get_select_matrix<Type>(visits, this->n_visits);
+      return this->sel_mat_cache[visits];
+    }
+  }
+  matrix<Type> get_sigma_derivative1(std::vector<int> visits) {
```
danielinteractive, line 113: please also here on top of the internal functions/methods please add comment lines to explain.

danielinteractive (commented):
Thanks @clarkliming ! Looks great. See comments inside. I also added checkboxes to the issue description.

danielinteractive:
> there are major changes in tmb, mmrm methods
> 
> 1. component(x, "beta_vcov") will return x$beta_vcov for tmb or ST, return x$beta_vcov_adj for KR, KRLin

that is good

> 2. component(x, "method") not implemented as this component only requires "mmrm_tmb".

for this one - we could as a follow up issue make `component()` a generic function and add a method for `mmrm` class that could return then the method.

> 3. summary print will also print adjust method

nice

> 4. mmrm_tmb do not have `method` attribute so printing will not print adjust

that is ok yeah

> 5. print.mmrm is not implemented and it will use print.mmrm_tmb so the adjust method will not be displayed

yep

> 6. replace x$beta_vcov with component in nearly all places

nice thanks

> 7. checking in df_1d (now df_1d_sat) removed, and added in new df_1d (similar for df_md)

good

> 8. P, Q and R are still long matrix; we can split it into list (of list) of matrices, or it can be just fine to use it for now

can be a follow up issue if the list makes the code easier / more readable / faster

R/kenwardroger.R
```diff
@@ -1,6 +1,8 @@
-#' obtain Kenward-Roger adjustment component
+#' Obtain Kenward-Roger Adjustment Component
```
danielinteractive, line 1:
```suggestion
#' Obtain Kenward-Roger Adjustment Components
```

R/kenwardroger.R
```diff
@@ -9,98 +11,118 @@
...
   }
   kr_comp <- object$kr_comp
-  w <- solve(object$tmb_obj$he(object$theta_est))
+  w <- component(object, "theta_vcov")
   v_adj <- object$beta_vcov_adj
-  df <- h_kr_df(object$beta_vcov, contrast, w, kr_comp$P)
-  vsolve <- solve(contrast %*% v_adj %*% t(contrast))
-  f_statistic <- 1 / nrow(contrast) * object$beta_est %*% t(contrast) %*% vsolve %*% contrast %*% object$beta_est
+  df <- h_kr_df(v0 = object$beta_vcov, l = contrast, w = w, p = kr_comp$P)
+  vsolve <- solve(h_quad_form_mat(contrast, v_adj))
```
danielinteractive, line 43:
```suggestion
  prec_contrast <- solve(h_quad_form_mat(contrast, v_adj))
```
since this is the precision matrix (inverse variance matrix) of the contrast

R/kenwardroger.R
```diff
@@ -9,98 +11,118 @@
...
+  beta_contrast <- component(object, "beta_est") %*% t(contrast)
```
danielinteractive, line 44:
```suggestion
  contrast_est <- component(object, "beta_est") %*% t(contrast)
```
since this is the estimate of the contrast

R/kenwardroger.R
```diff
@@ -9,98 +11,118 @@
...
-  if (object$tmb_data$reml != 1) {
+h_df_1d_kr <- function(object, contrast, linear = FALSE) {
+  if (component(object, "reml") != 1) {
     stop("Kenward-Roger is only for REML!")
   }
   assert_numeric(contrast, len = length(component(object, "beta_est")))
   est <- sum(contrast * component(object, "beta_est"))
-  kr_comp <- object$kr_comp
-  w <- solve(object$tmb_obj$he(object$theta_est))
+  w <- component(object, "theta_vcov")
```
danielinteractive, line 71: maybe nicer to avoid this temporary variable and instead just directly do this in the `h_kr_df()` call

R/kenwardroger.R
```diff
@@ -9,98 +11,118 @@
...
   v_adj <- object$beta_vcov_adj
```
danielinteractive, line 72: same here, would directly do this in the `h_quad_form_vec()` call below instead of having this temp var

R/kenwardroger.R
```diff
@@ -9,98 +11,118 @@
...
-  ret
 }
 
-#' obtain the adjusted Kenward-Roger degree of freedom
+#' Obtain the Adjusted Kenward-Roger Degree of Freedom
+#'
 #' @param v0 unadjusted covariance matrix
 #' @param l linear combination matrix
 #' @param w hessian matrix
 #' @param p P matrix from `h_get_kr_comp`
```
danielinteractive, line 93: please follow the convention with the argument types

R/kenwardroger.R
```diff
@@ -9,98 +11,118 @@
...
   )
-  t_stat <- abs(est) / se
-  ret$t_stat <- t_stat
-  ret$p_val <- pt(t_stat, df$m, lower.tail = FALSE) / 2
-  ret
 }
 
-#' obtain the adjusted Kenward-Roger degree of freedom
+#' Obtain the Adjusted Kenward-Roger Degree of Freedom
+#'
```
danielinteractive, line 89: please add short sentence what this is doing (convention)

R/kenwardroger.R
```diff
@@ -9,98 +11,118 @@
...
 #' @param v0 unadjusted covariance matrix
 #' @param l linear combination matrix
 #' @param w hessian matrix
 #' @param p P matrix from `h_get_kr_comp`
+#'
 #' @return List with `m` and `lambda`
```
danielinteractive, line 95:
```suggestion
#' @return List with `m` and `lambda`.
```
please also explain what is `m` and `lambda` e.g. in parentheses after them

R/kenwardroger.R
```diff
@@ -118,40 +140,51 @@ h_kr_df <- function(v0, l, w, p) {
...
 
-#' obtain the adjusted covariance matrix
+#' Obtain the Adjusted Covariance Matrix
+#'
 #' @param v unadjusted covariance matrix
 #' @param w hessian matrix
 #' @param p P matrix from `h_get_kr_comp`
 #' @param q Q matrix from `h_get_kr_comp`
 #' @param r R matrix from `h_get_kr_comp`
 #' @param linear `logical`\cr whether to use linear Kenward-Roger approximation
```
danielinteractive, line 150: same here

R/kenwardroger.R
```diff
@@ -118,40 +140,51 @@ h_kr_df <- function(v0, l, w, p) {
...
+#'
 #' @return matrix of adjusted covariance matrix
```
danielinteractive, line 152: Please use sentence style here.

R/utils.R
```diff
@@ -148,29 +148,16 @@ cov_type <- c("us", "toep", "toeph", "ar1", "ar1h", "ad", "adh", "cs", "csh")
 cov_type_spatial <- c("sp_exp")
 
 
-#' trace of a matrix
+#' Trace of a Matrix
+#'
 #' @param x square matrix input
+#'
 #' @return the trace of the square matrix
```
danielinteractive, line 155: sentence style please

R/utils.R
```diff
@@ -148,29 +148,16 @@ cov_type <- c("us", "toep", "toeph", "ar1", "ar1h", "ad", "adh", "cs", "csh")
 cov_type_spatial <- c("sp_exp")
 
 
-#' trace of a matrix
+#' Trace of a Matrix
+#'
 #' @param x square matrix input
```
danielinteractive, line 153: type please

danielinteractive (commented):
thanks @clarkliming , see few additional comments inside. Something I did not check yet in my review but what would be very important is that you have also unit tests for all internal helper functions, can you please double check that?

tests/testthat/test-krcomp.R
```diff
@@ -71,3 +71,50 @@ test_that("kr give similar results as SAS", {
   expect_equal(res$df, expected[2], tolerance = 1e-3)
   expect_equal(res$se, expected[1], tolerance = 1e-3)
 })
+
+# h_df_1d_kr
+test_that("h_df_1d_kr works as expected in the standard case", {
+  object_mmrm_kr <- get_mmrm_kr()
+  expect_snapshot(h_df_1d_kr(object_mmrm_kr, c(0, 1), TRUE))
```
danielinteractive, line 78: hm this is not ideal, as this will test for exact identity I think, can you check that and consider using `expect_snapshot_value()` instead or coding the expectations explicitly? (that would be better re: readability of the tests)

tests/testthat/test-krcomp.R
```diff
@@ -71,3 +71,50 @@ test_that("kr give similar results as SAS", {
   expect_equal(res$df, expected[2], tolerance = 1e-3)
   expect_equal(res$se, expected[1], tolerance = 1e-3)
 })
+
+# h_df_1d_kr
```
danielinteractive, line 75:
```suggestion
# h_df_1d_kr ---
```
and elsewhere as well. see conventions in other files. this helps with navigating large tests files. thanks!

tests/testthat/test-testings.R
```diff
@@ -0,0 +1,22 @@
...
+  contrast_md <- matrix(0, nrow = 2, ncol = length(object$beta_est))
+  contrast_md[2, 1] <- 1
+  contrast_md[1, 3] <- 1
+  expect_identical(df_md(object, contrast_md), h_df_md_sat(object, contrast_md))
+
+  object_kr <- get_mmrm_kr()
+  contrast <- rep(0, length(object_kr$beta_est))
+  contrast[2] <- 1
+  expect_identical(df_1d(object_kr, contrast), h_df_1d_kr(object_kr, contrast))
+
```
danielinteractive, line 17: better split up into multiple tests and give specific test descriptions

R/kenwardroger.R
```diff
@@ -25,9 +25,9 @@ h_get_kr_comp <- function(tmb_data, theta) {
 
 #' Calculation of Kenward-Roger Degrees of Freedom for Multi-Dimensional Contrast
 #'
-#' @param object `mmrm` object
-#' @param contrast `matrix` contrast matrix
-#' @param linear `logical`\cr whether to use linear Kenward-Roger approximation
+#' @param object (`mmrm`)\cr object
+#' @param contrast (`matrix`)\cr contrast matrix
+#' @param linear (`logical`)\cr whether to use linear Kenward-Roger approximation
```
danielinteractive, line 30:
```suggestion
#' @param object (`mmrm`)\cr object.
#' @param contrast (`matrix`)\cr contrast matrix.
#' @param linear (`logical`)\cr whether to use linear Kenward-Roger approximation.
```
another convention, we finish the argument description with full stop, thanks for doing that everywhere

clarkliming:
comments addressed now

NEWS.md
```diff
@@ -1,5 +1,12 @@
 # mmrm 0.1.5.9011
 
+### New Features
+
+- Add support for Kenward-Roger adjusted covariance and degree of freedom
+  in `mmrm` function call with argument `method`. It now accept "Kenward-Roger",
```
danielinteractive, line 6:
```suggestion
  in `mmrm` function call with argument `method`. It now accepts "Kenward-Roger",
```

NEWS.md
```diff
@@ -1,5 +1,12 @@
...
+  "Kenward-Roger-Linear" or "Satterthwaite". Subsequent `summary`, `print` or `component` will
```
danielinteractive, line 7:
```suggestion
  "Kenward-Roger-Linear" or "Satterthwaite" (which is still the default). Subsequent `summary`, `print` or `component` calls will
```

NEWS.md
```diff
@@ -1,5 +1,12 @@
...
+  also be based on this `method` argument, e.g. `component(fit, "beta_vcov")` will return
```
danielinteractive, line 8:
```suggestion
  also be based on this `method` argument, e.g. `component(fit, "beta_vcov")` will return the
```

NEWS.md
```diff
@@ -1,5 +1,12 @@
...
+  adjusted covariance matrix for Kenward-Roger or Kenward-Roger-Linear.
```
danielinteractive, line 9:
```suggestion
  adjusted covariance matrix if a Kenward-Roger method has been used.
```

R/kenwardroger.R
```diff
@@ -0,0 +1,198 @@
+#' Obtain Kenward-Roger Adjustment Components
+#'
```
danielinteractive, line 2: Please add a one sentence description as a minimum for all functions. E.g. here:

```suggestion
#'
#' Obtains the components needed downstream for the computation of Kenward-Roger degrees of freedom.
#'
```

R/kenwardroger.R
```diff
@@ -0,0 +1,198 @@
...
+#' - `R`: `matrix` of \eqn{R} component.
+#'
+#' @keywords internal
+h_get_kr_comp <- function(tmb_data, theta) {
+  .Call(`_mmrm_get_pqr`, PACKAGE = "mmrm", tmb_data, theta)
+}
+
+
+#' Calculation of Kenward-Roger Degrees of Freedom for Multi-Dimensional Contrast
+#'
```
danielinteractive, line 27: same here and for all other functions

R/kenwardroger.R
```diff
@@ -0,0 +1,198 @@
...
+#' \eqn{Q} and \eqn{R} only contains intra-group results and inter-group results should be all zero matrices
+#' so they are not stacked in the result.
+#'
+#' @return Named list with elements:
+#' - `P`: `matrix` of \eqn{P} component.
+#' - `Q`: `matrix` of \eqn{Q} component.
+#' - `R`: `matrix` of \eqn{R} component.
+#'
+#' @keywords internal
+h_get_kr_comp <- function(tmb_data, theta) {
```
danielinteractive, line 21: please add assertions (one line each) for the arguments

R/kenwardroger.R
```diff
@@ -0,0 +1,198 @@
...
+#'
+#' @param object (`mmrm`)\cr object.
+#' @param contrast (`matrix`)\cr contrast matrix.
+#' @param linear (`logical`)\cr whether to use linear Kenward-Roger approximation.
+#'
+#' @return List with `num_df`, `denom_df`, `f_stat` and `p_val` (2-sided p-value).
+#'
+#' @keywords internal
+h_df_md_kr <- function(object, contrast, linear = FALSE) {
+  if (component(object, "reml") != 1) {
```
danielinteractive, line 36: please also here minimum assertions

R/kenwardroger.R
```diff
@@ -0,0 +1,198 @@
...
+  m <- 4 + (nl + 2)  / (nl * rho - 1)
+  lambda <- m / (e_star * (m - 2))
+  list(m = m, lambda = lambda)
+}
+
+#' Obtain the Adjusted Covariance Matrix
+#'
+#' @param v (`matrix`)\cr unadjusted covariance matrix.
+#' @param w (`matrix`)\cr hessian matrix.
+#' @param p (`matrix`)\cr P matrix from `h_get_kr_comp`.
```
danielinteractive, line 152:
```suggestion
#' @param p (`matrix`)\cr P matrix from [h_get_kr_comp()].
```
similar below. generally use markdown syntax for linking to other function docs

R/mmrm-methods.R
```diff
@@ -139,7 +140,7 @@ h_print_call <- function(call, n_obs, n_subjects, n_timepoints) {
 #'
 #' @param cov_type (`string`)\cr covariance structure abbreviation.
 #' @param n_theta (`int`)\cr number of variance parameters.
-#' @param n_groups (`int`)\cr number of groups
+#' @param n_groups (`int`)\cr number of groups.
```
danielinteractive, line 143: very good, thanks

src/kr_comp.cpp
danielinteractive: @clarkliming ok I guess we shelve this for now...

src/tmb_includes.h
danielinteractive: @clarkliming please file a follow-up issue so that we don't forget this, thanks

src/utility.cpp
```diff
@@ -0,0 +1,43 @@
+#include "utils.h"
+using namespace Rcpp;
+// conversion between Rcpp data and eigen vector/matrix.
```
danielinteractive, line 3: Did you double check that we cannot use the `Rcpp::as` and `Rcpp::wrap` functions as mentioned here: 
https://github.com/RcppCore/RcppEigen/blob/87ef9a9f8c21ddb8a4c771c912366a155270005c/vignettes/RcppEigen-Introduction.Rnw#L163
? since we already use RcppEigen anyway

tests/testthat/test-utils.R
```diff
@@ -24,3 +26,24 @@ test_that("h_record_all_outputs works as expected with no removal list given for
   messages = c("O nearly done", "oh noo"))
   expect_identical(result, expected)
 })
+
+
+# trace of matrix ---
```
danielinteractive, line 31:
```suggestion
# h_tr ----
```
I think we need four dashes so it shows up as a section in RStudio. Plus let's follow our convention here to use the tested function name

tests/testthat/test-utils.R
```diff
@@ -24,3 +26,24 @@ test_that("h_record_all_outputs works as expected with no removal list given for
...
+
+test_that("trace of a matrix works as expected", {
+  mx <- matrix(0, nrow = 3, ncol = 4)
+  expect_error(h_tr(mx), "x must be square matrix")
+  v <- c(1, 3, 2)
+  expect_equal(h_tr(diag(v)), 6)
+})
+
+# free_cors ---
```
danielinteractive, line 40:
```suggestion
# free_cores ----
```

tests/testthat/test-krcomp.R
```diff
@@ -0,0 +1,126 @@
+# get kr comp ---
```
danielinteractive, line 1:
```suggestion
# h_get_kr_comp ----
```

tests/testthat/test-krcomp.R
```diff
@@ -0,0 +1,126 @@
...
+test_that("h_get_kr_comp works as expected on grouped/ungrouped mmrm", {
+    fit <- mmrm(FEV1 ~ ARMCD + ar1(AVISIT | USUBJID), data = fev_data, reml = TRUE, method = "Kenward-Roger")
+    krcomp <- lapply(fit$kr_comp, round, digits = 3)
+    expect_snapshot(krcomp)
+    fit <- mmrm(FEV1 ~ ARMCD + ar1(AVISIT | SEX / USUBJID), data = fev_data, reml = TRUE, method = "Kenward-Roger")
```
danielinteractive, line 6: better split up in two small tests, one for grouped and one for ungrouped.

tests/testthat/test-krcomp.R
```diff
@@ -0,0 +1,126 @@
...
+
+# test adjusted covariance matrix and degree of freedom ---
+
+test_that("kr give similar results as SAS", {
+  fit <- mmrm(FEV1 ~ ARMCD + ar1(AVISIT | USUBJID), data = fev_data, method = "Kenward-Roger")
+  res <- df_1d(fit, contrast = c(0, 1))
+  expected <- c(0.95865439662225, 188.46934887972)
+  expect_equal(res$df, expected[2], tolerance = 1e-4)
+  expect_equal(res$se, expected[1], tolerance = 1e-4)
+
```
danielinteractive, line 20: please split up into separate tests with appropriate labels.

tests/testthat/test-krcomp.R
```diff
@@ -0,0 +1,126 @@
...
+  expected <- c(0.66158550758897, 162.393385281755)
+  expect_equal(res$df, expected[2], tolerance = 1e-3)
+  expect_equal(res$se, expected[1], tolerance = 1e-3)
+})
+
+# h_df_1d_kr ---
+
+test_that("h_df_1d_kr works as expected in the standard case", {
+  object_mmrm_kr <- get_mmrm_kr()
+  expect_snapshot_value(h_df_1d_kr(object_mmrm_kr, c(0, 1), TRUE), style = "deparse")
```
danielinteractive, line 81: I wonder if it is worth to add a small test helper `expect_snapshot_tolerance()` e.g. that combines the deparse style with a more relaxed tolerance e.g. 1e-4. can simplify maintenance

tests/testthat/test-krcomp.R
```diff
@@ -0,0 +1,126 @@
...
+  object_mmrm_kr <- get_mmrm_kr()
+  expect_snapshot_value(h_var_adj(
+    v = object_mmrm_kr$beta_vcov,
+    w = component(object_mmrm_kr, "theta_vcov"),
+    p = object_mmrm_kr$kr_comp$P,
+    q = object_mmrm_kr$kr_comp$Q,
+    r = object_mmrm_kr$kr_comp$R,
+    linear = TRUE
+  ), style = "deparse")
+  expect_snapshot_value(h_var_adj(
```
danielinteractive, line 118: also here would split up into two tests

tests/testthat/test-testings.R
```diff
@@ -0,0 +1,29 @@
+# df_1d ---
+
+test_that("df_1d use correct function on different method fits", {
+  object <- get_mmrm()
+  contrast <- rep(0, length(object$beta_est))
+  contrast[3] <- 1
+  expect_identical(df_1d(object, contrast), h_df_1d_sat(object, contrast))
+
```
danielinteractive, line 8: please split up, use more specific labels

tests/testthat/test-testings.R
danielinteractive: has not been addressed yet

danielinteractive (commented):
thanks @clarkliming , almost there! See comments inside. 

Please take another clean pass through everything, including the small comments applied consistently which I mention here. In particular, please add minimum assertions for all functions and have at least one sentence description in the roxygen2 part.

R/kenwardroger.R
clarkliming: in `df_1d` and `df_md` there are already assertions. shall we still duplicate the assertions here (and also h_df_md_sat) ?

src/utility.cpp
clarkliming: oh I should update the wording here; `Rcpp::as` and `Rcpp::wrap` can support the conversion between `NumericVector` and `eigen::MatrixXd` but in `TMB` are actually using `TMB::matrix` (which has its base `eigen::Array`) so we still have some difficulties here using these functions.

clarkliming:
one additional thing to mention: in unstructured covariance, in SAS we have identical results for KR and KR linear. I think the reason is in SAS's parameterization the second order derivatives (R matrix) is 0 (this is true if we directly use the $\Sigma_{ij}$ element to do the parameterization. However, in our case, our parameterization is using the Cholesky factors, which means that the second order derivatives will not be 0 any more. 

We can see 
image
which shows that UN is not included in the methods that has non-zero second order derivatives in SAS.

R/kenwardroger.R
danielinteractive: Basically in every function we need to assert the inputs that we use directly in this function. If arguments are just passed on to other function calls, no need to assert them in this function. No worries about speed when duplicating assertions, as checkmate is very fast, so that is not a concern.

danielinteractive:
> one additional thing to mention: in unstructured covariance, in SAS we have identical results for KR and KR linear. I think the reason is in SAS's parameterization the second order derivatives (R matrix) is 0 (this is true if we directly use the Σij element to do the parameterization. However, in our case, our parameterization is using the Cholesky factors, which means that the second order derivatives will not be 0 any more.
> 
> We can see image which shows that UN is not included in the methods that has non-zero second order derivatives in SAS.

Thanks @clarkliming , interesting - can you please
1) include this in the KR vignette
2) make this transparent for developers in the unstructured cov integration tests for KR (e.g. by showing that while we match SAS up to certain precision, our results are different for unstructured)
thanks!

vignettes/kenward.Rmd
```diff
@@ -313,3 +316,11 @@ $\lambda$ and $m$ can be calculated through
 
 \[m = 4 + \frac{c+2}{c\rho - 1}\]
 \[\lambda = \frac{m}{E^*(m-2)}\]
+
+## Parameterization methods and Kenward-Roger
+
+While the Kenward-Roger adjusted covariance matrix is adopting Taylor series to approximate the true value, the choices
+of parameterization can change the result. In a simple example of unstructured covariance structure, in our current approach,
+where the parameters are elements of the Cholesky factor of $\Sigma$(see [parameterization](covariance.html)), the second-order derivatives
```
danielinteractive, line 324:
```suggestion
where the parameters are elements of the Cholesky factor of $\Sigma$ (see [parameterization](covariance.html)), the second-order derivatives
```

danielinteractive (approved):

PR #184, opened 2022-12-01
Allow tests to run without `parallelly` installed

### LINKED ISSUE #177: skip the `parallelly` tests if not installed
Since `parallelly` is only in Suggests.

dgkf:
Closes #177 

The issue suggested skipping tests that required `parallelly`, but I felt like this was a useful test even if not installed and could otherwise be performed just fine.

Instead, I check whether multiple cores are available. If `parallelly` is available it is used to detect a maximum number of cores, but otherwise the tests are run on a minimum 2 cores.

It sounds like the edge cases that `parallelly` addresses are unlikely to be hit during unit testing and a more conservative fallback would be a suitable substitute. 

Tested against an install without `parallelly` installed. As far as I can tell it's just used for `availableCores` and this is the only thing that would be affected.

### DIFF
--- a/tests/testthat/test-fit.R
+++ b/tests/testthat/test-fit.R
@@ -208,6 +208,10 @@ test_that("refit_multiple_optimizers works as expected with default arguments", 
 
 test_that("refit_multiple_optimizers works with parallel computations and selected optimizers", {
   skip_on_cran()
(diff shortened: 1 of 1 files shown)


clarkliming:
I think it is really a good idea to have a edge case of two core here

danielinteractive (approved):
thanks @dgkf !

PR #194, opened 2022-12-12
Skip error-prone tests on Fedora + Clang

### LINKED ISSUE #186: Fix CRAN check problems
https://cran.r-project.org/web/checks/check_results_mmrm.html

To do:

- [x] Make test `h_summarize_all_fits works as expected` not check the numbers themselves. We have other tests that check numbers. (To address https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-fedora-clang/mmrm-00check.html and https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-debian-clang/mmrm-00check.html, so it seems this is `clang` related on Linux OS)
- [x] Write an issue for `testthat` repo to ask for advice on https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-debian-gcc/mmrm-00check.html, see https://github.com/r-lib/testthat/issues/1715 -> solved, we just need to wait for new `testthat` version on CRAN

Note that we cannot and don't need to do anything about the installed package size (as these are just notes and we have tried this before already).
We need to correct this before 2022-12-19 to safely retain our package on CRAN.

dgkf:
closes #186 

I'd still like to investigate this to constrain the test a bit more, but as a start I'm just disabling this test.

I'm a bit concerned because the test immediately below this one uses the same optimizer with all the same parameters and expects the same output, so this might just be a stochastic error which we _may_ hit again with the next test. 

If we want to be more mindful about the possibility of some model fits failing during testing, then we might want to build our own context manager that stores a tally of successful or failed model fits, skipping the first few failed fit tests and only failing outright if, for example, >5 models have been fit and >10% ever fail. This would give us some resilience to stochastic failures on CRAN without forcing us to continue skipping any test that goes haywire. 

That said, this assumes the failure to fit is stochastic, which I haven't confirmed yet.

### Other debugging steps:

- Running on `r-hub`:
   1. [Attempt \#1](https://builder.r-hub.io/status/mmrm_0.1.5.tar.gz-9be9fd9f1d4775055c9b4f6ee7fd647a) (failed due to vignette re-build error)
   2. [Attempt \#2](https://builder.r-hub.io/status/mmrm_0.1.5.tar.gz-02b82cea914d4f6bafe7ada0558a1f94) (success - no test failures)
- Going to pull `rhub/fedora-clang-devel` locally to investigate this specific test case

### DIFF
--- a/tests/testthat/test-fit.R
+++ b/tests/testthat/test-fit.R
@@ -136,6 +136,16 @@ test_that("fit_single_optimizer catches convergence warning as expected", {
 # h_summarize_all_fits ----
 
 test_that("h_summarize_all_fits works as expected", {
+  # NOTE:
+  # Test currently fails on R-devel on Fedora with clang for the first optimizer
+  # mmrm v0.1.5 @ r-devel-linux-x86_64-fedora-clang
+  #   `actual$log_liks`: -2155.280410081641 -1693.224938122514
+  #   `expected$log_liks`: -1693.224935585730 -1693.224938122510
+  #
+  # As this failure only appears on R-devel on Fedora with clang, we are
+  # temporarily skipping this test on CRAN until we can reproduce the behavior.
+  skip_on_cran()
+
   mod_fit <- fit_single_optimizer(
     formula = FEV1 ~ RACE + SEX + ARMCD * AVISIT + us(AVISIT | USUBJID),
     data = fev_data,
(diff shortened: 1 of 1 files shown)


danielinteractive:
Thanks @dgkf , to preserve more of the test, could we split this into a test that runs the function but does not compare numbers and another test that compares numbers and is skipped?

dgkf:
> Thanks @dgkf , to preserve more of the test, could we split this into a test that runs the function but does not compare numbers and another test that compares numbers and is skipped?

Sure - will do after I get off meetings for today.

danielinteractive (approved):

PR #195, opened 2022-12-13
update mmrm_control

### LINKED ISSUE #188: Add `control` argument to `mmrm()` interface to simplify it
Motivation: We already have quite a few arguments in `mmrm()` right now, and we might be getting more. E.g. #170 might add another flag. Most of these are quite advanced options that most users won't need to use often. Therefore it could be nice to simplify the interface to:
```r
mmrm <- function(
	formula,
    data,
	weights = NULL,
	reml = TRUE,
	control = mmrm_control()
) 
```
Here `reml` could still be excluded from `control` since it could be used more often. Then the `mmrm_control` interface would be
```r
mmrm_control <- function(
	optimizer = c("automatic", "L-BFGS-B", "BFGS", "CG", "nlminb"),
	n_cores = 1L,
	optimizer_fun = if (optimizer == "nlminb") stats::nlminb else stats::optim,
    optimizer_args = if (optimizer == "nlminb") list() else list(method = if (optimizer == "automatic") "L-BFGS-B" else optimizer),
	optimizer_control = if (optimizer == "nlminb") list(iter.max = 300, eval.max = 400) else list(),
	start = NULL,
	accept_singular = TRUE)
```

clarkliming:
close #188

to keep a consistent style with previous version, and to achieve that `...` go into `mmrm_control`, a lot of functions are updated.

Mainly related to `mmrm`, `fit_single_optimizer`, `refit_multiple_optimizer`, `fit_mmrm`.

here are the changes made:

1. mmrm_control will create list of "optimizers" from "optimizer" character vector
2. "optimizers" are functions with "args" attribute (to serve as additional argument for the optimizer). Another attribute "use_hessian" is also possible to allow optimizer to use "hessian" from tmb_data.
3. fit_mmrm will **only** use the first optimizer (to keep consistent on default argument of `mmrm_control` where multiple optimizers will be created)
4. refit_multiple_optimizers will use all the available optimizers instead of comparing the optimizers used in `fit`
5. custom optimizer supported

As there are relatively large change,  @danielinteractive could you please have a look first and then I will finish the other part of documentation, vignettes and tests.


R/tmb.R
```diff
@@ -595,15 +596,15 @@ fit_mmrm <- function(formula,
   args <- with(
     tmb_object,
     c(
-      list(par, fn, gr, control = control$optimizer_control),
-      control$optimizer_args
+      list(par, fn, gr),
+      attr(control$optimizers[[1]], "args")
```
danielinteractive, line 600: Here the position access feels a bit risky. Can we make this more robust?

danielinteractive (commented):
Thanks @clarkliming ! 
The interface in the mmrm function looks better now. Can you add tests how it would look like and work when user sets optimizer options?

There is new logic in here for the optimizer handling- can you read @dgkf comments on the issue and try to take those into account? 

One question I have is whether it would still be possible to call fit_mmrm with a custom optimizer function? Can we add a test for that? (There are other optimizer packages we could use in Suggests to use in test)

clarkliming:
> Thanks @clarkliming ! The interface in the mmrm function looks better now. Can you add tests how it would look like and work when user sets optimizer options?
> 
> There is new logic in here for the optimizer handling- can you read @dgkf comments on the issue and try to take those into account?
> 
> One question I have is whether it would still be possible to call fit_mmrm with a custom optimizer function? Can we add a test for that? (There are other optimizer packages we could use in Suggests to use in test)

Thank you for the quick response, I just want to first have a look at the skeleton of the current design and then implement the tests/vignettes update (in case of design changes)

For the new logic of optimizer handling, I do have read @dgkf comments, and the current "partial functions" are designed for it. However, as there are both "optimizer related arguments" and "non-optimizer related arguments" (like Satterthewaite choices, n_cores,  etc). To make things that changes at minimum, I still use a single `mmrm_control` to hold everything, including optimizers. It is now part of the `mmrm_control$optimizers` (a list containing all optimizers). Otherwise there could be larger changes to the syntax. (e.g., in `mmrm` call, how to pass the optimizers, etc). 

So here I have a solution that is a compromise of consistency and more flexible optimizer choices here.

danielinteractive:
Sounds good!
My question and the test mentioned above was more for me to see how it works now.

R/tmb.R
clarkliming: we can add an assertion to make sure at least one optimizer exists. We still use the first optimizer here. Or, do you think that `fit_mmrm` can obtain multiple "optimizer" in the same time and return multiple fits? but that can lead to down stream changes

R/fit.R
```diff
@@ -117,45 +109,45 @@ h_summarize_all_fits <- function(all_fits) {
...
+      weights = old_weights,
+      reml = fit$reml
+    ),
     mc.cores = n_cores_used,
-    mc.silent = TRUE
+    mc.silent = TRUE,
+    SIMPLIFY = FALSE
   ))
   all_fits <- c(all_fits, list(old_result = fit))
-  names(all_fits) <- c(optimizers, old_optimizer)
```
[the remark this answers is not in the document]
clarkliming: this should still work; but since it does not make much sense in "customized optimizer" things I removed this ( optimizers are now partial functions with some arguments instead of a bare string, having to name them all became a little challenging)

R/tmb.R
danielinteractive: Ok then maybe add comment or in doc to describe that first optimizer is used- and yeah assertion sounds good

clarkliming:
> Suggests

```{r}
silly_optimizer <- function(par, objective, gr, value_add, message, control, ...) {
  result <- par + value_add
  list(
    par = result,
    objective = objective(result),
    convergence = 0,
    message = message
  )
}


fit <- mmrm(
  FEV1 ~ ARMCD + ar1(AVISIT | SEX / USUBJID),
  data = fev_data,
  reml = TRUE,
  control = mmrm_control(optimizers = h_get_optimizers(optimizer_fun = silly_optimizer, optimizer_args = list(value_add = 2, message = "this is wrong")) , method = "Kenward-Roger")
)

fit$theta_est

attried_silly <- silly_optimizer
attr(attried_silly, "args") <- list(message = "this is also wrong", value_add = 1)

fit2 <- mmrm(
  FEV1 ~ ARMCD + ar1(AVISIT | SEX / USUBJID),
  data = fev_data,
  reml = TRUE,
  control = mmrm_control(optimizers = h_get_optimizers(optimizer_fun = attried_silly) , method = "Kenward-Roger")
)
fit2$theta_est
```

is this sufficient now as a demo? if not, I will add tests later under testthat

danielinteractive:
@clarkliming great - especially the `fit3` looks very easy. So green light from me to go forward toward completion of this :-)

R/fit.R
```diff
@@ -117,45 +109,45 @@ h_summarize_all_fits <- function(all_fits) {
...
   assert_class(fit, "mmrm_fit")
-  assert_int(n_cores, lower = 1L)
-  optimizers <- match.arg(optimizers, several.ok = TRUE)
-
+  assert_class(control, "mmrm_control")
   # Extract the components of the original fit.
   old_formula <- formula(fit)
   old_data <- fit$data
   old_weights <- fit$weights
-  old_optimizer <- attr(fit, "optimizer")
```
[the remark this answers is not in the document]
clarkliming: because of the fact that the label of the optimizer can be really custom, and it can be hard to infer the optimizer used, e.g. fit_single_optimizer using  "BFGS" different start/control can lead to fail, and if we change the control use refit_multiple_optimizers, it is strange that we omit the optimizer we used (as they are using different control/start or other args). So I moved the logic of "not use the already fitted" into `mmrm`, skipping the first optimizer in `mmrm`.

R/utils.R
```diff
@@ -163,3 +165,108 @@ h_tr <- function(x) {
...
   sum(diag(x))
 }
+
+#' Split Control List
+#' @description Split the [mmrm_control()] object according to its optimizers and use additional arguments
+#' to replace the elements in the original object.
+#'
+#' @param control (`mmrm_control`)\cr object.
+#' @param ... additional parameters to update the `control` object.
+#'
```
danielinteractive, line 175: please always include `@return` part

R/utils.R
```diff
@@ -163,3 +165,108 @@ h_tr <- function(x) {
...
+#' If there are other named arguments in front of these, make sure they are correctly specified through
+#' `optimizer_args`. If hessian function can be applied, please make sure its argument name is `hessian` and
+#' please add attribute `use_hessian = TRUE` to the function, using `attr(fun, "use_hessian) <- TRUE`
+#' If `optimizer_fun` is not provided, it will be created using the `optimizer` character. Other arguments should
+#' go into `optimizer_args`.
+#' Allowed are "L-BFGS-B", "BFGS", "CG"(using [stats::optim()] with corresponding method) and
+#' "nlminb"(using [stats::nlminb()]).
+#'
+#' @keywords internal
+h_get_optimizers <- function(optimizer = c("L-BFGS-B", "BFGS", "CG", "nlminb"),
```
danielinteractive, line 208: what happens if user supplies `automatic`? can we fail with a deprecation message instead of just somehow?

R/utils.R
```diff
@@ -163,3 +165,108 @@ h_tr <- function(x) {
...
+#' @param fun (`function`)\cr to be wrapped.
+#' @param ... Additional arguments for `fun`.
+#' @param additional_attr (`list`)\cr of additional attributes to apply to the result.
+#'
+#' @details This function add `args` attribute to the original function,
+#' and add an extra class `partial` to the function.
+#' `args` is the argument for the function, and elements in `...` will override the existing
+#' arguments in attribute `args`. `additional_attr` will override the existing attributes.
+#'
+#' @return S3 class "partial", a (`function`)\cr with `args` and `fun_label` attributes.
```
danielinteractive, line 254:
```suggestion
#' @return S3 class "partial", a `function` with `args` and `fun_label` attributes.
```

danielinteractive (commented):
thanks @clarkliming ! please see inside

R/utils.R
```diff
@@ -204,14 +206,22 @@ h_split_control <- function(control, ...) {
 #' Allowed are "L-BFGS-B", "BFGS", "CG"(using [stats::optim()] with corresponding method) and
 #' "nlminb"(using [stats::nlminb()]).
 #'
+#' @return named (`list`)\cr of optimizers created by [h_optimizer_fun()].
```
danielinteractive, line 209:
```suggestion
#' @return Named `list` of optimizers created by [h_optimizer_fun()].
```
note that we only use this parentheses plus cr thing in the params, not in return

R/utils.R
```diff
@@ -167,12 +167,14 @@ h_tr <- function(x) {
...
 
 #' Split Control List
+#'
 #' @description Split the [mmrm_control()] object according to its optimizers and use additional arguments
 #' to replace the elements in the original object.
 #'
 #' @param control (`mmrm_control`)\cr object.
 #' @param ... additional parameters to update the `control` object.
 #'
+#' @return A (`list`)\cr of `mmrm_control`\cr.
```
danielinteractive, line 177:
```suggestion
#' @return A `list` with of `mmrm_control` entries.
```

R/utils.R
```diff
@@ -251,7 +261,7 @@ h_optimizer_fun <- function(optimizer = c("L-BFGS-B", "BFGS", "CG", "nlminb")) {
 #' `args` is the argument for the function, and elements in `...` will override the existing
 #' arguments in attribute `args`. `additional_attr` will override the existing attributes.
 #'
-#' @return S3 class "partial", a (`function`)\cr with `args` and `fun_label` attributes.
+#' @return S3 class "partial", a `function` with `args` and `fun_label` attributes.
```
danielinteractive, line 264:
```suggestion
#' @return Object with S3 class `"partial"`, a `function` with `args` and `fun_label` attributes.
```

danielinteractive (commented):
Thanks @clarkliming almost there!

R/utils.R
clarkliming: got it!

R/utils.R
```diff
@@ -261,7 +261,8 @@ h_optimizer_fun <- function(optimizer = c("L-BFGS-B", "BFGS", "CG", "nlminb")) {
 #' `args` is the argument for the function, and elements in `...` will override the existing
 #' arguments in attribute `args`. `additional_attr` will override the existing attributes.
 #'
-#' @return S3 class "partial", a `function` with `args` and `fun_label` attributes.
+#' @return Object with S3 class `"partial"`, a `function` with `args` attribute(and possible more
```
danielinteractive, line 264:
```suggestion
#' @return Object with S3 class `"partial"`, a `function` with `args` attribute (and possibly more
```

tests/testthat/_snaps/fit.md
```diff
@@ -0,0 +1,42 @@
...
+      Formula:     FEV1 ~ ARMCD + ar1(AVISIT | SEX/USUBJID)
+      Data:        fev_data (used 537 observations from 197 subjects with maximum 4 
+      timepoints)
+      Covariance:  auto-regressive order one (4 variance parameters of 2 groups)
+      Method:      REML
+      Deviance:    3855.007
+      
+      Coefficients: 
+      (Intercept)    ARMCDTRT 
+        40.290948    4.227962
```
danielinteractive, line 38: This kind of snapshot (where we have numbers printed) is problematic, because on CRAN / different machines etc. we might have slightly different results. so I would highly recommend to omit these snapshots.
If you just want to test that mmrm works in certain situation, then best work with expecting silent processing, plus it converged e.g., and if comparing numbers always work with `expect_equal` or the new custom helper (which won't be applicable for this print snapshots)

danielinteractive (commented):
Thanks @clarkliming , let's remove the problematic snapshots, afterwards should be good to merge

tests/testthat/_snaps/fit.md
clarkliming: oh yes... I use the silly optimizer and thought it is fine to leave a snapshot, and in the next test I forgot that I am using a normal optimizer...

danielinteractive (approved):

PR #196, opened 2022-12-15
update contributing guide

### LINKED ISSUE #162: Add Imports explanations to contributing guide
Currently we follow the following (non-standard to be clear, but useful) convention in `mmrm`:
- When only few functions from other packages are needed:
	- We want to use those only.
	- Therefore need to put the package in the `Imports`.
	- We always use `package::function()`.
	- We import (any random) single function from the package to make `R CMD check` happy.
- On the other hand, for completely imported packages such as `checkmate` we don't prefix with the package name.
We need to add this to the contributing guide.

clarkliming:
close #162

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -38,7 +38,7 @@ In package `mmrm`, we follow the following convention in package imports.
     1. Example: `stats`.
 1. When many functions from another package are needed:
     1. In `DESCRIPTION`, add that package into `Imports`.
-    1. In `mmrm-package.R`, we use a `import` with to import every function.
+    1. In `mmrm-package.R`, we use a `import` to import every function.
     1. Use the functions directly without prefix.
     1. Example: `checkmate`.
 
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -30,7 +30,7 @@ deviate from this are explained below.
 
 In package `mmrm`, we follow the following convention in package imports.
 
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -26,6 +26,22 @@ deviate from this are explained below.
     are not exported).
 4.  Functions must be unit tested (even when they are not exported).
 
+### Imports
+
+In package `mmrm`, we follow the following convention in package imports.
+
+1. When only a few function from another package is needed:
```
danielinteractive, line 33:
```suggestion
1. When only a few functions from another package are needed:
```

.github/CONTRIBUTING.md
```diff
@@ -26,6 +26,22 @@ deviate from this are explained below.
...
+    1. In `DESCRIPTION`, add that package into `Imports`.
+    1. In `mmrm-package.R`, we add a `importFrom` with a single function from the package.
+        1. The function imported can be any random function within the package.
+    1. Always use `package::function` style wherever you need to use the function.
+    1. Example: `stats`.
+1. When many functions from another package are needed:
+    1. In `DESCRIPTION`, add that package into `Imports`.
+    1. In `mmrm-package.R`, we use a `import` with to import every function.
```
danielinteractive, line 41:
```suggestion
    1. In `mmrm-package.R`, we use a `import` to import every function.
```

danielinteractive (approved):
Thanks @clarkliming

PR #197, opened 2022-12-15
181: Add Satterthwaite vignette

### LINKED ISSUE #181: add degree of freedom for Satterthwaite vignettes

danielinteractive:
closes #181


vignettes/satterthwaite.Rmd
```diff
@@ -0,0 +1,122 @@
...
+it is only a matter of putting the different quantities together to
+compute the estimate of the denominator degrees of freedom,
+$\hat\nu(\hat\theta)$.
+
+### Jacobian calculation
+
+Currently, we evaluate the gradient of $\Phi(\theta)$ (which is created
+as an R function using `h_covbeta_fun()`) with regards to $\theta$
+evaluated at $\hat\theta$, i.e. the Jacobian $\mathcal{J}(\hat\theta)$,
+numerically using the `jacobian` function from the `numDeriv` package
```
clarkliming, line 93: when we need the `()` in referencing the functions, when we don't need?I am a bit confused

vignettes/satterthwaite.Rmd
```diff
@@ -0,0 +1,122 @@
...
+`h_jac_list()`.
+
+## Multi-dimensional contrast
+
+When $c > 1$ we are testing multiple contrasts at once. Here an F-statistic
+\[
+F = \frac{1}{c} (C\hat\beta)^\top  (C \Phi(\hat\theta) C^\top)^{-1} C^\top (C\hat\beta)
+\]
+is calculated, and we are interested in estimating an appropriate denominator degrees of freedom for $F$,
+while assuming $c$ numerator degrees of freedom. Note that only in special cases,
```
clarkliming, line 106: assuming $c$ as/is the numerator degree of freedom? (c is still a singular instead of plural?)

vignettes/satterthwaite.Rmd
danielinteractive: hm good question :-) In the roxygen documentation we naturally use it with parentheses for functions, because that is how the md syntax works. So I would suggest to do the same in the vignettes. Let me add the hyperlinks here as well though.

danielinteractive: hm I can't because these are not exported and therefore not part of pkgdown reference. ok

clarkliming: oh I see... these [function()] only work for internal function references in the package, and for external function a simple `function` is sufficient. Thank you!

danielinteractive: well it is complicated - in Roxygen, we can use `[function()]` also for internal functions, because we have documentation for all of them. However in the vignette, we cannot reference help pages but only html populated on the pkgdown website. but that includes only exported functions. but to make it look similar to the others I would say here also use `function()` style. I am adding a paragraph in the contributing guide for this now

vignettes/satterthwaite.Rmd
danielinteractive: mh probably better plural, corrected now. thanks!

PR #200, opened 2022-12-16
allow levels to be kept

### LINKED ISSUE #170: Warn when visit levels are dropped / consider flag to include all levels
To do:

- [x] add a warning when time point levels are dropped
- [x] a flag that allows to include all visit levels optionally 
  - [x] needs to go to `mmrm_control()`
- [x] modify https://github.com/openpharma/mmrm/blob/39eb47441e714fd6a584b8946a8d9341c96b754a/R/tmb.R#L268
  - [x] maybe only specifically drop levels of the visit? on the other hand, `lm()` also drops non-observed factor levels silently, see below
- [x] add tests
- [x] add `NEWS` entry

Would be good to be consistent with standard R behavior of other modeling functions. E.g. consider:
```r
df <- data.frame(y = c(0.1, 0.2), x = factor(c("a", "b"), levels = c("a", "b", "c")))
mod <- lm(y ~ x, data = df, singular.ok = FALSE)                 
mod
m <- model.frame(y ~ x, df)
str(m)
mm <- model.matrix(y ~ x, m)
mm
lm
```

Motivation:

mmrm models drop levels for visits, this is usually fine, but when a visit is dropped in all observations, the timepoint will be changed. e.g, if all subjects completes visit 1, 2 and 4, 3 is missed. Then it will be converted to 0, 1 and 2 in mmrm models. This changes the covariance matrix structure.

```r

df <- data.frame(
  usubjid = c("1", "1", "1"),
  visits = factor(c("1", "2", "4"), levels = as.character(1:4)),
  resp = rnorm(3)
)

fit <- mmrm(resp ~ ar1(visits|usubjid), data = df)


fit$tmb_data$visits_zero_inds  # get 0,1,2

fit$cov  # names are 1, 2 and 4

```

although we have  https://github.com/openpharma/mmrm/blob/88836158c2a71826ee98fbe997a80654ad961f49/tests/testthat/test-tmb.R#L1535

but it still seems not very reasonable.
e.g. a study has 4 scheduled visits, week 1, 2, 3 and 4. Due to some kind of lockdown, week3 data is not available. When users try to use "AR1", they assume that the correlation between each consecutive timepoint is the same. However, here in our model, the correlation between week 1 and 2 is the same as the correlation between 2 and 4.

I think we should at least provide a "Note" to let users know that their assumption might get violated?

clarkliming:
close #170


tests/testthat/test-tmb.R
```diff
@@ -1609,18 +1610,61 @@ test_that("h_mmrm_tmb works even when time point variable has unused factor leve
...
-  ))
+  ), "In AVISIT there are dropped visits: SCREENING")
   expect_class(result, "mmrm_tmb")
   expect_identical(
     rownames(VarCorr(result)),
     c("VIS1", "VIS2", "VIS3", "VIS4")
   )
 })
 
+test_that("h_mmrm_tmb works if we keep the unused factor levels for specific covariance structure", {
```
danielinteractive, line 1625:
```suggestion
test_that("fit_mmrm works if we keep the unused factor levels for specific covariance structure", {
```

danielinteractive: probably need to adjust the other test labels above too, below as well

NEWS.md
```diff
@@ -11,6 +11,13 @@
   `mmrm(..., start = start, optimizer = c("BFGS", "nlminb"))`. These arguments will go into `mmrm_control`.
   In addition, `optimizer = "automatic"` is deprecated and will give you warning. It is a default behavior now to
   try all optimizers (if multiple optimizer provided) if the first optimizer fails.
+- Add new arguments to `mmrm_control` to allow users to keep all levels in visits. This allows user to specify the distance of
```
danielinteractive, line 14:
```suggestion
- Add new argument `drop_visit_levels` to allow users to keep all levels in visits. This allows user to specify the distance of
```

danielinteractive: I would move the details what this allows to do etc. to the `mmrm_control ` function documentation, I feel a bit too much detail for the `NEWS` here

danielinteractive: here in addition please mention that now a warning is raised if a visit level is dropped, which was done silently before

tests/testthat/test-tmb.R
```diff
@@ -1609,18 +1610,61 @@ test_that("h_mmrm_tmb works even when time point variable has unused factor leve
...
+  # Create a data set where one visit level only has NA in the data.
```
danielinteractive, line 1626: hm so looking at this now, we need to repeat this data manipulation. maybe nicer to already include an `AVISIT2` variable in the `fev_data` and do it there, document accordingly, and then just use here?

danielinteractive (commented):
thanks @clarkliming , nice work, see comments inside

tests/testthat/test-tmb.R
clarkliming: Here I use a simpler approach of add a "SCREENING" level. I think now this is much simpler and no need to update the `fev_data` now.

danielinteractive: cool

NEWS.md
danielinteractive: This is still too much detail , please move to `mmrm_control` documentation

PR #210, opened 2022-12-22
residuals S3 method for mmrm_tmb

### LINKED ISSUE #191: Add `residuals()` method for `mmrm_tmb`
Motivation: 1) kind of a standard for models, 2) needed for support in the simulation framework [DHARMa](https://github.com/florianhartig/DHARMa/wiki/Adding-new-R-packages-to-DHARMA#integrating-a-package-with-dharma).

To do:

- [ ] Add a short design doc to work out the approach
	- [ ] include raw and Pearson (standardized) residuals
	- [ ] think about optional "decorrelating" of residuals by multiplying with inverse Cholesky factor of covariance matrix
	- [ ] see `tern.mmrm` for starting points since we calculate the residuals there
- [ ] Add method
- [ ] Add tests
- [ ] Add to NEWS

danjamstat:
closes #191

Currently the method returns raw / 'response', 'Pearson' or scaled / 'normalized' residuals. Weights are incorporated into the Pearson residuals, but currently are ignored for the normalized residuals.

Grouped covariance structures are allowed for.


danielinteractive (commented):
Thanks a lot @danjamstat for the PR! Much appreciated :-)

Re: your comment that weights are ignored for normalized residuals, is this something you need help with / have questions how to include the weights? 

Some high level comments to start with: (you can tick them off here as you work through them) Please
- [ ] add `NEWS` entry
- [ ] include yourself in authors and include AstraZeneca in copyright holders
- [ ] consider including `residuals` section in the introduction vignette
- [ ] The residuals method is quite long. consider refactoring with helper functions (see contributing guide) to make the method smaller and can test separately the helper functions
- [ ] It would be very good to compare the results with SAS results in additional integration tests to get further confidence
- [ ] make sure that the tests cover all the cases (in particular all if/else conditions) and compare with numbers instead of just checking type (this is important here because there are calculations in the method, not just extraction of prepared results)

(let me know if a chat would be helpful, happy to set that up)

danjamstat:
Thanks for the review, Daniel. I'll work my way through your comments. I've resolved the issue with the weights and compared vs SAS. One question regarding the use of helper functions - should they be in the same script as the main method (tmb-methods.R)?

danielinteractive:
Great thanks @danjamstat ! For now can put them in the same file, we could always move them later without effort.

danjamstat:
- [ ] add NEWS entry
- [x]  include yourself in authors and include AstraZeneca in copyright holders
- [ ] consider including residuals section in the introduction vignette
- [x] The residuals method is quite long. consider refactoring with helper functions (see contributing guide) to make the method smaller and can test separately the helper functions
- [x] It would be very good to compare the results with SAS results in additional integration tests to get further confidence
* But I should probably add a test for residuals from a model using weights 
- [x]  make sure that the tests cover all the cases (in particular all if/else conditions) and compare with numbers instead of just checking type (this is important here because there are calculations in the method, not just extraction of prepared results)

DESCRIPTION
```diff
@@ -11,10 +11,12 @@ Authors@R: c(
     person("Liming", "Li", , "liming.li@roche.com", role = "aut"),
     person("Ya", "Wang", , "ya.wang10@gilead.com", role = "aut"),
     person("Craig", "Gower-Page", , "craig.gower-page@roche.com", role = "ctb"),
+    person("Dan", "James", , "dan.james@astrazeneca.com", role = "ctb"),
     person("Boehringer Ingelheim Ltd.", role = c("cph", "fnd")),
     person("Gilead Sciences, Inc.", role = c("cph", "fnd")),
     person("F. Hoffmann-La Roche AG", role = c("cph", "fnd")),
-    person("Merck Sharp & Dohme, Inc.", role = c("cph", "fnd"))
+    person("Merck Sharp & Dohme, Inc.", role = c("cph", "fnd")),
+    person("AstraZeneca plc", role = c("cph"))
```
danielinteractive, line 19: not also funder? like if you do this on work time, it would be funder too

danielinteractive:
@danjamstat cool thanks for the updates, can you also add a test for residuals from a model using weights as suggested? once done I can re-review

DESCRIPTION
danjamstat: Thanks Daniel, I've updated that.

danjamstat:
@danielinteractive I might be doing something wrong, but I'm getting very slightly different fixed effect estimates in SAS compared with get_mmrm_weighted() . The residuals between R and SAS are close but not exact, but it's hard to know if this is due to the residuals code or due to differences in the effect estimates (or both).

danielinteractive:
Thanks @danjamstat , so very small differences could be there. Did you start from something like https://github.com/openpharma/mmrm/blob/main/design/SAS/sas_weighted.R ?

Also note that https://github.com/openpharma/mmrm/blob/11e3b2d0f28cf98a526843f240f8ae8a6c79a92d/tests/testthat/helper-examples.R#L24 actually uses random weights, not the ones in `fev_data`. We actually only use that for non-numeric tests, see https://github.com/openpharma/mmrm/search?q=get_mmrm_weighted search results.

tests/testthat/test-tmb-methods.R
```diff
@@ -164,3 +164,65 @@ test_that("print.mmrm_tmb works as expected for rank deficient fits", {
...
+  result_norm <- expect_silent(residuals(object, type = "normalized"))
+  expect_double(result_norm, len = length(object$tmb_data$y_vector))
+  expect_equal(head(result_norm, 5), c(-0.29917, -4.07727, -0.51393, -0.96419, 0.45466), tolerance = 1e-4)
+})
+
+test_that("residuals helper functions work as expected", {
+  object <- get_mmrm()
+  resid_response <- residuals(object, type = "response")
+
+  result_pearson <- h_residuals_pearson(object, resid_response)
```
danielinteractive, line 222: best to 1) split this into 2 small tests (one for each helper function), and 2) do at least some sampled numeric comparisons on result, not just checking length and type

danielinteractive:
nice progress, thanks @danjamstat ! please request review when all boxes ticked off.

danjamstat:
Hi @danielinteractive , thanks for the comment regarding the weighted fit, the problem was mostly me rather than the minor discrepancies vs. SAS  😄 

Possible a silly question, but when updating the NEWS, should I add text above '# mmrm 0.2.2.9001' and if so, should I specify a new version number?

danielinteractive:
Thanks @danjamstat , please update below that version number

DESCRIPTION
```diff
@@ -11,10 +11,12 @@ Authors@R: c(
     person("Liming", "Li", , "liming.li@roche.com", role = "aut"),
     person("Ya", "Wang", , "ya.wang10@gilead.com", role = "aut"),
     person("Craig", "Gower-Page", , "craig.gower-page@roche.com", role = "ctb"),
+    person("Dan", "James", , "dan.james@astrazeneca.com", role = "ctb"),
```
danielinteractive, line 14:
```suggestion
    person("Dan", "James", , "dan.james@astrazeneca.com", role = "aut"),
```
because it is a significant addition

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
   cat("\n")
   invisible(x)
 }
+
+
+#' Residuals method for mmrm objects
+#'
```
danielinteractive, line 213:
```suggestion
```
no need for this since you document this in the `mmrm_tmb_methods` page

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
...
+#' @describeIn mmrm_tmb_method to obtain residuals - either unscaled or normalized
```
danielinteractive, line 214:
```suggestion
#' @describeIn mmrm_tmb_methods to obtain residuals - either unscaled ('response'), 'pearson' or 'normalized'.
```

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
...
+#' @param object (`mmrm_tmb`)\cr the fitted MMRM.
```
danielinteractive, line 215:
```suggestion
```
not needed because we have that documented on the page already (see above)

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
...
+#' @param type (`character`)\cr unscaled ('response'), 'pearson' or 'normalized'. Default is 'response',
```
danielinteractive, line 216:
```suggestion
#' @param type (`string`)\cr unscaled (`response`), `pearson` or `normalized`. Default is `response`,
```
since needs to be scalar.

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
...
+#' and this is the only type available for use with models with a spatial covariance structure.
+#' @param ... \cr not used in this method.
```
danielinteractive, line 218:
```suggestion
```

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
...
+#' @return Vector of residuals
```
danielinteractive, line 219:
```suggestion
```

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
...
+#' @exportS3Method
+#' @examples
+#' # residuals:
+#' residuals(object, type = "response")
+#' residuals(object, type = "pearson")
+#' residuals(object, type = "normalized")
+#' @references
+#' - \insertRef{galecki2013linear}{mmrm}
+residuals.mmrm_tmb <- function(object, type = c("response", "pearson", "normalized"), ...) {
+  assert_class(object, "mmrm_tmb")
```
danielinteractive, line 230:
```suggestion
```
not needed because the method dispatch will only work for `mmrm_tmb` objects

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
...
+  assert_character(type)
+  type <- match.arg(type, several.ok = FALSE)
```
danielinteractive, line 232:
```suggestion
  type <- match.arg(type)
```
is sufficient

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
...
+  resids_unscaled <- component(object, "y_vector") - fitted(object)
+  if (type == "response") {
+    return(unname(resids_unscaled))
```
danielinteractive, line 235:
```suggestion
    unname(resids_unscaled)
```
do we need the `unname` here? or would it be better to do that already when defining `resids_unscaled`?

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
...
+  } else {
+    if (object$formula_parts$is_spatial) {
+      stop("Only raw / response residuals are available for models with spatial covariance structures.")
```
danielinteractive, line 238:
```suggestion
      stop("Only 'response' residuals are available for models with spatial covariance structures.")
```
to be more clear

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
...
+    }
+    if (type == "pearson") {
+      h_residuals_pearson(object, resids_unscaled)
```
danielinteractive, line 241: nice job!!

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
...
+    } else if (type == "normalized") {
+      h_residuals_normalized(object, resids_unscaled)
+    }
+  }
+}
+
+#' Calculate Pearson residuals
```
danielinteractive, line 248:
```suggestion
#' Calculate Pearson Residuals
```
Title Case

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
...
+#'
+#' This is used by [residuals.mmrm_tmb()] to calculate Pearson residuals.
+#'
+#' @param object (`mmrm_tmb`)\cr the fitted MMRM.
+#' @param resids_unscaled (`numeric`)\cr the raw/response residuals.
```
danielinteractive, line 253:
```suggestion
#' @param resids_unscaled (`numeric`)\cr the response residuals.
```

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
...
+#'
+#' @return Vector of residuals
```
danielinteractive, line 255:
```suggestion
#' @return Vector of residuals.
```

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
...
+#'
+#' @keywords internal
+h_residuals_pearson <- function(object, resids_unscaled) {
+  visits <- as.numeric(object$tmb_data$full_frame[[object$formula_parts$visit_var]])
```
danielinteractive, line 259: please add two assertions - since this is not a method we need to also assert `object`, and the other argument anyway

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
...
+  if (component(object, "n_groups") == 1) {
+    visit_sigmas <- sqrt(diag(object$cov))
+    resids <- resids_unscaled / visit_sigmas[visits] * sqrt(object$tmb_data$weights_vector)
+  } else {
```
danielinteractive, line 263: I think the code here becomes less duplicate/simpler if you first define 
```r
visit_sigmas <- if (component(object, "n_groups") == 1) {
  sqrt(diag(object$cov))
} else {
  # work with grp_visit_sigmas etc.
}
```
and then use that because rest of code is the same for both cases

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
...
+    grp_visit_sigmas <- lapply(object$cov, function(x) sqrt(diag(x)))
+    subject_grps <- object$tmb_data$full_frame[[object$formula_parts$group_var]]
+    nobs <- nrow(object$tmb_data$full_frame)
+    resids <- sapply(1:nobs, function(x) {
+      resids_unscaled[x] / grp_visit_sigmas[[subject_grps[x]]][visits[x]] * sqrt(object$tmb_data$weights_vector[x])
+    })
+  }
+  return(unname(resids))
```
danielinteractive, line 271: avoid explicit return, also see comment above on where to put the `unname` in central place

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
...
+}
+
+#' Calculate normalized residuals
```
danielinteractive, line 274: similar comments as above apply here for style

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
...
+#' @param resids_unscaled (`numeric`)\cr the raw/response residuals.
+#'
+#' @return Vector of residuals
+#'
+#' @keywords internal
+h_residuals_normalized <- function(object, resids_unscaled) {
+  resid_df <- data.frame(subject = object$tmb_data$full_frame[[object$formula_parts$subject_var]],
+                         time = as.numeric(object$tmb_data$full_frame[[object$formula_parts$visit_var]]),
+                         residual = resids_unscaled,
+                         weights = object$tmb_data$weights_vector)
```
danielinteractive, line 288:
```suggestion
  resid_df <- data.frame(
    subject = object$tmb_data$full_frame[[object$formula_parts$subject_var]],
    time = as.numeric(object$tmb_data$full_frame[[object$formula_parts$visit_var]]),
    residual = resids_unscaled,
    weights = object$tmb_data$weights_vector
  )
```
(style)

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
...
+
+  subject_list <- split(resid_df, resid_df$subject)
+
+  if (component(object, "n_groups") == 1) {
```
danielinteractive, line 292:
```suggestion
  lower_chol_list <- if (component(object, "n_groups") == 1) {
```
and then remove inner assignment within curly brackets

R/tmb-methods.R
```diff
@@ -207,3 +207,111 @@ print.mmrm_tmb <- function(x,
...
+    lower_chol_list <- lapply(seq_along(subject_list), function(x) {
```
danielinteractive, line 293: so, in https://github.com/openpharma/mmrm/blob/35da65b6c0f624e60ba142969455a5de165518cd/R/tmb.R#L456 we see that we actually have the lower Cholesky factors already in `tmb_report$covariance_lower_chol`. So we should use those, instead of recalculating them. If there is too much logic then duplicated with `h_mmrm_tmb_extract_cov` we could also save this explicitly as a list of Cholesky factors in the object.

tests/testthat/test-tmb-methods.R
```diff
@@ -164,3 +164,81 @@ test_that("print.mmrm_tmb works as expected for rank deficient fits", {
...
+  result_norm <- expect_silent(residuals(object, type = "normalized"))
+  expect_double(result_norm, len = length(object$tmb_data$y_vector))
+  expect_equal(head(result_norm, 5), c(-0.29917, -4.07727, -0.51393, -0.96419, 0.45466), tolerance = 1e-4)
+})
+
+test_that("residuals works as expected with a model using spatial covariance structure", {
+  object <- get_mmrm_spatial()
+
+  result_resp <- expect_silent(residuals(object, type = "response"))
+  expect_double(result_resp, len = length(object$tmb_data$y_vector))
```
danielinteractive, line 222: here better also check actual values, comparing with SAS results

vignettes/introduction.Rmd
```diff
@@ -413,6 +413,32 @@ if (require(emmeans)) {
 }
 ```
 
+# Residuals
```
danielinteractive, line 416: better move this under `Extraction of Model Features` above

danielinteractive (commented):
Thanks a lot @danjamstat ! Please see comments inside. One major comment is about the normalized residuals - we don't need to recompute the Cholesky factors. Please let me know if any questions or if you'd like to chat about it

danielinteractive (approved):
many thanks Dan!

PR #211, opened 2022-12-23
add covrpage vignette to communicate package testing health

yonicd:
added [covrpage](https://yonicd.github.io/covrpage/index.html) output to the package [vignettes/site](https://sagerx.github.io/mmrm/main/articles/tests_and_coverage.html) and [tests/readme.md](https://github.com/sagerx/mmrm/tree/main/tests) to help to communicate package health to prospective users.

[gha](https://github.com/sagerx/mmrm/blob/main/.github/workflows/R-linux.yml) also added to allow for autogeneration of readme on commit/PR

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@
 # R specific hooks: https://github.com/lorenzwalthert/precommit
 repos:
   - repo: https://github.com/lorenzwalthert/precommit
-    rev: v0.3.2.9001
+    rev: v0.3.2.9007
     hooks:
       - id: style-files
         args:
@@ -76,7 +76,7 @@ repos:
           design/.*
           )$
   - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v3.0.0-alpha.0
+    rev: v3.0.0-alpha.4
     hooks:
       - id: prettier
         exclude: >
@@ -87,7 +87,7 @@ repos:
           README.md
           )$
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.3.0
+    rev: v4.4.0
     hooks:
       - id: check-added-large-files
         args: ["--maxkb=200"]
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: mmrm
 Title: Mixed Models for Repeated Measures
-Version: 0.2.2.9010
+Version: 0.2.2.9011
 Authors@R: c(
     person("Daniel", "Sabanes Bove", , "daniel.sabanes_bove@roche.com", role = c("aut", "cre")),
     person("Julia", "Dedic", , "julia.dedic@roche.com", role = "aut"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# mmrm 0.2.2.9010
+# mmrm 0.2.2.9011
 
 ### New Features
 


danielinteractive:
Thanks @yonicd ! 
@cicdguy cqn you please review re: automation?

.github/workflows/R-linux.yml
```diff
@@ -0,0 +1,133 @@
...
+      - name: covrpage-setup
+        run: pak::pkg_install(c('hadley/emo','yonicd/covrpage@actions'), ask = FALSE)
+        shell: Rscript {0}
+
+      - name: covrpage
+        run: |
+          git config --local user.email "actions@github.com"
+          git config --local user.name "GitHub Actions"
+          Rscript -e 'remotes::install_github("hadley/emo")' \
+                  -e 'remotes::install_github("yonicd/covrpage@actions")' \
```
danielinteractive, line 129: do we need that specific branch? Did you already try to make a PR towards the main `covrpage` repo? just thinking for longer term maintenance that would be better.

danielinteractive (commented):
thanks @yonicd , nice page! see one question inside. Also:
- Would it also be possible to let the reader click on a file name e.g. and then see highlighted which lines of code are not covered by tests?
- I would like to let @cicdguy review the github workflows
- you can add yourself to contributors in the package description

yonicd:
This tool is meant as a communication to general users the testing diversity and coverage in a package and have it accessible on the github level (can't read html in github pages). The covr html output leaves a lot of information out of the summary report which is less useful for non-programmers to digest.

.github/workflows/R-linux.yml
yonicd: Yes. The links for the tests are taking you directly to the lines that were not run. There are icons next to tests that gave either warnings/failures as a way to know which ones were the offending ones. 

I merge that branch into main, so that can be amended to be `yonicd/covrpage`.

yonicd:
those are great. but to @danielinteractive's question above. covrpage gives you links to each test and it will run even if test(s) fail so you get the coverage regardless and you can link directly to what was the problem area. Users of covrpage have found that having the report tests/readme.md makes the package more accessible to prospective users and helps with developmemt collaborations.

danielinteractive:
Thanks @cicdguy and @yonicd - I am open to this change but I wonder a bit if we really need this and if we can only have one coverage run instead of two for each commit. 

In this package coverage is so high that the coverage badge which I think we have basically says all for users?

For developers on the other hand the coverage report that is already generated as part of GHA is more relevant as it shows during PR what is missing.

danielinteractive:
Sounds like a good idea to me - @yonicd what do you think?

yonicd:
It is up to you guys. 

- We will be talking about the transparency and application of tests in other content in the working group. Part of the planned discussion will be different strategies of communication, and the comparison of `covr` and `covrpage` will be added there. 
- wrt to package size, I don't think that 2 small text files will move the needle.
- Part of my motivation of `covrpage` was that the `covr` badge is a poor way to communicate package health. It is a crude summary statistic that can mask a lot of problems. 
- Report information is comparable, the only differences are that you do not get a report if the package fails a test and there are no links to specific lines for tests in `covr`.

danielinteractive:
Thanks @yonicd!
I think the `covr` and `covrpage` options are both nice and worth discussing and introducing as part of the SWE WG.
Since in this package and repo we already use `covr` I would tend to keep using that route for this package.

How about @cicdguy would you create a separate PR with the proposed changes and then we can compare more easily, regarding output and also regarding extra changes in the code? I guess that PR could be nice to include then also in the R package template repo so other projects can benefit.

danielinteractive:
@cicdguy any chance we can get the alternative PR soon?

danielinteractive:
Great thanks @cicdguy for the good news :-) Looking forward! Let me know when I can help testing or so.

yonicd:
looks interesting. 

A few questions

coverage: 
- The coverage page is the html output the covr package gives? how large is that output (package size-wise) and additional js deps to support the tabs (datatables, crosstalk, highlight...)?

unit testing page: 
- Is there any way to present the unit testing in a more inviting format to regular users trying to understand what has been tested. The current format looks like a long log file that has been dumped into html format, I am not sure if that serves the intended purpose of communicating to general users.

PR #213, opened 2023-01-03
add spatial exponential KR

### LINKED ISSUE #199: Add Kenward-Roger implementation for spatial covariance structures
This is not supported at the moment (see e.g. https://github.com/openpharma/mmrm/blob/main/src/kr_comp.cpp#L19).

To do:

- [x] Add corresponding algorithm considerations to the vignette
- [x] Implement 
	- [x] R side
	- [x] C++ side
- [x] Integration test comparing vs. SAS results
- [x] Add to NEWS

clarkliming:
close #199

a parent class derivatives_base is created to use only one map<std::vector<int>, derivatives_base<double>*> to hold the by group derivative class, no matter spatial or not. The spatial derivatives class is now named "derivatives_sp_exp" which is the only spatial covariance structure we have, and for non-spatial where auto diff is utilized we change the name "derivatives_nonspatial"

to update in the future:

1. the possibility of harmonizing the coordinates and visits. currently, they are two separate objects so we are always using both to calculate the derivatives, whether it is spatial(only required coordinates) or not (only require visits in integer, similar to coordinates). This issue is already created


src/kr_comp.cpp
```diff
@@ -16,37 +16,55 @@ List get_pqr(List mmrm_fit, NumericVector theta) {
   int n_groups = mmrm_fit["n_groups"];
   IntegerVector subject_groups = mmrm_fit["subject_groups"];
   NumericVector weights_vector = mmrm_fit["weights_vector"];
-  if (is_spatial) {
-    Rf_error("Spatial Kenward-Roger not implemented yet!");
-  }
+  NumericMatrix coordinates = mmrm_fit["coordinates"];
+  matrix<double> coords = as_matrix(coordinates);
```
clarkliming, line 20: currently there are no checks for other spatial covariance structure

src/kr_comp.cpp
```diff
@@ -16,37 +16,55 @@ List get_pqr(List mmrm_fit, NumericVector theta) {
...
+    matrix<double> dist_i(n_visits_i, n_visits_i);
     for (int i = 0; i < n_visits_i; i++) {
       visit_i[i] = visits_zero_inds[i + start_i];
     }
+    dist_i = euclidean(matrix<double>(coords.block(start_i, 0, n_visits_i, coordinates.cols())));
     int subject_group_i = subject_groups[i] - 1;
+    matrix<double> sigma_inv, sigma_d1, sigma_d2, sigma, sigma_inv_d1;
+
+    if (is_spatial) {
+      sigma_inv = spatial_by_group[subject_group_i].get_inverse(dist_i);
```
clarkliming, line 53: currently dist_i and visit_i are used separately

vignettes/covariance.Rmd
```diff
@@ -226,7 +226,7 @@ The distance matrix is
 For spatial exponential, the covariance structure is defined as follows:
 
 \[
-  \rho_{ij} = \rho^{-d_{ij}}
+  \rho_{ij} = \rho^{d_{ij}}
```
clarkliming, line 229: fix of the previous error

NEWS.md
```diff
@@ -1,5 +1,11 @@
 # mmrm 0.2.2.9000
 
+### New Features
+- Add support for Kenward-Roger adjusted coefficients covariance matrix and
+  degrees of freedom in `mmrm` for spatial exponential covariance structure.
```
danielinteractive, line 5:
```suggestion
- Add Kenward-Roger support for spatial covariance structures.
```
just a bit more concise

vignettes/covariance.Rmd
danielinteractive: ah good that you caught that

vignettes/kenward.Rmd
```diff
@@ -324,4 +324,59 @@ of $\Sigma$ over our parameters, are non-zero matrices. However, if we use the e
...
+## Implementations in `mmrm`
+
+In package `mmrm`, we have implemented Kenward-Roger calculations based on the previous sections.
+Specially, for the first-order and second-order derivatives, we use automatic differentiation to obtain
+the results easily for non-spatial covariance structure. For spatial covariance structure, we derive the
+exact results.
+
+### Spatial Exponential Derivatives
+
+For spatial exponential covariance structure, we have
```
danielinteractive, line 336: very good that you add this here

danielinteractive (commented):
looks good, thanks @clarkliming !
I did not check the math or math/code equivalence. Can you just do a last comparison yourself maybe to check math vs. code?
Re: improvement ideas - I think no. 1 we have an issue already. for no. 2 and 3, can you please write issues? try to make them detailed enough also with links to lines of code etc. so that others can understand too. thank you

danielinteractive (approved):
thanks Liming, this `auto` stuff is interesting. Just looked it up at https://www.geeksforgeeks.org/type-inference-in-c-auto-and-decltype/

As discussed, either address the small issues mentioned now, or can also make separate issues and we can put them on backlog, then just need detailed issue description.

clarkliming:
thank you @danielinteractive I am working on the harmonization part now in this  PR

src/derivatives.h
```diff
@@ -113,22 +123,23 @@ struct derivatives_nonspatial {
...
+  matrix<Type> get_sigma_derivative1(std::vector<int> visits, matrix<Type> dist) override {
      if (this->sigmad1_cache.count(visits) > 0) {
       return this->sigmad1_cache[visits];
     } else {
       Eigen::SparseMatrix<Type> sel_mat = this->get_sel_mat(visits);
       int n_visists_i = visits.size();
       matrix<Type> ret = matrix<Type>(this->n_theta * n_visists_i, n_visists_i);
       for (int i = 0; i < this->n_theta; i++) {
-        ret.block(i  * n_visists_i, 0, n_visists_i, n_visists_i) = sel_mat * this->sigmad1_cache[this->full_visit].block(i  * this->n_visits, 0, this->n_visits, this->n_visits) * sel_mat.transpose();
+        auto zz = sel_mat * this->sigmad1_cache[this->full_visit].block(i  * this->n_visits, 0, this->n_visits, this->n_visits) * sel_mat.transpose();
```
danielinteractive, line 134: what is the benefit of changing this? just curious

src/kr_comp.cpp
```diff
@@ -27,13 +27,15 @@ List get_pqr(List mmrm_fit, NumericVector theta) {
   matrix<double> P = matrix<double>::Zero(p * n_theta, p);
   matrix<double> Q = matrix<double>::Zero(p * theta_size_per_group * n_theta, p);
   matrix<double> R = matrix<double>::Zero(p * theta_size_per_group * n_theta, p);
-  std::map<int, derivatives_sp_exp<double>> spatial_by_group;
-  std::map<int, derivatives_nonspatial<double>> nonspatial_by_group;
+  // Use map to hold these base class pointer(can also work for child class)
```
danielinteractive, line 30:
```suggestion
  // Use map to hold these base class pointers (can also work for child class objects).
```

src/kr_comp.cpp
```diff
@@ -47,20 +49,12 @@ List get_pqr(List mmrm_fit, NumericVector theta) {
...
-      sigma_inv_d1 = spatial_by_group[subject_group_i].get_inverse_derivative(dist_i);
-    } else {
-      sigma_inv = nonspatial_by_group[subject_group_i].get_inverse(visit_i);
-      sigma_d1 = nonspatial_by_group[subject_group_i].get_sigma_derivative1(visit_i);
-      sigma_d2 = nonspatial_by_group[subject_group_i].get_sigma_derivative2(visit_i);
-      sigma = nonspatial_by_group[subject_group_i].get_sigma(visit_i);
-      sigma_inv_d1 = nonspatial_by_group[subject_group_i].get_inverse_derivative(visit_i);
-    }
+    
+    sigma_inv = derivatives_by_group[subject_group_i]->get_inverse(visit_i, dist_i);
```
danielinteractive, line 53: nice

src/test-derivatives.cpp
```diff
@@ -80,7 +81,8 @@ context("derivatives_sp_exp struct works as expected") {
       0, 0.5, 1,
       0.5, 0, 0.5,
       1, 0.5, 0;
-    auto sigma = sp.get_sigma(dist);
+    std::vector<int> v(3);
```
danielinteractive, line 84: should we better explicitly say which values should be in `v`?

danielinteractive (commented):
thanks @clarkliming , looks good.
Related: can we rename https://github.com/openpharma/mmrm/blob/main/tests/testthat/test-krcomp.R to `test-kenwardroger.R` (just test prefix and the file name from R folder) plus
add more unit tests for `h_get_kr_comp` to make sure we cover all cases in `get_pqr`?

src/test-derivatives.cpp
clarkliming: for spatial derivatives, v is not used at all; so this is can be any uninitialized value; or we can make it 0 length vector

src/derivatives.h
clarkliming: oh it is just too wide to be in one line on my screen; actually there are no benefit

danielinteractive: :-D

clarkliming:
action ghcr.io/insightsengineering/rstudio_4.2.1_bioc_3.15, version latest  is not completed. any ideas @cicdguy ?

danielinteractive:
@clarkliming I think we need to take that version out - since https://github.com/openpharma/mmrm/actions/runs/3845502481/jobs/6549719107 has completed successfully

PR #216, opened 2023-01-12
`PROC MIXED` Repeated Measures Variable Reordering

PhilBoileau:
Comparisons between `mmrm`'s `mmrm()` and SAS's `PROC MIXED` can be complicated by the latter's handling of repeated measures variables.  `PROC MIXED` sometimes reorders and relabels this variable, producing jumbled parameter estimates. This PR explores this behaviour. Recommendations for performing a fair comparison of `mmrm()` and `PROC MIXED` are forthcoming.


danielinteractive (approved):

clarkliming:
some background here:

1. in R we have factors (although the factors do not have order, they are encoded using integers and we can use the integer representation to have the order)
2. when data is transferred to Python, the factors are converted to Categorical. For Categorical dtype, its kind is "O", which means that it is an object.
3. when data is transferred from Python to SAS, the data are processed according to their data types. The transfer is implemented through some SAS code (and the data types are set up after checking the python data frame data types). Objects are recognized as strings.
4. in SAS you will get only strings instead of factor/integers.

And in SAS, there are not factors. So in proc mixed the class variables will be ordered by format (the printed value) by default. That's why you have 1, 10, 2, ...

PhilBoileau:
@clarkliming Thank you so much for your thorough explanation. This confirms what we believed was happening.

@danielinteractive How do you feel about me moving the code demonstrating `PROC MIXED`'s behaviour to a README at the top level of `simulations/`? With the addition of @clarkliming's explanation, I think such a note would be helpful for future collaborators wishing to benchmark `mmrm()` and `PROC MIXED`.

danielinteractive:
good idea, thanks @PhilBoileau for the initiative

danielinteractive (approved):

Issue #174, opened 2022-11-08
Consider nested covariance structures?


danielinteractive:
So recently we had a user who wanted to model multiple assessments per visit per subject. We currently cannot model this adequately with the existing covariance structure framework, because:

1) we cannot have multiple observations per visit (checked since #173)
2) also it does not really make sense to just define `time = interaction(assessment, visit)` and use `time` 

What could be better "nested" covariance structures here? Basically we still want to keep independence between subjects (subject blocks in overall covariance matrix) because that allows us to scale the computations well with number of observations. On the other hand we would like to include additional structure within subjects.

I am wondering if the Kronecker idea (@clarkliming) could be useful here or whether @kkmann has other ideas.

kkmann:
Very interesting point, I wonder how sas handles this?

This reminds me of the discussion we had a while ago on the relation between MMRM and Gaussian Processes. The Kronecker idea is (I think) equivalent to something that is quite common in the GP literature of composing elementary covariance functions to model more complex structures, see e.g. https://www.cs.toronto.edu/~duvenaud/cookbook/. A population-marginal MMRM is (haven't x-ed it out yet) equivalent to a GP with mean structure (usually the mean structure is not modelled in GP literature) and a covariance function for the Gaussian residuals. The "usual" covariance structures are all stationary covariance functions for a discrete metric on the space of visits. I.e. the covariance only depends on the visits through the number of visits between observations, not the  actual timings. The "spatial" covariance structures relax this by depending on the actual time difference. From a covariance function perspective the problem of multiple observations per individual/visit arises because the difference (in either type of structure, spatial or not) being zero, hence the observations would be perfectly correlated and cannot differ (singular vcov matrix). Numerically the fix is to add a diagnoal to the vcov matrix. In covariance function terms that would be the composition of the original kernel with a "nugget" kernel. The Kronecker trick is (I think) nothing but the product of two covariance functions (see e.g. http://gaussianprocess.org/gpml/chapters/RW4.pdf or the kernel cookbook link above again). Here it seems more natural to add kernels though. If mmrm were to support such compositions (product and weighted sum), this would also allow to perform implicit covariance structure selection if the weights of a sum of covariance structures are also estimated. Probably hard to identify well in practice though.

garcia-mc:
Let us assume we have two measurements, A and B, per visit per patient. Then Daniel suggested to write each individual-block in the overall covariance matrix as (assuming there are only two visits present in the data):

[...]

PR #225, opened 2023-01-20
Refactor to use `cov_struct` for covariance structure

dgkf:
# Description

Adds a declarative covariance structure objects outside of `mmrm`'s internals.

Passing covariance structure via a formula still works as usual, but it can alternatively be provided by a `covariance = cov_struct("us", "time", "subject")` syntax.

- `cov_struct` for decoratively specifying structure variables
- `as.cov_struct` for coercion from formula & calls
- `format.cov_struct` for prettier cli communication
- A few helpers for handling formula processing
- Started breaking up some nasty functions into smaller, cleaner parts. Namely `h_mmrm_tmb_formula_parts` is now effectively just shuffling data around instead of doing a bunch of formula black magic. 

# Still TODO

- [x] I saw that there's an existing `cov_type` object that I need to reconcile against a similar object I created in the process.
- [x] use `checkmate::makeAssertCollection()` to handle simultaneous assertion feedback, currently handled by some custom handlers.
- [x] lots of testing still to be done


R/cov_struct.R
```diff
@@ -0,0 +1,289 @@
...
+
+#' Coerce into a Covariance Structure Definition
+#'
+#' @details
+#' A covariance structure can be parsed from a model definition formula or call.
+#' Generally, covariance structures defined using non-standard evaluation take
+#' the following form:
+#'
+#' ```
+#' type( (visit, )* visit | (group /)? subject )
```
danielinteractive, line 185: I would separate the spatial and non-spatial covariance structures here for the explanation, because for spatial ones we talk about coordinates instead of visits, see e.g. https://github.com/openpharma/mmrm/blob/main/vignettes/algorithm.Rmd#L168

R/cov_struct.R
```diff
@@ -0,0 +1,289 @@
...
+#' ```
+#'
+#' For example, formulas may include terms such as
+#'
+#' ```r
+#' us(time | subject)
+#' cp(time | group / subject)
+#' sp_exp(time1, time2 | group / subject)
```
danielinteractive, line 193:
```suggestion
#' sp_exp(coord1, coord2 | group / subject)
```

R/cov_struct.R
```diff
@@ -0,0 +1,289 @@
...
+#' ```
+#'
+#' Note that only `sp_exp` (spatial) covariance structures may provide multiple
+#' longitudinal variables.
```
danielinteractive, line 197:
```suggestion
#' coordinates, which identify the Euclidean distance between the time points.
```

R/fit.R
```diff
@@ -30,9 +30,17 @@
 fit_single_optimizer <- function(formula,
                                  data,
                                  weights,
+                                 covariance = as.cov_struct(
```
danielinteractive, line 33: maybe move after `reml`

R/tmb.R
```diff
@@ -1,183 +1,39 @@
...
-    )
-  }
+  full_formula <- add_covariance_variable_terms(formula, covariance)
+
   structure(
     list(
-      formula = formula,
-      model_formula = model_formula,
+      # formula = formula,
+      model_formula = formula,
```
danielinteractive, line 30: I guess we need to be careful here that we don't break anything downstream

R/utils-validation.R
```diff
@@ -0,0 +1,75 @@
...
+#' @param ...
+#'   A message to emit
+#' @param envir
+#'   The environment in which a hidden `..checks..` object should be
+#'   accumulated. Defaults to the parent frame.
+#'
+#' @return The condition used to decide whether to emit a message
+#'
+#' @keywords internal
+check <- function(severity, cond, ..., envir = parent.frame()) {
```
danielinteractive, line 16: let's not go into this here pls. We can first have a discussion issue on this kind of stuff.

dgkf: Yep - see my TODO. I've since replaced it with `checkmate::makeAssertCollection`. I'll circle back to the rest of your comments after I wrap the the TODOs, but I think I've already addressed the majority.

R/tmb.R
dgkf: I scanned the repo and the first `formula` was never used. Still need to confirm in testing, but I'd prefer to prune some dead code if we can.

R/cov_struct.R
```diff
@@ -0,0 +1,289 @@
+#' Covariance Types
+#'
+#' An internal constant for covariance type lookup
```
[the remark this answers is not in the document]
dgkf: More docs to follow! I've since revamped this because I didn't feel like the original structure was friendly enough for maintenance.

R/fit.R
```diff
@@ -315,9 +328,11 @@ mmrm_control <- function(n_cores = 1L,
 mmrm <- function(formula,
                  data,
                  weights = NULL,
+                 covariance = as.cov_struct(formula, warn_partial = FALSE),
```
danielinteractive, line 331: then also move here after `reml` to be consistent

R/tmb.R
```diff
@@ -1,183 +1,41 @@
...
-
 #' Processing the Formula for `TMB` Fit
 #'
-#' @param formula (`formula`)\cr original formula.
+#' @param formula (`formula`)\cr
+#'   original formula.
+#' @param covariance (`cov_struct`)\cr
+#'   A covariance structure from which additional formula parts should be added.
 #'
 #' @return List of class `mmrm_tmb_formula_parts` with elements:
```
danielinteractive, line 8: seems you keep `formula` now which I think is good for downstream, please keep then the documentation bullet too

R/tmb.R
```diff
@@ -562,22 +419,32 @@ h_mmrm_tmb_fit <- function(tmb_object,
 #' try different optimizers or adds Jacobian information etc. in contrast to
 #' [mmrm()].
 #'
-#' @param formula (`formula`)\cr model formula with exactly one special term
-#'   specifying the visits within subjects, see details.
-#' @param data (`data.frame`)\cr input data containing the variables used
-#'   in `formula`.
-#' @param weights (`vector`)\cr input vector containing the weights.
+#' @param formula (`formula`)\cr
```
danielinteractive, line 422: remember my comment about please keeping current documentation convention, i.e. no newline after `\cr`

tests/testthat/test-cov-struct.R
```diff
@@ -0,0 +1,114 @@
+test_that("cov_struct can be produced from string variable names", {
```
danielinteractive, line 1: please add section dividers `# cov_struct ----` etc in the test files, keeping with the conventions

tests/testthat/test-utils-checkmate.R
```diff
@@ -0,0 +1,30 @@
...
+
+  # confirm undesirable checkmate behavior. if this ever fails, we can probably
+  # remove `with_error_call`
+  expect_true({
+    f <- function() checkmate::reportAssertions(checks)
+    e <- tryCatch(f(), error = identity)
+    deparse(e$call[[1]])[[1]] != "f"
+  })
+
+  # error reports correct calling function
```
danielinteractive, line 13: please remember: sentence style for comments

R/cov_struct.R
```diff
@@ -0,0 +1,442 @@
...
+
+    if (!type %in% cov_types(filter = "spatial") && length(visits) > 1) {
+      checks$push(paste(
+        "Non-spatial covariance structures must have a single longitudinal",
+        "variable"
+      ))
+    }
+  })
+
+  with_error_call(reportAssertions(checks), sys.call(-1))
```
danielinteractive, line 292: this is nice. however I would still like to keep the current assertion structure for now, and have a discussion first whether we want to go into this direction. can you please take that code out of this PR and move on a separate branch/ draft PR?

danielinteractive (commented):
thanks @dgkf , please see inside. As mentioned before, let's please keep the scope focused on the new feature now. in particular, the assertion additions need a separate discussion/PR. thank you

R/cov_struct.R
```diff
@@ -0,0 +1,289 @@
+#' Covariance Types
+#'
+#' An internal constant for covariance type lookup
+#'
+#' @keywords internal
+COV_TYPES <- matrix(  # nolint
```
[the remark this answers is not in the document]
dgkf: I generally prefer `ALL_CAPS` for data constants and `.fn` for exported, but "hidden" functions. I think this distinction is helpful for diving into a codebase. I would prefer to keep this as-is.

R/fit.R
```diff
@@ -252,11 +252,10 @@ mmrm_control <- function(n_cores = 1L,
 #'   the fitting process. Should be `NULL` or a numeric vector.
 #' @param reml (`flag`)\cr whether restricted maximum likelihood (REML)
 #'   estimation is used, otherwise maximum likelihood (ML) is used.
-#' @param covariance (`cov_struct`)\cr
-#'   A covariance structure type definition as produced with [cov_struct()],
-#'   or value that can be coerced to a covariance structure using
-#'   [as.cov_struct()]. If no value is provided, alstructure is derived from
-#'   the provided formula.
+#' @param covariance (`cov_struct`)\cr A covariance structure type definition
```
danielinteractive, line 255:
```suggestion
#' @param covariance (`cov_struct`)\cr a covariance structure type definition
```
after the `@param` etc we start with lower case, end with full stop.

but thanks for getting closer already! almost there

dgkf: `%s/\(@param.*\\cr \)\(.\)/\1\L\2/g`'d all my files.

NAMESPACE
```diff
@@ -52,6 +52,8 @@ importFrom(stats,model.frame)
 importFrom(stats,terms)
 importFrom(stats,vcov)
 importFrom(stringr,boundary)
+importFrom(utils,capture.output)
```
danielinteractive, line 55: can we please use `utils::capture.output` etc. i.e. only import methods when needed, for normal functions just use `::`

NAMESPACE
```diff
@@ -31,6 +31,7 @@ export(mmrm)
 export(mmrm_control)
 export(refit_multiple_optimizers)
 import(checkmate)
+import(utils)
```
danielinteractive, line 34: ahm no, we don't want to import the whole package. imports should be as minimal as possible. especially here we just use 5 functions so far from the pkg.

tests/testthat/test-fit.R
```diff
@@ -70,15 +70,17 @@ test_that("fit_single_optimizer gives error messages", {
...
   expect_error(
     fit_single_optimizer(formula, fev_data, weights = rep(1, nrow(fev_data))),
     paste(
       "Covariance structure must be specified in formula.",
-      "Possible covariance structures include: us, toep, toeph, ar1, ar1h, ad, adh, cs, csh"
+      "Possible covariance structures include:",
+      "us, toep, toeph, ar1, ar1h, ad, adh, cs, csh"
     ),
     fixed = TRUE
   )
```
danielinteractive, line 86: Could you add tests for:
1. Using `covariance` from `mmrm()` (highest level user interface)
2. Using both a `covariance` argument and a formula specification (to see if this fails as expected)

danielinteractive (approved):
Thanks a lot Doug, great work! Quite heavy lifting here, thanks for pulling this through.
Good to merge after additional tests mentioned inside are in.

One additional comment: It would be great to also mention this in the introduction vignette. But given that we don't mention tidymodels there yet at all, it is probably best to have a new issue and PR for that later.

R/utils-formula.R
```diff
@@ -0,0 +1,58 @@
+#' Extract Formula Terms used for Covariance Structure Definition
+#'
+#' @param f (`formula`)\cr a formula from which covariance terms should be
+#'   extracted.
+#'
+#' @return A list of covariance structure expressions found in `f`.
+#'
+#' @importFrom stats terms
+#' @keywords internal
+extract_covariance_terms <- function(f) {
```
clarkliming, line 10: a formula like `a ~ a+ ar1(a,b|c/d) ` will give error because the term in the left is also in the right part; using the `length(f)>` to decide the position is not robust here

clarkliming: of cource `a ~ a + ar1(a,b|c,d)` should not be used; but this is a legal formula

clarkliming:
I really like the `cov_struct` here and it possibly enables #174 in the future

PR #238, opened 2023-02-15
fix na errors

### LINKED ISSUE #237: na.action option
**Summary**

Your brief description of the problem

Cannot compile when na.action option is "na.pass"
Is it not necessary to depend on the value of the na.action option?
```r

# your reproducible example here
options(na.action = "na.pass")
mmrm(FEV1 ~ ARMCD * AVISIT + us(AVISIT | USUBJID), data = fev_data)
```

**R session info**

```sh

# R -e "utils::sessionInfo()" output goes here
Error in fit_single_optimizer(formula = formula, data = data, weights = weights,  : 
  negative log-likelihood is NaN at starting parameter values
```

**OS / Environment**

- OS: Windows 11

clarkliming:
close #237 

na.action is always na.omit after the update

### DIFF
--- a/R/tmb.R
+++ b/R/tmb.R
@@ -129,7 +129,11 @@ h_mmrm_tmb_data <- function(formula_parts,
   # weights is always the last column
   weights_name <- colnames(data)[ncol(data)]
   full_frame <- eval(
-    bquote(stats::model.frame(formula_parts$full_formula, data = data, weights = .(as.symbol(weights_name))))
+    bquote(stats::model.frame(
+      formula_parts$full_formula, data = data,
+      weights = .(as.symbol(weights_name)),
+      na.action = na.omit
+    ))
   )
   full_frame <- droplevels(full_frame, except = formula_parts$visit_var)
   if (drop_visit_levels && !formula_parts$is_spatial && is.factor(full_frame[[formula_parts$visit_var]])) {
--- a/R/tmb.R
+++ b/R/tmb.R
@@ -20,9 +20,9 @@
 #'   this element is `NULL`.
 #'
 #' @keywords internal
-h_mmrm_tmb_formula_parts <- function(formula,
+h_mmrm_tmb_formula_parts <- function(
+    formula,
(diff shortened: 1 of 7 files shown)


R/tmb.R
```diff
@@ -129,7 +129,12 @@ h_mmrm_tmb_data <- function(formula_parts,
   # weights is always the last column
   weights_name <- colnames(data)[ncol(data)]
   full_frame <- eval(
```
danielinteractive, line 131: I wonder if we should raise a warning if the global option is set differently than na.omit?

danielinteractive (commented):
Thanks Liming, very nice! See one question inside

R/tmb.R
clarkliming: it is good idea to pop up the error here; however, this can be quite a large update; I can open new issues to address. The reason that I say this is quite large because currently `h_mmrm_tmb_data` is called multiple times in different fit (it should be done only once as the data should be the same, but the optimizer changes). If the warning messages are created, then in every fit try, the warnings are captured by `h_record_all_output` and only when number of warnings is 0 we say this is converged/ or we handle that warning so that make the converge but this message will be discarded in `h_record_all_output` just like "NA/NAN function evaluation" . So having a warning here will not display the message directly, but difficult to handle. We need to separate the "h_mmrm_tmb_data` and `fit_single_optimizer`, but this will make `fit_single_optimizer` a big breaking change (it could not be called directly, it need to be feeded with tmb_data) and I doubt if it is possible

clarkliming: instead, we can check the options in `mmrm` call, or in `fit_single_optimizer` call. But this is really not the best place to give the warning message because it is in`h_mmrm_tmb_data` the `na.action` is used and checked.

clarkliming: oh probably message can be used. let me check...

danielinteractive: ok that would be good then

danielinteractive (approved):

PR #239, opened 2023-02-15
220 time as factor

### LINKED ISSUE #220: Clarify in documentation that non-spatial covariance structure needs factor variable
Motivation: As shown in https://github.com/openpharma/mmrm/issues/217 it is not obvious right now that time point variable needs to be factor variable for non-spatial covariance structures.

To do:

- [x] Clarify in
	- [x] intro vignette
	- [x] `mmrm()` doc
	- [x] `cov_type` doc
- [x] Improve error message: Currently the user gets `Assertion on 'coordinates[[1L]]' failed: Must be of type 'factor', not 'double'` and they should get something like `Time points variable needs to be a factor`

danleibovitz:
closes #220 

- Inserts a check and corresponding error message in R/tmb.R, within the fit_mmrm() function; add a corresponding test in test-tmb.R under 'misc'.
- Updates the documentation in R/fit.R for mmrm(), in cov_struct.R for cov_types, and in the intro vignette.

### DIFF
--- a/R/cov_struct.R
+++ b/R/cov_struct.R
@@ -26,7 +26,7 @@
 #' }
 #'
 #' @keywords internal
-COV_TYPES <- local({  # nolint
+COV_TYPES <- local({ # nolint
   type <- function(name, abbr, habbr, heterogeneous, spatial) {
     args <- as.list(match.call()[-1])
     do.call(data.frame, args)
@@ -35,12 +35,12 @@ COV_TYPES <- local({  # nolint
   as.data.frame(
     col.names = names(formals(type)),
     rbind(
-      type("unstructured",              "us",     NA,      FALSE, FALSE),
-      type("Toeplitz",                  "toep",   "toeph", TRUE,  FALSE),
-      type("auto-regressive order one", "ar1",    "ar1h",  TRUE,  FALSE),
-      type("ante-dependence",           "ad",     "adh",   TRUE,  FALSE),
-      type("compound symmetry",         "cs",     "csh",   TRUE,  FALSE),
-      type("spatial exponential",       "sp_exp", NA,      FALSE, TRUE)
+      type("unstructured", "us", NA, FALSE, FALSE),
+      type("Toeplitz", "toep", "toeph", TRUE, FALSE),
+      type("auto-regressive order one", "ar1", "ar1h", TRUE, FALSE),
+      type("ante-dependence", "ad", "adh", TRUE, FALSE),
+      type("compound symmetry", "cs", "csh", TRUE, FALSE),
+      type("spatial exponential", "sp_exp", NA, FALSE, TRUE)
     )
   )
 })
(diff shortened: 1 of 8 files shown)


R/cov_struct.R
```diff
@@ -133,6 +133,9 @@ COV_TYPES <- local({  # nolint
 #' from SAS `PROC MIXED` refers to heterogeneous ante-dependence and the
 #' homogeneous version is not available in SAS.
 #'
+#' Note that for all non-spatial covariance structures, the time variable must
```
danielinteractive, line 136: maybe add this and above note under `@note` section

R/tmb.R
```diff
@@ -472,6 +472,8 @@ fit_mmrm <- function(formula,
   covariance <- reconcile_cov_struct(formula, covariance)
   formula_parts <- h_mmrm_tmb_formula_parts(formula, covariance)
 
+  if (!formula_parts$is_spatial && !is.factor(data[[formula_parts$visit_var]])) stop("Time variable must be a factor for non-spatial covariance structures")
```
danielinteractive, line 475:
```suggestion
  if (!formula_parts$is_spatial && !is.factor(data[[formula_parts$visit_var]])) {
    stop("Time variable must be a factor for non-spatial covariance structures")
  }
```

R/fit.R
```diff
@@ -359,6 +360,8 @@ mmrm <- function(formula,
   if (control$method %in% c("Kenward-Roger", "Kenward-Roger-Linear") && !reml) {
     stop("Kenward-Roger only works for REML")
   }
+  # TODO insert informative error message if cov struc is non-spatial and time point variables is not factor
```
danielinteractive, line 363:
```suggestion
```

R/tmb.R
danielinteractive: I wonder though, now that we have the `validate_cov_struct()` function, would this assertion better fit there? can you please check?

danielinteractive (commented):
thanks a lot @danleibovitz ! please see inside

R/cov_struct.R
danleibovitz: Good point, done.

R/tmb.R
danleibovitz: I don't think the `validate_cov_struct()` as is helps us here because the object is valid by construction. We could: 

1. Write a new `checkmate` style check, e.g., `validate_time_as_factor()`,
2. Include a new check within the `validate_cov_struct()` definition, and then run it here.

R/fit.R
danleibovitz: This is a silly oversight on my part -- text removed.

R/tmb.R
danielinteractive: Hm ok, then let's keep it as is. Thanks for checking!

danielinteractive (approved):

danielinteractive (approved):

PR #235, opened 2023-02-08
Pkgdown dropdowns

cicdguy:
This will create docs and other relevant artifacts for future releases.

PR #240, opened 2023-02-16
Fix pkgdown workflow

walkowif:
Fixes issue reported [here](https://github.com/openpharma/mmrm/pull/235#issuecomment-1432663383).


clarkliming (approved):

clarkliming:
thank you, this should work now

PR #244, opened 2023-02-28
219 confirmation on large tp

### LINKED ISSUE #219: Ask for confirmation to fit model if more than xxx time points (for non-spatial)
Background: Currently if you try to fit a model with say 800 time points your R session will crash because of insufficient memory. Typically non-spatial covariance structure model are used for less than say 100 time points only, so we could ask for user confirmation before trying to fit a model with more time points.

To do:
- [x] Find the right place in the code where to check this
- [x] Ask for user confirmation with simple question and expecting "yes" for confirmation
- [x] add test
- [x] add NEWS entry
- [x] add note of this to introduction vignette

clarkliming:
close #219 

`fit_mmrm` , `fit_mmrm_single` are updated to receive `tmb_data` and `formula_parts` to avoid repetative computations

`mmrm.max_visits` is used to determine the maximum number of visit levels allowed.


R/fit.R
```diff
@@ -7,12 +7,16 @@
 #'
 #' @inheritParams mmrm
 #' @param control (`mmrm_control`)\cr object.
+#' @param tmb_data (`mmrm_tmb_data`)\cr object.
```
clarkliming, line 10: this is added to prevent `h_mmrm_tmb_data` called multiple times in `mmrm`

R/fit.R
```diff
@@ -32,27 +36,46 @@ fit_single_optimizer <- function(formula,
...
-      reml = reml,
-      covariance = covariance,
-      control = control
-    ),
-    remove = list(
-      warnings = c("NA/NaN function evaluation") # Transient visit to invalid parameters.
+  if (missing(tmb_data) || missing(formula_parts)) {
+    assert_formula(formula)
+    assert_data_frame(data)
+    assert_numeric(weights, any.missing = FALSE, lower = .Machine$double.xmin)
```
clarkliming, line 46: check weights as positive values

R/fit.R
```diff
@@ -32,27 +36,46 @@ fit_single_optimizer <- function(formula,
...
+    assert_flag(reml)
+    assert_class(control, "mmrm_control")
+    assert_list(control$optimizers, names = "unique", types = c("function", "partial"))
+    quiet_fit <- h_record_all_output(
+      fit_mmrm(
```
clarkliming, line 51: `fit_mmrm` is updated to accept two types of argument, and `fit_mmrm` still works as low level functionality

R/fit.R
```diff
@@ -145,9 +168,8 @@ refit_multiple_optimizers <- function(fit,
     FUN = fit_single_optimizer,
     control = controls,
     MoreArgs = list(
-      formula = old_formula,
-      data = old_data,
-      weights = old_weights,
+      tmb_data = fit$tmb_data,
+      formula_parts = fit$formula_parts,
       reml = fit$reml
```
clarkliming, line 173: reml can be removed as tmb_data already has it

R/fit.R
```diff
@@ -403,6 +427,8 @@ mmrm <- function(formula,
   if (!is.null(fit_msg)) {
     message(paste(fit_msg, collapse = "\n"))
   }
+  fit$call <- match.call()
+  fit$call$formula <- formula
```
clarkliming, line 431: obtain the formula so that it could be printed, instead of a object.

R/mmrm-methods.R
```diff
@@ -122,16 +122,16 @@ h_print_call <- function(call, n_obs, n_subjects, n_timepoints) {
...
       n_subjects, "subjects with maximum", n_timepoints, "timepoints)",
       fill = TRUE
     )
   }
-  # Only if call$weights is a string then it was explicitly given by the user.
-  if (test_string(tmp <- call$weights)) {
-    cat("Weights:    ", tmp, fill = TRUE)
+  # Display the expression of weights
+  if (!is.null(call$weights)) {
+    cat("Weights:    ", deparse(call$weights), fill = TRUE)
```
clarkliming, line 134: weights can be something like `rep(1, nrow(data)` so have to deparse it

R/tmb.R
```diff
@@ -514,20 +524,9 @@ fit_mmrm <- function(formula,
...
-    attr(data, which = "dataname"),
-    toString(match.call()$data)
-  )
-  weights_str <- attr(weights, which = "dataname")
-  if (!is.null(weights_str)) {
-    fun_call$weights <- weights_str
-  }
-  fit$call <- fun_call
+  fit$call <- match.call()
+  fit$call$formula <- formula_parts$formula
```
clarkliming, line 530: for the data, use the symbol directly; in `mmrm` the call is also retrieved. The reason to have two locations of call is that `fit_mmrm` is still exported so we need to use it.  For weights, I am just wondering if a "string" indicating the column in the data should be better than providing a numeric value

clarkliming:
how to trigger the actions manualy? in https://github.com/openpharma/mmrm/actions/runs/4292615230 all green

danielinteractive:
@clarkliming just another manual (also empty is ok) commit that is pushed on this branch will retrigger checks

R/fit.R
danielinteractive: nice

R/fit.R
danielinteractive: hm but it would be better if we keep named arguments in this `fit_mmrm` call?

R/fit.R
```diff
@@ -32,27 +36,46 @@ fit_single_optimizer <- function(formula,
...
+        control = control
+      ),
+      remove = list(
+        warnings = c("NA/NaN function evaluation") # Transient visit to invalid parameters.
+      )
     )
-  )
+  } else {
+    checkmate::assert_class(tmb_data, "mmrm_tmb_data")
+    checkmate::assert_class(formula_parts, "mmrm_tmb_formula_parts")
```
danielinteractive, line 65:
```suggestion
    assert_class(tmb_data, "mmrm_tmb_data")
    assert_class(formula_parts, "mmrm_tmb_formula_parts")
```

R/fit.R
danielinteractive: but now just `weights` is removed?

R/fit.R
```diff
@@ -368,13 +390,15 @@ mmrm <- function(formula,
   } else {
     attr(weights, which = "dataname") <- deparse(match.call()$weights)
   }
-
+  covariance <- reconcile_cov_struct(formula, covariance)
```
danielinteractive, line 393: is that a helper function call? then better prefix with `h_`

R/fit.R
danielinteractive: ok

R/mmrm-methods.R
danielinteractive: yeah this was a bit hacky, so maybe this is a bit better

R/tmb.R
```diff
@@ -468,22 +471,29 @@ fit_mmrm <- function(formula,
...
+    assert_list(control$optimizers, min.len = 1)
+    assert_numeric(weights, any.missing = FALSE)
+    assert_true(all(weights > 0))
+    tmb_data <- h_mmrm_tmb_data(
+      formula_parts, data, weights, reml,
+      accept_singular = control$accept_singular, drop_visit_levels = control$drop_visit_levels
+    )
+  } else {
+    checkmate::assert_class(tmb_data, "mmrm_tmb_data")
+    checkmate::assert_class(formula_parts, "mmrm_tmb_formula_parts")
```
danielinteractive, line 495:
```suggestion
    assert_class(tmb_data, "mmrm_tmb_data")
    assert_class(formula_parts, "mmrm_tmb_formula_parts")
```

R/tmb.R
danielinteractive: hm interesting. I remember we had some problem with the data thing before. but I guess as long as the tests pass this should be fine.

and then the same logic as for data applies to the weights thing?

R/utils.R
```diff
@@ -296,3 +296,33 @@ fill_names <- function(x) {
...
+#' Ask for Confirmation on Large Visit Levels
+#'
+#' @description Ask the user for confirmation if there are too many visit levels
+#' for non-spatial covariance structure in interactive sessions.
+#'
+#' @param x (`numeric`)\cr number of visit levels.
+#'
+#' @keywords internal
+h_confirm_large_levels <- function(x) {
+  checkmate::assert_numeric(x)
```
danielinteractive, line 309:
```suggestion
  assert_count(x)
```

tests/testthat/_snaps/tmb-methods.md
```diff
@@ -5,6 +5,7 @@
     Formula:     FEV1 ~ RACE + us(AVISIT | USUBJID)
     Data:        fev_data (used 537 observations from 197 subjects with maximum 4 
     timepoints)
+    Weights:     rep(1, nrow(fev_data))
```
danielinteractive, line 8: hm this is not very nice. I think therefore I had some additional logic before to circumvent this from being printed

vignettes/introduction.Rmd
```diff
@@ -273,6 +273,16 @@ and `VIS4` the same as the correlation between `VIS1` and `VIS2`. Hence we get a
...
 
+## Large Number of Visit Levels for non-spatial Covariance
+
+When the number of visit levels is large, it usually require large memory to create the
+covariance matrix. By default, the maximum allowed visit levels is 100, and if there are more
+visit levels, a confirmation is needed if run interactively.
+You can use `options(mmrm.max_visits = <target>)` to increase the maximum allowed number of visit
+levels.
+In non-interactive sessions the confirmation is not raised and will directly give you error if
+that number of visit levels exceed the maximum.
```
danielinteractive, line 284: hm I think this is maybe a bit too detailed for the introduction vignette. Maybe just move this to the `mmrm()` function documentation?

danielinteractive (commented):
thanks a lot Liming! please see few comments inside

tests/testthat/_snaps/tmb-methods.md
clarkliming: I think previously only when `weights` is assigned (not rep(1, nrow(data)) then it is displayed. do you mean this https://github.com/openpharma/mmrm/blob/b3f7e5c7d53ee605e4b2041aba48a3f9c42b7878/R/fit.R#L369

so in previous logic, weights get displayed when:

in `mmrm`, weights argument is used, not NULL; in `fit_mmrm`, no matter if weights is assigned, it is not displayed

here the change is that in `fit_mmrm` the assigned `weights` still get displayed. we can of course remove that "weights" for `fit_mmrm$call`. but why we should do that? in `mmrm` by default the weights are not displayed

R/fit.R
clarkliming: yes; `reml` is already part of `tmb_data`

tests/testthat/_snaps/tmb-methods.md
danielinteractive: ah ok cool - so when we use `mmrm()` without explicit weights assigned it is not displayed still right? then it is ok.

R/fit.R
```diff
@@ -345,6 +344,13 @@ mmrm_control <- function(n_cores = 1L,
 #'
 #' NA values are always omitted regardless of `na.action` setting.
 #'
+#' When the number of visit levels is large, it usually require large memory to create the
+#' covariance matrix. By default, the maximum allowed visit levels is 100, and if there are more
+#' visit levels, a confirmation is needed if run interactively.
+#' You can use `options(mmrm.max_visits = <target>)` to increase the maximum allowed number of visit
+#' levels. In non-interactive sessions the confirmation is not raised and will directly give you error if
+#' that number of visit levels exceed the maximum.
```
danielinteractive, line 352:
```suggestion
#' When the number of visit levels is large, it usually requires large memory to create the
#' covariance matrix. By default, the maximum allowed visit levels is 100, and if there are more
#' visit levels, a confirmation is needed if run interactively.
#' You can use `options(mmrm.max_visits = <target>)` to increase the maximum allowed number of visit
#' levels. In non-interactive sessions the confirmation is not raised and will directly give you an error if
#' the number of visit levels exceeds the maximum.
```

danielinteractive (approved):
thanks Liming!

tests/testthat/_snaps/tmb-methods.md
clarkliming: it only gets printed when you have "weights" argument specified; e.g. using `mmrm(formula, data = fev_data)` you don't have the weights printed. If you have `mmrm(weights = xxx)` it got printed (this is the same for previous version)

clarkliming: (and that's why `fit_mmrm` and `mmrm` both keep a `call` but they always use the call they received)

danielinteractive (approved):

PR #248, opened 2023-03-08
update question about brl

clarkliming:
add vignettes about the difference in BRL sandwich estimator

### DIFF
--- a/src/empirical.cpp
+++ b/src/empirical.cpp
@@ -1,9 +1,11 @@
 #include "derivatives.h"
+#include <unsupported/Eigen/MatrixFunctions>
+#include <chrono>
 
 using namespace Rcpp;
 using std::string;
 // Obtain the empirical given beta, beta_vcov, theta.
-List get_empirical(List mmrm_data, NumericVector theta, NumericVector beta, NumericMatrix beta_vcov, bool jackknife) {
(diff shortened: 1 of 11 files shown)


src/empirical.cpp
```diff
@@ -58,32 +65,43 @@ List get_empirical(List mmrm_data, NumericVector theta, NumericVector beta, Nume
...
     int n_visits_i = subject_n_visits[i];
-    g.block(0, i * p, n_observations, p) = imh.block(0, start_i, n_observations, n_visits_i) * ax_xtx.block(start_i, 0, n_visits_i, p);
+    matrix<double> ai = matrix<double>::Identity(n_visits_i, n_visits_i);
+    if (type == 1) { // 1 for jackknife
+      ai = (ai - hxw.block(start_i, start_i, n_visits_i, n_visits_i)).inverse();
+    } else if (type == 2) {
+      matrix<double> di = chols.block(start_i, start_i, n_visits_i, n_visits_i);
+      matrix<double> bi1 = di * i_hxw.block(start_i, 0, n_visits_i, n_observations);
+      matrix<double> bi = bi1 * phi * bi1.transpose();
+      matrix<double> bisq = matrix<double>(bi.completeOrthogonalDecomposition().pseudoInverse()).sqrt();
```
clarkliming: it was used here but seems not very handy (sqrt is from eigen/unspported); this part returns some random nan but not sure why this happens

PR #252, opened 2023-03-20
update model.frame and with functionality

### LINKED ISSUE #243: `emmeans()` fails when covariates are transformed in the model formula
**Summary**

`emmeans()` fails when covariates are transformed in the model formula.
It seems that `emmeans()` then looks for the original covariate but cannot find it. 
I wonder if we lose the original covariate when working with the model frames inside
our `emmeans` interface.
On the other hand, this works fine with `lm()` and `glmmTMB()` so we can compare our interface
implementation with that as well to find the cause and fix it.

To do:

- [x] Include below or shortened example as unit test to reproduce the bug
- [x] Find the root cause
- [x] Fix it
- [x] Make sure all tests pass
- [x] Mention under bug fixes in the `NEWS` file

**Reprex**

```r
library(mmrm)
library(emmeans)

ana_dat <- data.frame(
  expand.grid(
    USUBJID = factor(seq(1:100)), 
    AVISIT = c("V1", "V2")
  ), 
  Y = rnorm(200), 
  X = runif(200, 1, 5)
)

# Continuous covariate without transformation:
fit1 <- mmrm(Y ~ X + AVISIT + us(AVISIT | USUBJID), data = ana_dat)
emmeans(fit1, ~ AVISIT, adjust = "none")
# Works Well !

# Continuous covariate with transformation:
fit2 <- mmrm(Y ~ log(X) + AVISIT + us(AVISIT | USUBJID), data = ana_dat)
emmeans(fit2, ~ AVISIT, adjust = "none")
# Error in eval(predvars, data, env) : object 'X' not found
# Error in (function (object, at, cov.reduce = mean, cov.keep = get_emm_option("cov.keep"),  : 
#  Perhaps a 'data' or 'params' argument is needed

# Linear model:
lm1 <- lm(Y ~ log(X) + AVISIT, data = ana_dat)
emmeans(lm1, ~ AVISIT, adjust = "none")
# Works

# glmmTMB:
library(glmmTMB)
glmmtmb1 <- glmmTMB(Y ~ log(X) + AVISIT + us(AVISIT | USUBJID), data = ana_dat)
emmeans(glmmtmb1, ~ AVISIT, adjust = "none")
# Also works
```

**R session info**

```sh
R version 4.2.2 (2022-10-31)
Platform: x86_64-apple-darwin17.0 (64-bit)
Running under: macOS Monterey 12.6.1

Matrix products: default
LAPACK: /Library/Frameworks/R.framework/Versions/4.2/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] glmmTMB_1.1.4   emmeans_1.7.5   mmrm_0.2.2.9011

loaded via a namespace (and not attached):
 [1] Rcpp_1.0.9          rstudioapi_0.14     TMB_1.9.1           magrittr_2.0.3      splines_4.2.2       MASS_7.3-58.1      
 [7] xtable_1.8-4        lattice_0.20-45     rlang_1.0.6         minqa_1.2.4         stringr_1.4.1       tools_4.2.2        
[13] rbibutils_2.2.8     parallel_4.2.2      grid_4.2.2          checkmate_2.1.0     nlme_3.1-160        cli_3.4.0          
[19] coda_0.19-4         lme4_1.1-30         lifecycle_1.0.2     numDeriv_2016.8-1.1 Matrix_1.5-1        nloptr_2.0.3       
[25] Rdpack_2.4          stringi_1.7.8       estimability_1.4    compiler_4.2.2      backports_1.4.1     boot_1.3-28        
[31] mvtnorm_1.1-3  
```

clarkliming:
close #243 

also enables creating `mmrm` with variables from envrionment instead of providing data argument

further solve the issue

```{r}
fit <- mmrm(FEV1 ~ ARMCD + ar1(AVISIT|USUBJID), data = fev_data)
emmeans(fit, ~ ARMCD | AVISIT)
```
previously, `model.frame.mmrm` does not return AVISIT if it is not included in fixed effects. This requires update in `model.frame`.

also enables using variables from both environment and data
```{r}
x <- rnorm(800)
mmrm(x ~ ARMCD + ar1(AVISIT|USUBJID), data = fev_data) # this works now!
```


R/fit.R
```diff
@@ -391,16 +391,20 @@ mmrm <- function(formula,
...
     stop("Kenward-Roger only works for REML")
   }
+  covariance <- h_reconcile_cov_struct(formula, covariance)
+  formula_parts <- h_mmrm_tmb_formula_parts(formula, covariance)
 
-  attr(data, which = "dataname") <- toString(match.call()$data)
+  if (!missing(data)) {
+    attr(data, which = "dataname") <- toString(match.call()$data)
+  } else {
+    data <- model.frame(formula_parts$full_formula)
```
clarkliming, line 400: please note that the environment of the formula matters. So here in `h_mmrm_tmb_formula_parts` we are not creating new formulas but updating the original formula is good: the environment is not affected

R/mmrm-package.R
```diff
@@ -16,5 +16,6 @@
 #' @importFrom stringr boundary
 #' @importFrom parallel clusterApply
 #' @importFrom Rdpack reprompt
+#' @importFrom rlang %||%
```
clarkliming, line 19: `%||%` can be replaced by something like
```{r}
my_fun <- function(a,b) {
if (is.null(a)) b
else a
}
```

but here we can reduce the code/maintanence with rlang

R/tmb-methods.R
```diff
@@ -50,12 +50,34 @@ fitted.mmrm_tmb <- function(object, ...) {
...
 model.frame.mmrm_tmb <- function(formula, full = FALSE, ...) {
   assert_flag(full)
-  if (full) {
+  dots <- list(...)
+  if (!identical(dots$na.action %||% getOption("na.action"), "na.omit")) {
+    warning("na.action is always set to `na.omit` for `mmrm`!")
+  }
+  if (is.null(dots$data)) {
+    if (full) {
+      formula$tmb_data$full_frame
```
clarkliming, line 59: here we can further simplify the logic: if data is not provided, we always return `formula$tmb_data$full_frame`, with selected cols. If `full`, we provide all columns. If not `full`, we remove visits and usubjids.

danielinteractive (commented):
Thanks Liming, nice work! Can you please also add the `emmeans()` bug as kind of integration test? thanks

R/cov_struct.R
```diff
@@ -272,7 +272,7 @@ cov_struct <- function(
 #'
 #' @keywords internal
 h_reconcile_cov_struct <- function(formula = NULL, covariance = NULL) {
-  if (!is.null(covariance) && length(extract_covariance_terms(formula)) > 0) {
+  if (!is.null(covariance) && length(h_extract_covariance_terms(formula)) > 0) {
```
clarkliming, line 275: simple rename of helper functions

R/fit.R
```diff
@@ -391,16 +391,21 @@ mmrm <- function(formula,
...
   }
+  covariance <- h_reconcile_cov_struct(formula, covariance)
+  formula_parts <- h_mmrm_tmb_formula_parts(formula, covariance)
 
-  attr(data, which = "dataname") <- toString(match.call()$data)
+  if (!missing(data)) {
+    attr(data, which = "dataname") <- toString(match.call()$data)
+  } else {
+    # na.action set to na.pass to allow data to be full; will be futher trimmed later
+    data <- model.frame(formula_parts$full_formula, na.action = "na.pass")
```
clarkliming, line 401: create data will use full data (because the length of the variable should match from the beginning)

R/interop-emmeans.R
```diff
@@ -24,13 +24,12 @@ NULL
...
   fun_call <- stats::getCall(object)
-  model_frame <- stats::model.frame(object)
+  model_frame <- stats::model.frame(object, exclude = "subject_var")
   model_terms <- stats::delete.response(stats::terms(model_frame))
-  na_action <- attr(model_frame, "na.action")
   emmeans::recover_data(
     fun_call,
     trms = model_terms,
-    na.action = na_action,
+    na.action = "na.omit",
```
clarkliming, line 32: I think we still make sure na.action is "na.omit" in every cases

R/interop-emmeans.R
```diff
@@ -56,7 +55,7 @@ emm_basis.mmrm <- function(object, # nolint
     beta_hat[kept] <- component(object, "beta_est")
     orig_model_mat <- stats::model.matrix(
       trms,
-      stats::model.frame(object),
+      stats::model.frame(object, full = TRUE),
```
clarkliming, line 58: this is not correct probably

R/tmb-methods.R
```diff
@@ -41,21 +41,36 @@ fitted.mmrm_tmb <- function(object, ...) {
...
+  if (!is.null(dots$subset) || !is.null(dots$weights)) {
+    warning("subset and weights are not valid arguments for `mmrm` models.")
+  }
+  if (is.null(dots$data) && length(exclude) == 0L) {
     formula$tmb_data$full_frame
   } else {
+    drop_vars <- unlist(
+      lapply(
+        exclude,
+        function(i) formula$formula_parts[[i]]
```
clarkliming, line 66: can be simplified

clarkliming:
Hi @cicdguy there is some error with R cmd check, seems also related to latex? could you please have a look?
https://github.com/openpharma/mmrm/actions/runs/4485615906/jobs/7887336071?pr=252

R/utils-formula.R
```diff
@@ -64,6 +64,10 @@ h_add_covariance_terms <- function(f, covariance) {
 }
 
 #' Drop Formula Terms with Character
+#'
+#' @description `r lifecycle::badge("experimental")`
```
danielinteractive, line 68:
```suggestion
```
(no lifecycle badges for internal functions)

R/utils.R
```diff
@@ -328,7 +328,8 @@ h_confirm_large_levels <- function(x) {
 }
 
 #' Default Value on NULL
-#' @description Return default value when first argument is NULL.
+#' @description `r lifecycle::badge("experimental")`
```
danielinteractive, line 331: same here
```suggestion
```

NEWS.md
```diff
@@ -18,13 +18,23 @@
 - Add confirmation if the number of visit levels is too large. Use `options(mmrm.max_visits = )`
   to specify the maximum number of visits allowed in non-interactive mode, or asked in interactive
   sessions.
+- Previously `model.frame` does not work with new given `data`. It now works with new `data`.
```
danielinteractive, line 21: please integrate this into bugfixes, below

R/interop-emmeans.R
```diff
@@ -24,13 +24,12 @@ NULL
 #' @noRd
 recover_data.mmrm <- function(object, ...) { # nolint
   fun_call <- stats::getCall(object)
-  model_frame <- stats::model.frame(object)
+  model_frame <- stats::model.frame(object, exclude = "subject_var")
```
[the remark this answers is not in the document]
danielinteractive: please add that as a comment above the line of code so it is kept for maintenance information

R/utils-formula.R
```diff
@@ -48,9 +52,37 @@ drop_covariance_terms <- function(f) {
 #'
 #' @return A new formula with included covariance terms.
 #'
+#' @details `update` is used to append the covariance structure and the environment
```
danielinteractive, line 55:
```suggestion
#' @details [stats::update()] is used to append the covariance structure and the environment
```

tests/testthat/test-tmb.R
```diff
@@ -386,42 +387,41 @@ test_that("h_mmrm_tmb_data works even if na.action is not na.omit", {
     drop_visit_levels = TRUE
   )
   options(na.action = "na.pass")
-  res2 <- h_mmrm_tmb_data(
+  res2 <- suppressWarnings(h_mmrm_tmb_data(
```
danielinteractive, line 390: better use `expect_warning()` here and with the concrete warning we expect

tests/testthat/test-tmb.R
```diff
@@ -433,7 +433,7 @@ test_that("h_mmrm_tmb_data errors if too many visit levels", {
     accept_singular = FALSE,
     drop_visit_levels = TRUE
   ), "Visit levels too large!")
-  options("mmrm.max_visits" = NULL)
+  on.exit(options("mmrm.max_visits" = NULL))
```
danielinteractive, line 436: should be moved to line 428?

tests/testthat/test-utils.R
```diff
@@ -153,9 +153,19 @@ test_that("h_get_cov_default works correctly", {
 # h_confirm_large_levels ----
 
 test_that("h_confirm_large_levels errors for large number", {
+  skip_if(interactive())
```
danielinteractive, line 156: why do we need this? test should still be ok to run interactively I think

danielinteractive (commented):
thanks Liming, very nice work, see last comments inside

tests/testthat/test-utils.R
clarkliming: (it will ask you yes/no and if you provided wrong answer tests might fail. but we can make `askpass` always return FALSE use mockery

danielinteractive: ah ok then good as is, most important is that non-interactive tests run

danielinteractive:
@clarkliming can you please resolve comments above? (in general that is up to PR author :-)) thanks

clarkliming:
> @clarkliming can you please resolve comments above? (in general that is up to PR author :-)) thanks

all resolved now

danielinteractive (approved):

Issue #189, opened 2022-12-06
Add `predict()` method for `mmrm_tmb`


danielinteractive:
depends on #263

Motivation: 1) kind of a standard for models, 2) needed for support in the simulation framework [DHARMa](https://github.com/florianhartig/DHARMa/wiki/Adding-new-R-packages-to-DHARMA#integrating-a-package-with-dharma).

To do:

- [ ] Add a short design doc to work out the approach
- [ ] Add method
- [ ] Add tests
- [ ] Add to NEWS

danielinteractive:
@kkmann I guess you are busy with our course :-) let us know if you need start help for this one

danielinteractive:
Hi @kkmann do you think you can work on it this week?

kkmann:
Mea culpa, blocked some time for it next week!

kkmann:
@danielinteractive is there an interface for getting the covariance matrix of one or multiple individuals from a fitted mmrm object? It does not seem that any of the cov functionality is exposed to R at the moment. We would need that for prediction intervals. varcorr only seems to get a static matrix and is not enough for spatial covariance structures.

kkmann:
I added a vignette with a draft description of what we would need to implement this. Again highlighting the connection to Gaussian Processes quite nicely x) #254 

@danielinteractive could you take a look at it to see whether my understanding is correct? 

For standard errors, we would also have to get the Jacobians via AD again - a bit beyond my understanding of the mmrm/tmb code basis. Is there anyone who could help?

danielinteractive:
Thanks @kkmann , yeah we can help, either myself or maybe @clarkliming

Issue #232, opened 2023-02-02
Avoid the need to recalculate any Cholesky factors for normalized residuals


danielinteractive:
Motivation:
- In `h_residuals_pearson()` added by https://github.com/openpharma/mmrm/pull/210 we currently recalculate the Cholesky factors needed to obtain normalized residuals
- However, in https://github.com/openpharma/mmrm/blob/main/src/mmrm.cpp#L102 we actually have for every patient exactly the Cholesky factor we need (including weights, spatial or non-spatial etc.).

To do:
- [x] #264
	- [x] they can be reported back in the `tmb_report` list
	- [x] ~this is not super simple because each patient has different number of time points, so different dimension of the Cholesky factor~
	- [x] ~but it seems we can allocate a vector of matrices, see e.g. ~https://github.com/kaskr/adcomp/blob/b380fe8e78d16aba38d755ad4efbea3139d92212/tmb_syntax/report.cpp#L26
	- [x] ~this should allow us without too much logistics to pass the vector of Cholesky factors back to R via `REPORT`~
- [x] On the R side, refactor `h_residuals_pearson()` and required upstream code to use the list of Cholesky factors instead of calculating them again
- [x] make sure tests pass as before

PR #254, opened 2023-03-26
[WIP] implementing predict.mmrm_tmb()

kkmann:
See #189

We want to be able to use `predict(my_mmrm, my_newdata)` to get conditional means at new covariate combinations. Optionally, standard errors, confidence interval, or prediction interval should be available. Since we do not allow random effects, the `predict.lm()` interface is a good starting point.

- [x] ~~Add a short design doc to work out the approach~~ not necessary, use `predict.lm()` as guidance.
- [x] Add method for predicting conditional means without standard error or intervals
- [ ] fallback to `model.frame()` implmentation for mmrm - needs to be extended to allow dropping response
- [ ] Extend method to incorporate standard errors for the predicted conditional means & CI
- [ ] Add tests for conditional mean prediction
- [ ] Add tests for standard errors
- [ ] Add to NEWS

Currently out-of-scope: prediction interval. Requires within-subject covariance matrix of residual; how to propagate error (delta method?)


kkmann:
Some initial thoughts on challenges

1. A key problem is getting the X part of the model matrix in a consistent way with the initial model fit. At the moment, this is deeply buried in https://github.com/openpharma/mmrm/blob/c9dd43bd66b552d8b1abde41e53d17fc1f0749ef/R/tmb.R#L161-L179
2. The other important part is getting `beta_vcov` and `varcor` but they should be available from the fitted `mmrm_tmb` object, see  https://github.com/openpharma/mmrm/blob/c9dd43bd66b552d8b1abde41e53d17fc1f0749ef/R/component.R#L112-L128 (@danielinteractive: where is this aliasing mechanism explained?)

danielinteractive:
@kkmann thanks for the progress, great to see!! :-)

So the aliasing thing is only relevant for non-full rank design matrix. Not sure if we have this explained yet in detail anywhere. I think I put something into documentation for the corresponding argument `accept_singular` see https://openpharma.github.io/mmrm/main/reference/mmrm_control.html . Generally we follow here similar things as in `lm()`

kkmann:
There are a few preprocessing steps for the model matrix that don't make sense when building it for prediction. the main challenge seems to be making these optional to ensure we use the same functions to prepare the mm for fitting and prediciton. Another tricky bit is the input reordering - depending on how you look at it that could also be seen as a but in `fitted()` since it does not reorder to input ordering... For `predict()` we definitely need to respect input order.

kkmann:
Hm, the `varcor` part is not entirely clear to me yet - it also needs to work for the 'spatial' case.

What we need for predictive intervals (and `sample()`) is the conditioal vcov for each individual conditioning on observed values. @danielinteractive is there a function for getting the vcov for an individual?

kkmann:
pulled changes from pending #255

kkmann:
Potential to simplify further using the same approach as in emmeans. Would probably be good to have a consistent approach to computing model frames / matrices.

https://github.com/openpharma/mmrm/blob/c9dd43bd66b552d8b1abde41e53d17fc1f0749ef/R/interop-emmeans.R#L30-L31

danielinteractive:
> Hm, the `varcor` part is not entirely clear to me yet - it also needs to work for the 'spatial' case.
> 
> What we need for predictive intervals (and `sample()`) is the conditioal vcov for each individual conditioning on observed values. @danielinteractive is there a function for getting the vcov for an individual?

Yeah unfortunately at the moment we don't have a good way for that yet. I think that should be part of https://github.com/openpharma/mmrm/issues/232

kkmann:
> > Hm, the `varcor` part is not entirely clear to me yet - it also needs to work for the 'spatial' case.
> > What we need for predictive intervals (and `sample()`) is the conditioal vcov for each individual conditioning on observed values. @danielinteractive is there a function for getting the vcov for an individual?
> 
> Yeah unfortunately at the moment we don't have a good way for that yet. I think that should be part of #232

Hm, not so sure - to be fully generic this also needs to work for new timepoints and new individuals using temporal cov. To me this almost feels like implementing a covariance structure/function class that exposes the functionality in R. As long as the covariance structure is not really accessible from R the problem will persist. This would be a pretty major change though. Without access to the residual error covariance matrix we cannot do predictions for new timepoints of 'old' individuals.

danielinteractive:
@kkmann I see yeah I think you are right. We need to this also for new time points and new individuals ... The best would be if we can keep things consistent by only having one C++ level implementation that is then used both for the original model fitting but also for the post-hoc prediction/simulation part via R interface.

tests/testthat/test-tmb.R
```diff
@@ -711,6 +711,16 @@ test_that("h_mmrm_tmb_check_conv warns if convergence code signals non-convergen
...
 })
 
+# h_get_x_matrix -----
+
+test_that("h_get_x_matrix(object, newdata) works", {
+  formula <- FEV1 ~ SEX + ARMCD * AVISIT + us(AVISIT | USUBJID)
+  result <- mmrm(formula, fev_data[1:100, ], reml = FALSE)
+  newdata <- fev_data[101:nrow(fev_data), ]
+  mmrm:::h_get_x_matrix(result, newdata)
+  mmrm:::predict.mmrm_tmb(result, newdata)
```
danielinteractive, line 721: you don't need the `mmrm:::`, also internal functions are available in tests

vignettes/predict.Rmd
```diff
@@ -0,0 +1,109 @@
+---
+title: "Predicting from a MMRM"
```
danielinteractive, line 2: yes good idea to have a vignette for this! thanks!

vignettes/predict.Rmd
```diff
@@ -0,0 +1,109 @@
...
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+library(mmrm)
+```
+
+## Prediction of conditional mean
+
+Let $Y^{(i)}$ be the $n$ outcomes of the $i$-th subject and $X^{(i)}$ the $n\times p$
```
danielinteractive, line 28: sorry for being picky but it would be important to keep the same notation as in the other vignettes, you can take the model fitting algorithm one as the reference

danielinteractive: this will help the readers of the documentation, and also us once we decide to write a JSS paper sourcing material from the vignettes :-)

danielinteractive (commented):
thanks Kevin, some very initial comments for now, I will come back to the math tomorrow morning here, too late for my brain now

vignettes/predict.Rmd
```diff
@@ -0,0 +1,109 @@
...
+The special case of all outcomes being unobserved (new individual) is covered
+with $k=p$.
+
+Let further 
+$$
+\Sigma^{(i)}(\theta) = \begin{pmatrix} \Sigma^{(i)}_{new,new}(\theta) & \Sigma^{(i)}_{new,old}(\theta)\\ \Sigma^{(i)}_{old,new}(\theta) & \Sigma^{(i)}_{old,old}(\theta)\end{pmatrix}
+$$
+
+be a block decomposition where
+$\Sigma^{(i)}_{new,old}(\theta) = \big(\Sigma^{(i)}_{j,l}(\theta)\big)_{j = 1\dots k,\, l = 1\ldots k}$ and
```
danielinteractive, line 52:
```suggestion
$\Sigma^{(i)}_{new,new}(\theta) = \big(\Sigma^{(i)}_{j,l}(\theta)\big)_{j = 1\dots k,\, l = 1\ldots k}$ and similarly for the other blocks.
```

vignettes/predict.Rmd
```diff
@@ -0,0 +1,109 @@
...
+For implementing `predict()` $\mu$ is required. 
+
+For `predict(interval = "confidence")` additionally standard errors are required.
+These could be derived using the delta methods since $mu$ is a function of the
+estimated model parameters $\beta,\theta$.
+This would require the Jacobian of $mu(\beta,\theta)$.
+
+For `predict(interval = "prediction")` one would use the diagonal of $A$ instead,
+again the delta method could be used to make upper and lower boundaries reflect
+parameter estimation uncertainty.
```
danielinteractive, line 79: OK seems this is the hard part. So given beta and theta, we have the multivariate normal distribution. I wonder if it would be conceptually and even computationally simpler to draw beta, theta from the joint asymptotic normal distribution, centered around the estimates, and using the joint Hessian for the covariance. Then 2-step Monte Carlo sampling:
1) sample beta, theta, calculate from that mu and A.
2) sample the missing y's from the multivariate normal.
Then derive CI as sample quantiles. 
The sampling and computations are fast, can be done in C++, and we maybe need 1000 to have low MC error, so will be still good user experience.

danielinteractive: and when thinking about this further, instead of joint Hessian for beta, theta, could also just
a) draw theta from its asymptotic normal (we already have vcov for theta)
b) then draw beta from conditional asymptotic normal (we have that as well already)
we could prototype this in R already, without new code in C++.

vignettes/predict.Rmd
```diff
@@ -0,0 +1,109 @@
...
+    month = planned_visits
+  )
+tbl_data <- tbl_data[order(tbl_data$id, tbl_data$month), ]
+rownames(tbl_data) <- 1:nrow(tbl_data)
+tmp <- matrix(rnorm(n * length(planned_visits)), nrow = n) %*% chol(vcov) +
+  matrix(rep(3*planned_visits, n), nrow = n, byrow = TRUE)
+tbl_data$y <- as.vector(t(tmp)) 
+```
+
+```{r}
```
danielinteractive, line 101:
```suggestion
```{r, eval = FALSE}
```

R/tmb.R
```diff
@@ -427,6 +427,25 @@ h_mmrm_tmb_fit <- function(tmb_object,
...
+  for (i in seq_along(full_frame)) {
+    name <- names(full_frame)[i]
+    if (is.factor(full_frame[[i]]) & name %in% names(newdata)) {
+      xlevels[[name]] <- levels(full_frame[[i]])
+    }
+  }
+  # get terms object without response variable
+  fterms <- stats::delete.response(stats::terms(object$formula_parts$model_formula))
+  # first construct model frame, then the x_matrix
+  mframe <- model.frame(fterms, data = newdata)
```
clarkliming, line 444: `model.frame.mmrm_tmb` already implemented and supports new data now. I think we can use that (or update it as needed)

R/tmb-methods.R
```diff
@@ -40,6 +40,46 @@ fitted.mmrm_tmb <- function(object, ...) {
...
+    alpha <- 1 - level
+    # TODO: need to determine df, work with normal for now
+    if (interval == "confidence") {
+      res <- cbind(res,
+          lwr = res[, "fit"] - qnorm(1 - alpha/2) * se,
+          upr = res[, "fit"] + qnorm(1 - alpha/2) * se
+        )
+    }
+  }
+  if (ncol(res) == 1) { # return vector if only fit is computed
```
clarkliming, line 77: should we return early if only fit is needed?

vignettes/predict.Rmd
kkmann: hmhm, would it not be better to avoid sampling alltogether - I would prefer deterministic algos. We have the capability via TMB - it is just not exposed on the R side because all the cov stuff is hard coded in C++.

R/tmb.R
kkmann: @clarkliming could you squeeze that in here? I find the structure a bit confusing and don't want to make too many assumptions. We can keep the helper function for testing and if we reduce it to 2-3 lines with model.frame.mmrm_tmb we can remove it again.

tests/testthat/test-tmb.R
kkmann: helps with interactive testing, can remove later

vignettes/predict.Rmd
kkmann: hm, found that a bit clunky for this purpose - we can align those later once we know how to implement this - for now the vignette is the design doc / scratchbook

R/tmb-methods.R
kkmann: The performance overhead should be minimal (just a bunch of if/elses) and I like to avoid returning early unless there is an error etc.

vignettes/predict.Rmd
kkmann: updated, should be more aligned now

kkmann:
@danielinteractive @clarkliming Thanks for the comments. Maybe you could have another look at the updated specs - I like the idea of going with a parametric bootstrap sample of (beta, theta) first.
I would focus on predicting for a single individual first exposing the necessary functionality to R and then extending it to multiple individuals / handle in C++ directly once we have it functioning.

- [ ] expose an individual's covariance matrix to R via `get_cov(X_i, type, theta)` or similar.
- [ ] implement `bootstrap(mmrm_tmb, type = "raw")`  or similar for getting a sample from the joint asymptotic distribution of the estimation errors.
- [ ] code up predict() for single individual and test
- [ ] 'vectorize'
- [ ] merge this PR + start thinking about performance improvements by doing things directly in C++

The first two are a bit beyond me atm. Also happy to hand over this PR in case someone with more insights into the internals wants to pick it up for coding.

vignettes/predict.Rmd
clarkliming: we can expose `get_covariance(X_i, type, theta)` to R. however, it might also be good that this could be reported through the objective function? so that you can obtain the result directly.

clarkliming:
hi @kkmann for the reml and ml, I think the documentation is https://openpharma.github.io/mmrm/latest-tag/articles/algorithm.html#completing-the-square

kkmann:
> hi @kkmann for the reml and ml, I think the documentation is https://openpharma.github.io/mmrm/latest-tag/articles/algorithm.html#completing-the-square

... and it suggests that the you first solve for theta (objective function) and then for beta - but I did not see that in the code. Again, could be that I am overlooking the obvious here x) If we want the Fisher Information for the vector (beta, theta) via AD / TMB we would need an objective function over both (like ML) but I think ML is slightly biased in MMRM and usually ppl use REML, right?

vignettes/predict.Rmd
danielinteractive: Maybe through objective function we could get slower optimization? Maybe good to quickly do a option comparison between separate vs objective function interface

danielinteractive:
Hi @kkmann , @clarkliming and I just discussed this a bit more in the standup. We think that in order to close this PR soon, it would be good to first restrict this here to the conditional mean confidence intervals, i.e. without prediction intervals. This means just a "Xnew * beta_hat" calculation is needed with corresponding normal distribution based on the chosen cov matrix approximation for beta_hat.

Then in a different issue and PR we can look at prediction intervals. This could be either simpler (fixing variance parameters) or much more complicated (taking into account uncertainty about variance parameters), and will need additional methods investigation (papers, other packages etc)

vignettes/predict.Rmd
danielinteractive: we now created a new issue for this in #263

R/tmb.R
danielinteractive: see #262

danielinteractive: Hm ... actually @yonicd just mentioned to me rightfully that `model.matrix()` below won't include the response anyway... so we don't need to worry about dropping the response at all.

kkmann:
@clarkliming I was still waiting for a consitent way of obtaining the full vcov matrix from R (needs to be exposed from C++). Even if we drop the prediction/confidence intervals, we still need the vcov between old and new observatons per individual (see vignette). Would be very happy for you to step in here :)

danielinteractive:
I would say if @kkmann does not have time now then better @clarkliming continue this PR so that we can move forward to 0.3 release 👍🏼

danielinteractive:
Was implemented in #276

Issue #255, opened 2023-03-26
modularize `h_mmrm_tmb_data()`


kkmann:
@danielinteractive, to implement #254 in a clean way and to close #189 I feel we will need to expose the model matrix creation in a more flexible way. Currently, this is hidden away in `h_mmrm_tmb_data()`, see

https://github.com/openpharma/mmrm/blob/c9dd43bd66b552d8b1abde41e53d17fc1f0749ef/R/tmb.R#L161-L179

I suggest to refactor the construction of `full_frame` (also in `h_mmrm_tmb_data()`) and `x_matrix` into new internal functions
```
construct_full_frame(formula_parts, data, )
construct_x_matrix(formula_parts, data, full_frame, correct_singular)
```
that can be used as drop-in replacements in `h_mmrm_tmb_data()`. `construct_x_matrix()` would also have to support the case where the response variable is missing from data to enable consitent genration of the `x_matrix` between model fitting and prediction.

On that note - why is `reml` passed through `h_mmrm_tmb_data()` - does not seem like anything data related to me...

danielinteractive:
Thanks @kkmann ! Yeah sounds like a good idea. Let's just keep the `h_` prefix as long as this stays internal.
The `tmb_data()` function here means anything that is passed down to the TMB interface, not necessarily only `data` aka data frame stuff etc.

kkmann:
I feel it is easier to get the x_matrix post-hoc now, see https://github.com/openpharma/mmrm/pull/254/commits/ba8e1a25487a2a14798c163840a757c94e425aee. Since the model frame is saved we can revcover input factor levels from it.

PR #257, opened 2023-03-26
[WIP] refactor `h_mmrm_tmb_data()`

kkmann:
see #255 

- [x] make sure we can create the x_matrix for new data without the outcome variable for `predict()`
- [x] make sure we can restore the original order of the input data so that `fitted()` and `predict()` maintain input order in their returns
- [ ] make sure that factor levels are consistent between model matrix for old and new data


kkmann:
@danielinteractive bit early for a full review, but would be good to know whether this is going in the right direction.

kkmann:
I wonder whether this is the right way. Maybe `get_x_matrix(object, newdata)` is the better way to go.

kkmann:
see https://github.com/openpharma/mmrm/pull/254/commits/ba8e1a25487a2a14798c163840a757c94e425aee

PR #260, opened 2023-03-31
224 between within degree of freedom

danjamstat:
# Description

Draft vignette and initial code for between-within denominator degrees of freedom calculations.


danielinteractive:
Hi @chstock would you maybe have time to have a look at the vignette in this draft PR first? Thank you :-)

chstock:
> Hi @chstock would you maybe have time to have a look at the vignette in this draft PR first? Thank you :-)

Yes. I am happy to have a look, until Wednesday at the latest.

chstock (approved):
* Nice and concise - no major comments from my side.
* I think it make sense to consult Galecki (2013) or Pinheiro & Bates (2000) when trying to figure out how to best implement this; it appears less straightforward from the original publication.
* Just for consideration: Running a model that applied the between-within method:
    + We could actually run and report a model that uses the between-within method, i.e. `fit <- mmrm(..., mmrm_control(method="Between-within"))`. This may be a nice round off the vignette.
    + The between-within method was intended as a small sample adjustment (as was KR). It may be worth to point this out very briefly. Our example dataset is rather large. We could take only subset for illustration, especially in case the method would be contrasted to other small sample adjustments.

danielinteractive:
Hi @danjamstat , how is it going with this PR? would be great if we can progress this, let me know if we can help in any way.

danielinteractive:
Thanks @danjamstat for the update, I guess I proceed with the review from my side?

R/between_within.R
```diff
@@ -0,0 +1,105 @@
+#' Calculation of between-within Degrees of Freedom for One-Dimensional Contrast
```
danielinteractive, line 1:
```suggestion
#' Calculation of Between-Within Degrees of Freedom for One-Dimensional Contrast
```

R/between_within.R
```diff
@@ -0,0 +1,105 @@
...
+#'
```
danielinteractive, line 2: minor comment: can you change the file names to `between-within.R` and `between-within-classic.R` (only because all other files have the dash instead of the underscore)

R/between_within.R
```diff
@@ -0,0 +1,105 @@
...
+  list(
+    est = est,
+    se = se,
+    df = df,
+    t_stat = t_stat,
+    p_val = p_val
+  )
+}
+
+#' Calculation of between-within Degrees of Freedom for Multi-Dimensional Contrast
```
danielinteractive, line 54:
```suggestion
#' Calculation of Between-Within Degrees of Freedom for Multi-Dimensional Contrast
```

R/between_within.R
```diff
@@ -0,0 +1,105 @@
...
+  f_statistic <- as.numeric(1 / nrow(contrast) * h_quad_form_mat(contrast_est, prec_contrast))
+
+  n_subjects <- component(object, "n_subjects")
+  n_obs  <- component(object, "n_obs")
+  x_mat <- object$tmb_data$x_matrix
+  n_intercept <- if (any(colnames(x_mat) == "(Intercept)")) 1 else 0
+  x_mat_names <- colnames(x_mat)
+
+  bw_pars <- sapply(X = x_mat_names, function(x) {
+    if (x == "(Intercept)") {"within"}
```
danielinteractive, line 79: this function here seems duplicate with above, therefore please factor this out as separate helper, so that you can call e.g. just `bw_pars <- h_n_between_pars(object, x_mat, x_mat_names)` and avoid duplicate code

R/between_within_classic.R
```diff
@@ -0,0 +1,98 @@
+#' Calculation of between-within Degrees of Freedom for One-Dimensional Contrast
```
danielinteractive, line 1:
```suggestion
#' Calculation of Between-Within Degrees of Freedom for One-Dimensional Contrast
```

R/between_within_classic.R
```diff
@@ -0,0 +1,98 @@
...
+
+  n_subjects <- component(object, "n_subjects")
+  n_obs  <- component(object, "n_obs")
+
+  bw_pars <- apply(X = object$tmb_data$x_matrix, MARGIN = 2, FUN = function(x) {
+    n_unique <- nrow(unique(cbind(x, as.numeric(object$tmb_data$full_frame[[object$formula_parts$subject_var]]))))
+    if (n_unique > n_subjects) "within"
+    else "between"
+  })
+browser()
```
danielinteractive, line 29: reminder to remove all `browser()` statements

R/between_within_classic.R
```diff
@@ -0,0 +1,98 @@
+#' Calculation of between-within Degrees of Freedom for One-Dimensional Contrast
+#'
+#' @description Calculates the estimate, standard error, degrees of freedom,
+#' t statistic and p-value for one-dimensional contrast. Used in [df_1d()] if method is
+#' "Between-within".
```
danielinteractive, line 5: wait, I would expect this is used when the method is something like `Between-within-classic`?

R/between_within_classic.R
```diff
@@ -0,0 +1,98 @@
...
+#'
+#' @param object (`mmrm`)\cr the MMRM fit.
+#' @param contrast (`numeric`)\cr contrast vector. Note that this should not include
+#'   elements for singular coefficient estimates, i.e. only refer to the
+#'   actually estimated coefficients.
+#'
+#' @return List with `est`, `se`, `df`, `t_stat` and `p_val`.
+#' @keywords internal
+h_df_1d_bwc <- function(object, contrast) {
```
danielinteractive, line 14: the code looks very close to `h_df_1d_bw`. not sure what is the best path but can we avoid more the code duplication? e.g. how about giving another argument that switches the code between `standard` and `classic` behavior? (I guess classic is SAS similar?..)

tests/testthat/test-between_within.R
```diff
@@ -0,0 +1,80 @@
+# h_df_1d_bw ----
+
```
danielinteractive, line 2: these tests look good, thanks!
please also add integration type tests for the use in the other functions (i.e. upstream calls that ask for between within type ddf)

danielinteractive:
Hi @danjamstat , would you be able to implement the comments? :-) Thank you!

danielinteractive:
Checked with @danjamstat and we can take this over since unfortunately he does not have enough time anymore

R/between-within.R
```diff
@@ -0,0 +1,105 @@
...
+#'
+#' @param object (`mmrm`)\cr the MMRM fit.
+#' @param contrast (`numeric`)\cr contrast vector. Note that this should not include
+#'   elements for singular coefficient estimates, i.e. only refer to the
+#'   actually estimated coefficients.
+#'
+#' @return List with `est`, `se`, `df`, `t_stat` and `p_val`.
+#' @keywords internal
+h_df_1d_bw <- function(object, contrast) {
+  assert_class(object, "mmrm")
```
danielinteractive, line 15: lots of duplicate code here, between the two functions but also with other functions. Todo for Daniel: DRY

danielinteractive:
@danjamstat @chstock I was quite optimistic to finish this this week, but now hit a question I would appreciate your help: How are the between-within degrees of freedom defined for (arbitrary) linear combination of coefficients?

- Univariate: 
	- This would mean we e.g. have contrast vector `contrast = c(-1, 0, 2, 1)` and want to have the d.f. for that
	- This is not clear from current vignette or code
- Multivariate:
    - I could see from the code (see https://github.com/openpharma/mmrm/blob/224-between-within-degree-of-freedom/R/between-within.R#L103) that it was proposed to use the minimum of somehow the univariate d.f. somehow. 
    - But I am not sure this is what should be done?

Do the papers you cite in the vignette define the between-within degrees of freedom for these linear combinations? (If yes we would need to include those definitions also in the vignette)

Thanks for your help with this! I think we are getting closer to the finishing line.

danielinteractive:
- We can check again the original publication if there is anything in there for general contrast
- What about multiplying the design matrix with the contrast (vector/matrix) and then classifying the resulting vector/matrix accordingly? (just an idea)
- What is SAS doing for e.g. LS means / differences of factor levels for between/within degrees of freedom? (because that is the main application of contrast d.f.)

chstock:
> * What is SAS doing for e.g. LS means / differences of factor levels for between/within degrees of freedom? (because that is the main application of contrast d.f.)

SAS appears to apply a rather simple (though not necessarily very princpled) solution when specifying b/w-d.f. in the [CONTRAST ](https://documentation.sas.com/doc/en/pgmsascdc/9.4_3.4/statug/statug_mixed_syntax05.htm) or [ESTIMATE](https://documentation.sas.com/doc/en/pgmsascdc/9.4_3.4/statug/statug_mixed_syntax06.htm) statement in PROC MIXED. The documentation states that "...the default is the denominator degrees of freedom taken from the "Tests of Fixed Effects" table and corresponds to **the final effect you list in the CONTRAST/ESTIMATE statement**." This may reflect that there is no recommendation how to better handle this in the literature and would be straightforward to implement.

[LSMEANS](https://documentation.sas.com/doc/en/pgmsascdc/9.4_3.4/statug/statug_mixed_syntax08.htm#statug.mixed.lsmeansstmt_adjdfe) does not seem to allow b/w d.f. according to the documentation. Use in the [MODEL](https://documentation.sas.com/doc/en/pgmsascdc/9.4_3.4/statug/statug_mixed_syntax10.htm#statug.mixed.modelstmt_ddf) statement, i.e. for single coefficients, looks straightforward, as implemented.

(From my practicial experience, I would assume that most often the test results are drawn from a test of a single model coefficient (a time-by-treatment interaction) and the d.f. is specified in the MODEL statement.)

chstock:
> * We can check again the original publication if there is anything in there for general contrast

The original paper (by Schluchter and Elashoff, 1990) considers several related designs, none of which is an MMRM as we use it today. This makes it partially hard to translate. I understand that they distinguish between the "between-subject" and "within-subject" factors, as well as interactions. I found no hint that they propose how to deal with more custom contrasts, i.e. anything beyond tests of coefficients of fixed effects.

danielinteractive:
@chstock we have quite a few differences vs. SAS already so I am keeping for unstructured the same logic as for the others now (but the vignette highlights this and the other differences to SAS). I think in the future we could have another "Between-Within-Modified" method that matches SAS

danielinteractive (approved):

PR #265, opened 2023-04-21
add cholesky cache

### LINKED ISSUE #172: Shall we cache the `lower_chol` for subjects so that it will not be re-calculated?
currently, we derive the "lower_chol" for full visits, and for each subject that do not have full visits, we calculate their Cholesky based on the subset of the covariance matrix, and then we take the Cholesky decomposition again.

If there are subjects with the same missed visits (e.g., full visits is week 1, 2, 3 and 4, and some subjects only missed week 3, so their cholesky will computed multiple times (even though they are the same).

I wonder if we should use some kind of map to avoid repeated calculations?

### LINKED ISSUE #264: Save the Cholesky factors from
part of #232

clarkliming:
close #172


danielinteractive (commented):
Thanks Liming for trying out the caching, so the gain seems rather small in terms of computation time needed, correct? If that is so, I would tend to prioritize the code simplicity over the marginal time gains.

(I think you saw that https://github.com/openpharma/mmrm/issues/232 is a bit different, it would not be about caching but just reporting back the individual Cholesky factors)

clarkliming:
> Thanks Liming for trying out the caching, so the gain seems rather small in terms of computation time needed, correct? If that is so, I would tend to prioritize the code simplicity over the marginal time gains.
> 
> (I think you saw that #232 is a bit different, it would not be about caching but just reporting back the individual Cholesky factors)

ah yes this actually is for #172 (a bit different from #232 but still similar). But I do think that the current form is cleaner and moduelized ( no need to have two layers of if else check on spatial or full_visits)

src/chol_cache.h
```diff
@@ -0,0 +1,69 @@
...
+  Type rho;
+  Type logrho;
+  lower_chol_sp_exp() {
+    // This default constructor is needed because the use of `[]` in map.
+  }
+  // Constructor from theta. For now the cholesky does not need to be cached.
+  lower_chol_sp_exp(vector<Type> theta): const_sd(exp(theta(0))), rho(invlogit(theta(1))) {
+    this->logrho = log(this->rho);
+  }
+  matrix<Type> get_chol(std::vector<int> visits, matrix<Type> dist) {
```
danielinteractive, line 61: so once we would have this class and method, then we could remove some other code for spatial exponential to avoid duplication, correct?

src/derivatives.h
```diff
@@ -100,9 +100,7 @@ struct derivatives_nonspatial: public derivatives_base<Type> {
   // Constructor from theta, n_visits and cov_type, and cache full_visits values.
   derivatives_nonspatial(vector<Type> theta, int n_visits, std::string cov_type): cov_type(cov_type), n_visits(n_visits), full_visit(std::vector<int>(n_visits)) {
     this->theta = theta;
-    for (int i = 0; i < n_visits; i++) {
-      this->full_visit[i] = i;
-    }
+    std::iota(std::begin(this->full_visit), std::end(this->full_visit), 0);
```
danielinteractive, line 103: nice! I didn't know this before

src/mmrm.cpp
```diff
@@ -156,6 +147,21 @@ Type objective_function<Type>::operator() ()
...
+  if (is_spatial) {
+    dim = 2;
+  } else {
+    dim = n_visits;
+  }
+  matrix<Type> covariance_lower_chol = matrix<Type>::Zero(dim * n_groups, dim);
+  for (int r = 0; r < n_groups; r++) {
+    covariance_lower_chol.block(r * dim, 0, dim, dim) = chols_by_group[r]->get_chol(visit, dist);
+    delete chols_by_group[r];
+  }
```
danielinteractive, line 164: this code looks a bit low-level in this context, any chance to move this to a function?

src/mmrm.cpp
```diff
@@ -60,42 +60,33 @@ Type objective_function<Type>::operator() ()
...
-  } else {
-    dim_cov_mat = n_visits;
+  std::map<int, lower_chol_base<Type>*> chols_by_group;
+  for (int r = 0; r < n_groups; r++) {
+    // in loops using new keyword is required so that the objects stays on the heap
+    // otherwise this will be destroyed and you will get unexpected result
+    if (is_spatial) {
+      chols_by_group[r] = new lower_chol_sp_exp<Type>(theta.segment(r * theta_one_group_size, theta_one_group_size));
+    } else {
+      chols_by_group[r] = new lower_chol_nonspatial<Type>(theta.segment(r * theta_one_group_size, theta_one_group_size), n_visits, cov_type);
```
danielinteractive, line 70: so here we still need to differentiate between spatial and non-spatial, any chance to also unify this at this code level?

danielinteractive: if too complicated, not needed, or at least not in this PR

src/chol_cache.h
```diff
@@ -0,0 +1,69 @@
...
+  }
+  // Constructor from theta, n_visits and cov_type, and cache full_visits values.
+  lower_chol_nonspatial(vector<Type> theta, int n_visits, std::string cov_type): cov_type(cov_type), n_visits(n_visits), full_visit(std::vector<int>(n_visits)) {
+    this->theta = theta;
+    std::iota(std::begin(this->full_visit), std::end(this->full_visit), 0);
+    this->n_theta = theta.size();
+    this->chol_full = get_cov_lower_chol_grouped(this->theta, this->n_visits, this->cov_type, 1, false);
+    this->chols[full_visit]  = this->chol_full;
+  }
+  matrix<Type> get_chol(std::vector<int> visits, matrix<Type> dist) {
```
danielinteractive, line 32: I like that this is here, nice

danielinteractive (commented):
Thanks Liming, I had a more detailed look now, nice work! I agree that this simplifies the `mmrm.cpp` a bit so is a good direction. I adapted the PR description. Please see comments inside

src/mmrm.cpp
clarkliming: yes this part will be refined later. here we are trying to create "full_chol" and return this. However later we should return all subject (and actually, we only save the list of valid visits, not replication of all subjects, and we can easily obtain the chol for each subject depending on the visits)

src/mmrm.cpp
clarkliming: actually, they are somehow unified. they are both class lower_chol_base, and have the same method "get_chol". So here they are kind of unified. 

If we would unify further, then we probably will make the "lower_chol_nonspatial" and "lower_chol_sp_exp" one class (then we would store the type inside, and every time depending on the covariance type choose suitable method). I did not do this at first because, I doubt for spatial covariance anything should be cached (this cache only works for the categorical time points)

src/chol_cache.h
clarkliming: the function `get_cov_lower_chol_grouped` should be updated too as we are separating the calss in groups already (each group as a "lower_chol_base" class. The function `get_spatial_covariance_lower_chol` is becoming useless. should call `get_spatial_exponential` directly (this is the function to calculate the cholesky)

clarkliming:
this PR potentially helps #232 because while we return the cholesky for each subject, we don't really need to return all cholesky: only those unique ones (with unique vists)

src/chol_cache.h
```diff
@@ -0,0 +1,79 @@
...
+  virtual matrix<Type> get_chol(std::vector<int> visits, matrix<Type> dist) = 0;
+};
+// Struct to obtain Cholesky for non-spatial.
+template <class Type>
+struct lower_chol_nonspatial: public lower_chol_base<Type> {
+  std::map<std::vector<int>, matrix<Type>> chols;
+  std::string cov_type;
+  int n_visits;
+  std::vector<int> full_visit;
+  int n_theta; // theta.size()
```
danielinteractive, line 18: please remove any commented code, e.g. here:
```suggestion
  int n_theta;
```

src/chol_cache.h
```diff
@@ -0,0 +1,79 @@
...
+  }
+  // Constructor from theta. For now the cholesky does not need to be cached.
+  lower_chol_spatial(vector<Type> theta, std::string cov_type): theta(theta), cov_type(cov_type) {
+  }
+  matrix<Type> get_chol(std::vector<int> visits, matrix<Type> dist) {
+    return get_spatial_covariance_lower_chol(this->theta, dist, this->cov_type);
+  }
+};
+
+template <class Type>
```
danielinteractive, line 64: please add some comments on top of this function what it does, in particular highlight the modification of the argument `chols`

clarkliming:
by the way, the test is extremely slow (taking 30 minutes now; compiling the package is taking a lot of time)

I check the actions and there are two compilations. I wonder if we could, instead of build two times, we just compile once (one is for junit tests, the other is for R cmd check, between these checks the src is cleaned)

clarkliming:
another thing: I think we can use `unordered_map` here to replace `map`. any opinions?

danielinteractive (approved):
Thanks @clarkliming !
Re: speed - yeah we have seen that before that it is slow. I don't think we can do much for now, we would have to change the whole CI/CD set up I guess (@cicdguy might have ideas?). Good to merge if you are sure we have all unit tests in place

danielinteractive:
Ok thanks @cicdguy , yeah we don't want to pay for more speed here. Any chance to avoid the 2nd compilation which Liming described?

PR #267, opened 2023-05-08
update residual functions

### LINKED ISSUE #232: Avoid the need to recalculate any Cholesky factors for normalized residuals

clarkliming:
close #232 

key changes:

1. reuse `epsilonTilde` (directly report this, this is the normalized residual)
2. refactor the pearson calculation code
3. create `h_residual_rsp` to be consistent

### DIFF
--- a/R/tmb-methods.R
+++ b/R/tmb-methods.R
@@ -244,50 +244,42 @@ print.mmrm_tmb <- function(x,
 #' - \insertRef{galecki2013linear}{mmrm}
 residuals.mmrm_tmb <- function(object, type = c("response", "pearson", "normalized"), ...) {
   type <- match.arg(type)
-  resids_unscaled <- component(object, "y_vector") - unname(fitted(object))
-  if (type == "response") {
-    resids_unscaled
-  } else {
-    if (object$formula_parts$is_spatial) {
-      stop("Only 'response' residuals are available for models with spatial covariance structures.")
-    }
-    if (type == "pearson") {
-      h_residuals_pearson(object, resids_unscaled)
-    } else if (type == "normalized") {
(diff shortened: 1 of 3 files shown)


danielinteractive (commented):
Thanks a lot @clarkliming , really nice!
Can you also adjust the NEWS entry for `residuals` because now also "normalized" works for spatial covariance structures?

danielinteractive:
By the way, I was just thinking, could we maybe compute the Pearson residuals very easily as a byproduct in TMB and report them too? Then it would simplify the R code and enable Pearson residuals also for spatial covariance structures.

danielinteractive (approved):

PR #268, opened 2023-05-09
Impact of Missing Data in MMRM Software

PhilBoileau:
We assess the impact of missing data on various MMRMs implementations' ability to fit models. The MMRM implementations considered are the `mmrm`, `nlme` and `glmmTMB` R packages, as well as SAS's MIXED procedure. The data-generating processes used in this simulation study are inspired by those seen in longitudinal ophthalmology studies. More details are provided in `simulations/missing-data-benchmarks/README.md`.

A quick summary of the results: When considering unstructured covariance matrices, this simulation study demonstrates that `mmrm` is more robust to large amounts of missing data than any other implementation considered. `mmrm` is also the fastest implementation regardless of covariance matrix structured used.


danielinteractive (commented):
Thanks a lot @PhilBoileau for the great work on these simulations! Much appreciated. I think it looks great, and I will not need to check all details here, as it is not part of the R package. I wonder only if it could be worthwhile to take a condensed summary paragraph and add that to the comparison vignette we have in the package?

PhilBoileau:
That was a good idea, @danielinteractive! I added a very brief summary to the method comparison vignette.

danielinteractive (approved):

PR #269, opened 2023-05-16
report choleskys in tmb

### LINKED ISSUE #263: Add `get_cov_chols(<tbd>, type, theta)` R function to expose C++ cov Cholesky factor
- for all individuals with multiple time points (incl. unobserved ones in data set before) at once
- return list of Cholesky factors
	- needs to be appropriately weighted already

First idea was this:
- idea see https://github.com/openpharma/mmrm/blob/main/src/mmrm.cpp#L65 and following until line 93 - this could be factored out in helper function (then also need tests again for that)
- additional inputs needed:
	- `n_groups`
	- `n_visits`
	- `n_subjects`
	- `subject_zero_inds`
	- `visits_zero_inds`
	- `subject_n_visits`
	- `coordinates`
	- `weights`
	- this could be put in `tmb_data` and maybe just keep x and y empty or something

Second idea is:
- just do it on the R side, see https://github.com/openpharma/mmrm/blob/main/R/tmb.R#L471 as start
- use the reported Cholesky factors
- of course don't do any optimization

clarkliming:
close #263 

minimum changes to the current code to unblock other functionalities.


R/get_chol.R
```diff
@@ -0,0 +1,59 @@
+#' Obtain Cholesky for new data and theta.
+#'
+#' @description Use a fitted model to obtain (weighted) Cholesky factors
```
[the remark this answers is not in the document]
clarkliming: since it is internal function now, no need for the lifecycle badge now

clarkliming:
thank you @danielinteractive  for the comment. this PR was not finalized yet yesterday due to limited time, the PR is created just to say that I am working on it and has some progress to enable github action checks. Now it should be fine and ready for review.

For avoiding duplicated code, I think the majority of the code can not be reduced.  A lot of code is to check the consistency between the new data and old data. Only thing duplicated is the `TMB::MakeADFun` but this is super short.

R/get_chol.R
```diff
@@ -0,0 +1,58 @@
...
+#' with new data, weights and theta.
+#'
+#' @param fit (`mmrm_fit`)\cr produced by [mmrm()].
+#' @param data (`data.frame`)\cr input data frame.
+#' @param weights (`numeric`)\cr vector of weights.
+#' @param theta (`numeric`)\cr new theta estimates.
+#' @param complete_case (`flag`)\cr whether to use complete input `data`.
```
clarkliming, line 10: because we only need visit and subject id to obtain the subject Cholesky factors, so whether there is missing values in Y does not matter now. This option is added to ensure the consistency (with no default value, it give the same output as fit$tmb_object$report)

R/get_chol.R
```diff
@@ -0,0 +1,58 @@
...
+    data <- data[complete.cases(data), ]
+  }
+  assert_numeric(weights, len = nrow(data), null.ok = TRUE)
+  weights <- h_default_value(weights, rep(1, nrow(data)))
+  assert_numeric(theta, len = length(fit$theta_est), null.ok = TRUE)
+  theta <- h_default_value(theta, fit$theta_est)
+  visit_vars <- fit$formula_parts$visit_var
+  group_var <- fit$formula_parts$group_var
+  subject_var <- fit$formula_parts$subject_var
+  assert_names(names(data), must.include = c(visit_vars, subject_var, group_var))
```
clarkliming, line 30: these variables are required

R/get_chol.R
```diff
@@ -0,0 +1,58 @@
...
+    vn <- paste(vname(data), fit$formula_parts$group_var, sep = "$")
+    data[[group_var]] <- h_factor_ref(data[[group_var]], fit$tmb_data$full_frame[[group_var]], vn)
+  }
+  tmb_data <- h_mmrm_tmb_data(
+    h_mmrm_tmb_formula_parts(as.formula("rep(1, nrow(data))~1"), as.cov_struct(fit$formula_parts$formula)),
+    data = data, weights, identical(fit$tmb_data$reml, 1L),
+    accept_singular = TRUE, drop_visit_levels = FALSE
+  )
+  tmb_data$n_groups <- fit$tmb_data$n_groups
+  tmb_data$subject_groups <- h_factor_ref(tmb_data$subject_groups, fit$tmb_data$full_frame[[group_var]])
```
clarkliming, line 48: h_mmrm_tmb_data drops levels for group var, leading to incorrect number of groups (and the group levels)

R/get_chol.R
```diff
@@ -0,0 +1,58 @@
...
+  tmb_parameters <- h_mmrm_tmb_parameters(fit$formula_parts, tmb_data, start = theta, n_groups = tmb_data$n_groups)
+  tmb_object <- TMB::MakeADFun(
+    data = tmb_data,
+    parameters = tmb_parameters,
+    hessian = TRUE,
+    DLL = "mmrm",
+    silent = TRUE
+  )
+  tmb_object$report()$cholesky_all
```
clarkliming, line 57: potentially, we could return everything, not only the choleksy factors

danielinteractive: hm I would keep for now as is

R/utils.R
```diff
@@ -343,3 +343,20 @@ h_default_value <- function(x, y) {
     x
   }
 }
+
+#' Convert Character to Factor Following Reference
+#'
+#' @param x (`character`)\cr or (`factor`) input.
```
danielinteractive, line 349:
```suggestion
#' @param x (`character` or `factor`)\cr  input.
```

R/utils.R
```diff
@@ -343,3 +343,20 @@ h_default_value <- function(x, y) {
...
+#' @param ref (`factor`)\cr reference.
+#' @param var_name (`string`)\cr variable name of input `x`.
+#'
+#' @details Use `ref` to convert x into factor with the same level.
```
danielinteractive, line 353:
```suggestion
#' @details Use `ref` to convert `x` into factor with the same levels.
```

R/utils.R
```diff
@@ -343,3 +343,20 @@ h_default_value <- function(x, y) {
...
+#'
+#' @keywords internal
+h_factor_ref <- function(x, ref, var_name = vname(x)) {
+  assert_factor(ref)
```
danielinteractive, line 357: also assert `x` type (as documented)

tests/testthat/test-get_chol.R
```diff
@@ -0,0 +1,103 @@
...
+
+test_that("h_get_chol errors on non correct data", {
+  fit <- get_mmrm()
+  expect_error(
+    h_get_chol(fit, data = iris),
+    "Names must include the elements"
+  )
+})
+
+test_that("h_get_chol works with visit var as character subset of original fit", {
```
danielinteractive, line 46: can you add a test where we have additional subjects and also additional visits?
so e.g. original data only has subjects 1, 2, now we have 3 as well.
Also original data has only visits 1, 4, 5, for subject 1, now we look at visits 2, 3, 4, 5, 6.

danielinteractive (commented):
Thanks Liming! 
Yeah no worries, I just reviewed already yesterday because you requested review and it was not draft PR :-) 
Regarding the file location, how about moving this to `tmb.R`?

tests/testthat/test-get_chol.R
clarkliming: this is already covered in https://github.com/openpharma/mmrm/blob/15c7dae2e45946266a9fba5d66b8ccedeee6e8fe/tests/testthat/test-get_chol.R#L73 where the subject id are totally new (not part of the orginal PT1 things)

and also https://github.com/openpharma/mmrm/blob/15c7dae2e45946266a9fba5d66b8ccedeee6e8fe/tests/testthat/test-get_chol.R#L22 the visits provided is more than previously we have (all subjects has 4 visits)

and also https://github.com/openpharma/mmrm/blob/15c7dae2e45946266a9fba5d66b8ccedeee6e8fe/tests/testthat/test-get_chol.R#L65

additional visits will cause error

danielinteractive: hm ok I did not see that clearly from description. can you please add additional test with corresponding clear label? thanks

tests/testthat/test-utils.R
```diff
@@ -169,3 +169,21 @@ test_that("h_default_value works", {
   expect_identical(h_default_value(x, "test"), x)
   expect_identical(h_default_value(NULL, x), x)
 })
+
+# h_h_factor_ref ----
+
+test_that("h_factor_ref works", {
+  ref <- factor(c("a", "b", "c"), levels = c("a", "b", "c"))
```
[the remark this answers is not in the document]
clarkliming: my bad. I should use `c` `b` `a` as levels to make a non default factor

clarkliming:
> 

very interestingly that the trigger of review seems to be automatic (even if you do not include anyone as reviewer)

R/tmb.R
```diff
@@ -532,3 +532,69 @@ fit_mmrm <- function(formula,
   fit$call$formula <- formula_parts$formula
   fit
 }
+
+#' Obtain Cholesky for new data and theta.
```
danielinteractive, line 536:
```suggestion
#' Cholesky Factors for New Data and Variance Parameters
```

R/tmb.R
```diff
@@ -532,3 +532,69 @@ fit_mmrm <- function(formula,
...
+#'
+#' @description Use a fitted model to obtain (weighted) Cholesky factors
+#' with new data, weights and theta.
```
danielinteractive, line 539:
```suggestion
#' with new data, weights and `theta`.
```

R/tmb.R
```diff
@@ -532,3 +532,69 @@ fit_mmrm <- function(formula,
...
+#'
+#' @param fit (`mmrm_fit`)\cr produced by [mmrm()].
+#' @param data (`data.frame`)\cr input data frame.
+#' @param weights (`numeric`)\cr vector of weights.
+#' @param theta (`numeric`)\cr new theta estimates.
```
danielinteractive, line 544:
```suggestion
#' @param theta (`numeric`)\cr new variance parameter estimates.
```

R/tmb.R
```diff
@@ -532,3 +532,69 @@ fit_mmrm <- function(formula,
...
+#' @param complete_case (`flag`)\cr whether to use complete input `data`.
+#'
+#' @details
+#' In `mmrm` call, missing values (in response or covariates) will be removed prior to model fitting.
+#' However, if we only require the Cholesky factors for each subject, it is
+#' not necessary to remove NA values. The flag, `complete_case`, is used to decide whether the
+#' input data should follow a similar approach of removing NA values rows.
```
danielinteractive, line 551:
```suggestion
#' not necessary to remove `NA` values. The flag, `complete_case`, is used to decide whether the
#' input data should follow a similar approach of removing `NA` values rows.
```

R/tmb.R
```diff
@@ -532,3 +532,69 @@ fit_mmrm <- function(formula,
...
+#' @return `list` of Cholesky factors.
```
danielinteractive, line 552:
```suggestion
#' @return  The `list` of Cholesky factors for all individuals and time points in the new `data`.
```

R/tmb.R
```diff
@@ -532,3 +532,69 @@ fit_mmrm <- function(formula,
...
+  } else {
+    vn <- paste(vname(data), visit_vars, sep = "$")
+    data[[visit_vars]] <- h_factor_ref(data[[visit_vars]], fit$tmb_data$full_frame[[visit_vars]], vn)
+  }
+  if (!is.null(group_var)) {
+    vn <- paste(vname(data), fit$formula_parts$group_var, sep = "$")
+    data[[group_var]] <- h_factor_ref(data[[group_var]], fit$tmb_data$full_frame[[group_var]], vn)
+  }
+  # h_mmrm_tmb_data drops levels for group var, leading to incorrect number of groups (and the group levels)
+  # so here we add the levels back
```
danielinteractive, line 583:
```suggestion
  # `h_mmrm_tmb_data` drops levels for the group variable, leading to incorrect number of groups (and the group levels).
  # Therefore, here we add the levels back.
```

tests/testthat/test-tmb.R
```diff
@@ -1770,3 +1770,107 @@ test_that("fit_mmrm throws informative error when covariance structure is not
...
+    h_get_chol(fit)
+  )
+  new_theta <- seq_len(length(fit$theta_est)) / length(fit$theta_est)
+  expect_identical(
+    fit$tmb_object$report(new_theta)$cholesky_all,
+    h_get_chol(fit, theta = new_theta, data = fev_data, weights = NULL)
+  )
+})
+
+test_that("h_get_chol works correctly to provide cholesky on all visits(subjects have full visits)", {
```
danielinteractive, line 1789:
```suggestion
test_that("h_get_chol works correctly to provide cholesky on all visits (subjects have full visits)", {
```

danielinteractive (approved):
Thanks Liming, looks nice! Only few minor last suggestions, afterwards good to merge

danielinteractive (approved):

Issue #274, opened 2023-05-23
difference in simulation settings


clarkliming:
In the simulation folder, I found that for `proc mixed` we are using KR
https://github.com/openpharma/mmrm/blob/5a61959717d50394108c346c51b49967967083c4/simulations/missing-data-benchmarks/R/method/proc-mixed-wrapper.R#LL62C1-L63C1

while for `mmrm` we are using asymptotic

https://github.com/openpharma/mmrm/blob/5a61959717d50394108c346c51b49967967083c4/simulations/missing-data-benchmarks/R/method/mmrm-wrapper.R#L48

is this something we should make consistent @PhilBoileau

PR #275, opened 2023-05-23
DDFM option set to SAT instead of KR in PROC MIXED

PhilBoileau:
Addresses #274: The `PROC MIXED` wrapper function used in the simulation study assessing the impact of missing data erroneously set the degrees of freedom calculation to Kenward-Rogers. It is now set to Satterthwaite, the `mmrm()` default.


danielinteractive (approved):
Thanks @PhilBoileau ! Do we also need to rerun simulations now for the SAS part?

PhilBoileau:
Sorry for the delay, @danielinteractive, I'm at a conference. I think it would be worth re-running a small portion of the simulation study to assess if the results are any different. My guess is that they're negligible. I can't rerun them myself though because I no longer have access to SAS Studio. Perhaps @juliadedic1 would be able to do this?

danielinteractive:
Thanks @PhilBoileau , cool, no worries, it is not urgent - I will chat with Julia

PhilBoileau:
Sounds good, thanks!

PR #276, opened 2023-05-24
189 predict lm design

clarkliming:
add design prototype

major changes

1. model.frame is updated. `exclude_vars` argument is replaced by `include_vars` because, if we have visit var in model, like `FEV1 ~ AVISIT + ar1(AVISIT|USUBJID)`, if we exclude "AVISIT" then the model will have nothing in the righthand side. To be robust, we have to use add the grouping/visit/subject var to model
2. `model.frame` is used to create the frame with the same levels used in the fit.
3. allow `na_action` in `model.frame`, because we allow NA values in response.
4. `h_mmrm_tmb_data` have argument `allow_na_response`. the reason not to use na.action is that NA are never allowed for the right-hand side, but allowed in left-hand side
5. sampling approach is slow. we have experience in Kenward-Roger to obtain derivatices of the Cholesky, it is possible to have the variance of $cov(\beta)$ with the variance of $\theta$ using sandwich estimator (should be much faster than sampling)

c++ side

1. chol_cache struct can get sigma and sigma inverse now.
2. remove utility.cpp but use template function `as_vector` and `as_matrix` (to convert between Eigen matrix and Rcpp NumericMatrix)
3. add `segment` function of subset from a vector. it is needed because y can have NA values, using select matrix will fail. We can switch to segment to replace `select matrix` calculation (eigen 3.4 supports that, but RcppEigen has not support 3.4 yet)


clarkliming:
Sampling approach:
1. Draw a sample from the asymptotic normal theta distribution, using the theta estimate and the covariance matrix
1. Put the theta into the objective function to get the beta estimate and covariance matrix
1. Draw a sample from the asymptotic normal beta distribution
1. This yields a joint sample from (theta, beta)
    1. Note that this is not a normal distribution, because theta enters in a complex non-linear way into the conditional distribution for beta via the covariance matrix.


sampling from beta seems a bit unnecessary here? @danielinteractive

danielinteractive:
@clarkliming We also need to sample `beta`, as that goes into the means of the outcomes `y` and hence its uncertainty needs to be factored in.

clarkliming:
> @clarkliming We also need to sample `beta`, as that goes into the means of the outcomes `y` and hence its uncertainty needs to be factored in.

its uncertainty are factored in in two parts: 
image
the first part is the variance of expectation conditional on $\theta$ so it should be the $\hat\beta$ ($beta$ follows normal distribution and the expectation is $X\hat\beta$, does not need to sample on $\beta$)

the second part is the expectation of variance conditional on $\theta$ and the uncertainty are included in $X cov(\hat\beta) X^\top$ and $\Sigma$

danielinteractive:
@clarkliming That would maybe also work. As long as we factor in `beta` uncertainty, we are good. We can do that via Monte Carlo or via the asymptotic normal distribution conditional on `theta`.

clarkliming:
hi @danielinteractive  this is a draft version of confidence intervals. There are some key changes:

1. model.frame is updated. `exclude_vars` argument is replaced by `include_vars` because, if we have visit var in model, like `FEV1 ~ AVISIT + ar1(AVISIT|USUBJID)`, if we exclude "AVISIT" then the model will have nothing in the righthand side. To be robust, we have to use add the grouping/visit/subject var to model
2.  `h_get_one_sample` will return the expectation of y given theta, and the variance of y given theta. with this, we are able to use the conditional variance formula to obtain the unconditional variance
3. `h_get_prediction_variance` will simulate on theta and aggregate results from `h_get_one_sample`
    4. idea: since, we already have the derivatives of `covariance` on $\theta$ (it is calculated for Kenward-Roger, it is possible to use delta method to obtain the variance of `y`? the covariance of $\theta$ is clear and we can obtain at least first order approximation ( as a prototype, we can use numDeriv  to obtain the jacobians)
5. apparently, `h_get_chol` is not used; it seems unnecessary to have this function.

could you please have a brief look and provide your major comments? (I have not finalized the details so no worries on the detailed implementation, but the general design matters)

design/predict_interval/predict_interval.Rmd
```diff
@@ -0,0 +1,76 @@
...
+
+## Sampling approach
+
+To obtain $Var(y_{new})$, we need to integrate over $\theta$ to obtain the expected variance and vairance of expectation.
+
+However, if we use simulation based approach, we can:
+
+1. Sample $\theta$
+2. Obtain $Var(y_{new}|\theta)$ and $E(y_{new}|\theta)$
+3. Summarize over $\theta$
```
danielinteractive, line 74: how does this work exactly? I am not sure that is possible without additional assumptions. like if we have 10 samples of theta, and for each the variance of y, how do we want to aggregate this?
Usually, a pure sampling based approach is easier at this step.

danielinteractive (commented):
Thanks Liming, I think the direction is good.
Can you address my questions inside?
For the other small things:
- `model.frame` - I guess `exclude_vars` is not a standard expectation to have, so it is ok to change
- yeah very good to remove any dead code like `h_get_chol`

clarkliming:
[20230525102735-0001.pdf](https://github.com/openpharma/mmrm/files/11560577/20230525102735-0001.pdf)

@danielinteractive this is a basic implementation of aggregation over $theta$

danielinteractive:
Thanks @clarkliming , looks good - please proceed in this direction

clarkliming:
looking at the vignettes, we are also interested in prediction of "new y" given part of the observed old y values. So we will require "old y" exists. Is this correct that we also want this?

danielinteractive:
Yes, that is an important feature, imagine e.g. a new patient comes along in the new data set, or new visits for existing patients.

clarkliming:
> Yes, that is an important feature, imagine e.g. a new patient comes along in the new data set, or new visits for existing patients.

then we could have very different design. The thing is, we need to provide full `y` vector, including both NA and non-NA values. NA values are considered as missing values to be predicted; and in addition, there will be lots of "matrix inverse" to obtain the conditional mean and conditional variance for subset of matrix.

current implementation does not fulfill that requirement

danielinteractive:
@clarkliming Yes please see the description of the formulas that @kkmann put in for the conditional means of new visits given existing patients.

clarkliming:
the implementation of conditional mean/variance on $\theta$ are now added. haven't add any unit tests or R interface yet

danielinteractive:
OK thanks @clarkliming then maybe I wait until you have the R side and unit tests before review

clarkliming:
enable actions now

NEWS.md
```diff
@@ -17,7 +17,7 @@
 - Add confirmation if the number of visit levels is too large. Use `options(mmrm.max_visits = )`
   to specify the maximum number of visits allowed in non-interactive mode, or asked in interactive
   sessions.
-
+- Add `predict` method to obtain conditional mean estimate with or without observations.
```
danielinteractive, line 20:
```suggestion
- Add `predict` method to obtain conditional mean estimates and prediction intervals, both with or without previous observations.
```

NEWS.md
```diff
@@ -27,8 +27,8 @@
 - Previously `model.frame` call on `mmrm` object with transformed terms, or new
   given data, e.g. `model.frame(mmrm(Y ~ log(X) + ar1(VISIT|ID), data = <new data>)`,
   will cause errors. This is now fixed and `model.frame` returns the data correctly.
-  Now `na.action` argument always use "na.omit", `subset` argument is not used.
-  `full` argument is deprecated and use `exclude` instead.
+  Now `na.action` argument can be "na.omit" or "na.pass", `subset` argument is not used.
+  `full` argument is deprecated and use `include` instead.
```
danielinteractive, line 31:
```suggestion
  The `full` argument is deprecated and the `include` can be used instead.
```

R/tmb-methods.R
```diff
@@ -40,43 +40,146 @@ fitted.mmrm_tmb <- function(object, ...) {
...
 
+#' @describeIn mmrm_tmb_methods predict conditional means for new data;
+#'  optionally with standard errors and confidence or prediction intervals.
+#'  Returns a vector of predictions if `se.fit == FALSE` and
+#'  `interval == "none"`; otherwise it returns a data.frame with multiple
+#'  columns and one row per input data row.
+#' @importFrom stats predict
+#' @exportS3Method
+#' @examples
+#' stop("implement example")
```
danielinteractive, line 51: nice workflow here

R/tmb-methods.R
```diff
@@ -40,43 +40,146 @@ fitted.mmrm_tmb <- function(object, ...) {
...
+predict.mmrm_tmb <- function(
+    object, newdata, se.fit = FALSE, # nolint
+    interval = c("none", "confidence", "prediction"), level = 0.95,
+    na.action = na.pass, n_sim = 1000L, ...) { # nolint
+  interval <- match.arg(interval)
+  # make sure new data has the same levels as original data
```
danielinteractive, line 57: This comment is not easy to understand now in the code context, is it even still needed? maybe not?

R/tmb-methods.R
```diff
@@ -40,43 +40,146 @@ fitted.mmrm_tmb <- function(object, ...) {
...
+  full_frame <- model.frame(
+    object,
+    data = newdata,
+    include = c("subject_var", "visit_var", "group_var", "response_var"),
+    na.action = "na.pass"
+  )
+  tmb_data <- h_mmrm_tmb_data(
+    object$formula_parts, full_frame,
+    weights = rep(1, nrow(full_frame)), reml = TRUE,
+    singular = "keep", drop_visit_levels = FALSE, allow_na_response = TRUE, drop_levels = FALSE
```
danielinteractive, line 67: maybe better have line break after each argument

R/tmb-methods.R
```diff
@@ -40,43 +40,146 @@ fitted.mmrm_tmb <- function(object, ...) {
...
+  }
+  if (interval != "none") { # compute confidence interval
+    alpha <- 1 - level
+    z <- qnorm(1 - alpha / 2) * se
+    res <- cbind(res,
+      lwr = res[, "fit"] - z,
+      upr = res[, "fit"] + z
+    )
+  }
+  if (ncol(res) == 1) { # return vector if only fit is computed
```
danielinteractive, line 87: I would remove the comments here, as the code is clear enough

R/tmb-methods.R
```diff
@@ -40,43 +40,146 @@ fitted.mmrm_tmb <- function(object, ...) {
...
+    res <- res[, "fit"]
+  }
+  return(res)
+}
+
+#' Get Prediction
+#' @description Get prediction with given data, theta, beta, beta_vcov
```
danielinteractive, line 94:
```suggestion
#' @description Get predictions with given `data`, `theta`, `beta`, `beta_vcov`.
```

R/tmb-methods.R
```diff
@@ -40,43 +40,146 @@ fitted.mmrm_tmb <- function(object, ...) {
...
+  assert_class(tmb_data, "mmrm_tmb_data")
+  assert_numeric(theta)
+  n_beta <- ncol(tmb_data$x_matrix)
+  assert_numeric(beta, finite = TRUE, any.missing = FALSE, len = n_beta)
+  assert_matrix(beta_vcov, mode = "numeric", any.missing = FALSE, nrows = n_beta, ncols = n_beta)
+  .Call(`_mmrm_predict`, PACKAGE = "mmrm", tmb_data, theta, beta, beta_vcov)
+}
+
+#' Get Prediction Variance
+#' @description Get prediction variance with given fit, tmb_data with sampling method
```
danielinteractive, line 112:
```suggestion
#' @description Get prediction variance with given fit, `tmb_data` with the Monte Carlo sampling method.
```

R/tmb-methods.R
```diff
@@ -40,43 +40,146 @@ fitted.mmrm_tmb <- function(object, ...) {
...
+#'
+#' @param object (`mmrm_tmb`)\cr the fitted MMRM.
+#' @param n_sim (`integer`)\cr number of replication of sampling.
```
danielinteractive, line 115:
```suggestion
#' @param n_sim (`integer`)\cr number of samples.
```

R/tmb-methods.R
```diff
@@ -40,43 +40,146 @@ fitted.mmrm_tmb <- function(object, ...) {
...
+  assert_int(n_sim)
+  theta_chol <- chol(object$theta_vcov)
+  n_theta <- length(object$theta_est)
+  res <- replicate(n_sim, {
+    z <- rnorm(n = n_theta)
+    theta_sample <- object$theta_est + theta_chol %*% z
+    cond_beta_results <- object$tmb_object$report(theta_sample)
+    beta_mean <- cond_beta_results$beta
+    beta_cov <- cond_beta_results$beta_vcov
+    h_get_prediction(tmb_data, theta_sample, beta_mean, beta_cov)
```
danielinteractive, line 131: did you check whether we might be able to get the Cholesky factor of beta_cov directly from TMB reported? so that we don't need a new Cholesky decomposition in each sampling iteration?

R/tmb-methods.R
```diff
@@ -40,43 +40,146 @@ fitted.mmrm_tmb <- function(object, ...) {
...
+#' Possible options are "response_var", "subject_var", "visit_var" and "group_var", representing the
+#' response variable, subject variable, visit variable or group variable.
 #'
 #' @examples
 #' # Model frame:
 #' model.frame(object)
-#' model.frame(object, exclude = "subject_var")
-model.frame.mmrm_tmb <- function(formula, exclude = "subject_var", full, ...) {
+#' model.frame(object, include = "subject_var")
+model.frame.mmrm_tmb <- function(formula, include = NULL, full, na.action = "na.omit", ...) { # nolint
```
danielinteractive, line 155: here it will be important to have robust testing coverage including edge cases, since now it is quite complex

R/tmb.R
```diff
@@ -134,10 +135,13 @@ h_mmrm_tmb_data <- function(formula_parts,
   data <- data.frame(data, weights)
   # weights is always the last column
   weights_name <- colnames(data)[ncol(data)]
-  if (!identical(getOption("na.action"), "na.omit")) {
-    warning(
-      "NA values will always be removed regardless of na.action in options."
-    )
+  # if y allow to be NA, then first replace y with 1:n, then replace it with original y
```
danielinteractive, line 138: Please use sentence style for all comments.

R/utils-formula.R
```diff
@@ -67,25 +67,28 @@ h_add_covariance_terms <- function(f, covariance) {
...
+    add_terms <- stats::as.formula(sprintf(". ~ . + %s", paste(adds, collapse = "+")))
+    f <- stats::update(f, add_terms)
+  }
+  if (drop_response && length(f) == 3L) {
+    f[[2]] <- NULL
   }
-  drop_terms <- stats::as.formula(sprintf(". ~ . - %s", paste(drops, collapse = " - ")))
-  assert_subset(setdiff(all.vars(drop_terms), "."), all.vars(f))
-  stats::update(f, drop_terms)
+  return(f)
```
danielinteractive, line 93:
```suggestion
  f
```

design/predict_interval/predict_interval.Rmd
```diff
@@ -0,0 +1,76 @@
+---
+title: "predict intervals"
```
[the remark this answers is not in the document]
danielinteractive: this still applies I think - since we don't have a separate design PR now, just merge this content directly into vignettes/predict.Rmd and remove this one

src/chol_cache.h
```diff
@@ -43,6 +49,27 @@ struct lower_chol_nonspatial: public lower_chol_base<Type> {
...
+    } else {
+      matrix<Type> sel_mat = get_select_matrix<Type>(visits, this->n_visits);
+      matrix<Type> ret = sel_mat * sigma_full * sel_mat.transpose();
+      this->sigmas[visits] = ret;
+      return ret;
+    }
+  }
+  matrix<Type> get_sigma_inverse(std::vector<int> visits, matrix<Type> dist) {
+    auto target = this->simgas_inv.find(visits);
+    if (target != this->simgas_inv.end()) {
```
danielinteractive, line 65: there are typos here - do we have tests in place for this and the other new C++ functions?

src/empirical.cpp
```diff
@@ -5,9 +5,9 @@ using std::string;
 // Obtain the empirical given beta, beta_vcov, theta.
 List get_empirical(List mmrm_data, NumericVector theta, NumericVector beta, NumericMatrix beta_vcov, string type) {
   NumericMatrix x = mmrm_data["x_matrix"];
-  matrix<double> x_matrix = as_matrix(x);
+  matrix<double> x_matrix = as_matrix<matrix<double>, NumericMatrix>(x);
```
danielinteractive, line 8: this repeats quite a bit below, similarly for vector, maybe it helps to define a quick helper like `as_num_matrix()`?

src/mmrm.cpp
```diff
@@ -151,8 +149,6 @@ Type objective_function<Type>::operator() ()
   REPORT(beta_vcov);
```
danielinteractive, line 149: in line 115 we already do the decomposition of the beta_cov. that should be sufficient so that we don't have to do it again - if we report it back to the R side?

danielinteractive (commented):
looks very nice, thanks @clarkliming ! please see few comments inside. for the predict tests, also good to check all options (conf intervals, prediction intervals, etc.). I wonder if we could do some level of integration tests by comparing with nlme results e.g.?

R/tmb-methods.R
clarkliming: it is to annotate line 58, where we add another "model.frame" prior to `h_mmrm_tmb_data` (as `h_mmrm_tmb_data` does not know the original data).

so basically it is saying that the seemingly unnecessary `model.frame` call is needed

src/mmrm.cpp
clarkliming: yes no decompose is needed in prediction, only the variance itself is needed

src/empirical.cpp
clarkliming: it is two way conversion between matrix<double> and  NumericMatrix , and we will need two quick helpers for this if needed

R/tmb-methods.R
clarkliming: there is no cholesky decomposition for each sampling iteration; however, there is some cholesky decomposition for each sigma (to construct the sigma from theta need cholesky, and that is what I mentioned previously (which could, also reduce the computation time for autodiff as not cholesky is needed anymore)

src/empirical.cpp
clarkliming: another possibility is to use `as_matrix<matrix<double>>(x)` here, and the second typename can be deduced from the type of x.

src/chol_cache.h
clarkliming: not yet but we can add it

design/predict_interval/predict_interval.Rmd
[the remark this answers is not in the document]
clarkliming: it is combined into the vignettes with a section "Parametric Sampling for prediction interval", mainly covers the conditional variance formula.

src/empirical.cpp
clarkliming: or, if this is still too long, we can use some alias to make it shorter. Anyway we use templates to avoid duplicated code so it is still not so nice to again implement some new functions?

danielinteractive: yeah maybe an alias. I would just like to make it more easily readable

src/kr_comp.cpp
```diff
@@ -4,8 +4,12 @@ using namespace Rcpp;
 using std::string;
 // Obtain P,Q,R element from a mmrm fit, given theta.
 List get_pqr(List mmrm_fit, NumericVector theta) {
+  auto as_num_matrix_tmb = as_matrix<matrix<double>, NumericMatrix>;
```
danielinteractive, line 7: I guess we can't define these shortcuts globally right?

vignettes/predict.Rmd
```diff
@@ -82,6 +82,9 @@ Alternatively, both intervals can be derived using a parametric bootstrap sample
 of the unrestricted parameters $\theta$.
 This would probably also be easier for the `= "prediction"` case.
 
+Please note that for these intervals, we assume that the distribution is normal: we use $estimate \pm Z_{\alpha} * se$
```
danielinteractive, line 85:
```suggestion
Please note that for these intervals, we assume that the distribution is approximately normal: we use $estimate \pm Z_{\alpha} * se$
```

danielinteractive: better than the text "estimate" and "se" here would be to use variable names introduced above

danielinteractive (commented):
Thanks Liming, looks nice! Great work. Do we have enough `predict()` integration tests? Any comparison with `nlme::gls()` possible?

src/kr_comp.cpp
clarkliming: if defined globally there will be conflicts; to be defined globally we do need the function name in header file and the function definition in a cpp file.

what about just using `as_matrix<matrix<double>>()` ? this is shorter and the other type can be deduced.
Further more, we could `typedef matrix<double> mxd;` annd then `as_matrix<mxd>` ?

danielinteractive: hm ok then maybe we keep as is

danielinteractive:
> > Thanks Liming, looks nice! Great work. Do we have enough `predict()` integration tests? Any comparison with `nlme::gls()` possible?
> 
> not yet, I did not find the correct function to compare so I asked `rbmi`. `predict.gls` is just using xβ so it is not conditional on previous obsevations

Ah interesting, so they don't really do the correct way then ... but maybe at least for the new patients we can compare with `predict.gls` result?

clarkliming:
> > > Thanks Liming, looks nice! Great work. Do we have enough `predict()` integration tests? Any comparison with `nlme::gls()` possible?
> > 
> > 
> > not yet, I did not find the correct function to compare so I asked `rbmi`. `predict.gls` is just using xβ so it is not conditional on previous obsevations
> 
> Ah interesting, so they don't really do the correct way then ... but maybe at least for the new patients we can compare with `predict.gls` result?

for this part actually we have the comparison of predict() vs $X\beta$, see https://github.com/openpharma/mmrm/blob/51fb37b03b8e4d035221ad99172d7d4bd2d35a30/tests/testthat/test-tmb-methods.R#L56

is that sufficient?

clarkliming:
it seems that sometimes we have convergence issue, see

image

any ideas why we have this? @danielinteractive

danielinteractive:
Ok then it is sufficient. 

Not sure ... if too flaky we need to change something

danielinteractive (approved):

danielinteractive:
> it seems that sometimes we have convergence issue, see
> 
> image
> 
> any ideas why we have this? @danielinteractive

Could it be a real problem? I would run this locally on this branch and see if you can reproduce it. Since this is a big change we need to be careful

clarkliming:
> > it seems that sometimes we have convergence issue, see
> > image
> > any ideas why we have this? @danielinteractive
> 
> Could it be a real problem? I would run this locally on this branch and see if you can reproduce it. Since this is a big change we need to be careful

the reason why I ask is that I just have everything fine on my local machine, and on github CI sometimes it works sometimes it does not. this is very strange

danielinteractive:
Ok, understand- could be due to different CPU on their machines or so. I would say let's make this test more specific and less reliant on numerical behavior: e.g. define one or two fake optimizers that return false for convergence status and one that returns true. And then we just want to only test that this function tries them as expected

clarkliming:
add some example here where the tests passes (all other tests are removed). So I think this is some by-chance issue and we can update this to make it work

danielinteractive (approved):

PR #286, opened 2023-06-12
add messages about KR linear

### LINKED ISSUE #285: Make a note about `Kenward-Roger-Linear` in `mmrm_control`
When using the Kenward Rogers method it would be nice if it was mentioned in the help-file form mmrm_control, that "Kenward-Roger-Linear" is the one to use (and not "Kenward-Roger"), if you wish to recreate results from SAS where the Kenward-Roger option has been used. It is mentioned in the vignette, but it would be nice to have in the help file also

clarkliming:
close #285

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,6 +18,7 @@
   to specify the maximum number of visits allowed in non-interactive mode, or asked in interactive
   sessions.
 - Add `predict` method to obtain conditional mean estimates and prediction intervals, both with or without previous observations.
+- Add message about SAS and mmrm consistency using Kenward-Roger coefficient covariance and unstructured covariance structure.
 
 ### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


danielinteractive (commented):
Thanks Liming, but I would prefer not to give message. Instead could we either:
- make the default option like this? So detect KR and US and then make vcov KRL?
- otherwise add a note in the mmrm_control documentation

clarkliming:
> Thanks Liming, but I would prefer not to give message. Instead could we either:
> 
> * make the default option like this? So detect KR and US and then make vcov KRL?
> * otherwise add a note in the mmrm_control documentation

currently `mmrm_control` and `formula_parts` are not in the same funciton so difficult to do option 1. I can use a note in `mmrm_control` documentation

danielinteractive (approved):

PR #290, opened 2023-06-13
simplify objective function

### LINKED ISSUE #273: make objective function simpler
currently in `mmrm.cpp`, in objective function, we have some `std::map<int, lower_chol_base<Type>*>` object creation, and some ugly clean up in the end.

we should create a class to hold these information to make the objective function cleaner (no new /delete in objective function but inside the class object)

clarkliming:
close #273 

key changes
1. create a class of chol_by_group to reduce code in main
2. use unique pointers to avoid new/delete


danielinteractive (approved):
nice, thanks @clarkliming !

danielinteractive (approved):

PR #294, opened 2023-06-15
272 reduce replicated

### LINKED ISSUE #272: reduce repeated code in `chol_cache.h` and `derivatives.h`
in `chol_cache.h` and `derivatives.h` we have some duplicated code, like obtain sigma, or cholesky factors.

it is possible to reuse the code, instead of replication

clarkliming:
close #272 

initial attempt to clean up code

### DIFF
--- a/src/derivatives.h
+++ b/src/derivatives.h
@@ -38,7 +38,6 @@ std::map<std::string, matrix<Type>> derivatives(int n_visits, std::string cov_ty
   chol_jacobian chol_jac_obj(n_visits, cov_type);
   matrix<Type> l = chol_obj(theta).matrix();
   l.resize(n_visits, n_visits);
-  ret["chol"] = l;
   vector<Type> chol_d1_vec = autodiff::jacobian(chol_obj, theta).vec(); // chol_d1_vec is (dim * dim * l_theta)
   vector<Type> chol_d2_vec = autodiff::jacobian(chol_jac_obj, theta).vec(); // chol_d2_vec is (dim * dim * l_theta * l_theta)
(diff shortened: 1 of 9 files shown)


src/derivatives.h
```diff
@@ -67,51 +67,36 @@ std::map<std::string, matrix<Type>> derivatives(int n_visits, std::string cov_ty
...
 template <class Type>
-struct derivatives_base {
-  virtual matrix<Type> get_inverse(std::vector<int> visits, matrix<Type> dist) = 0;
+struct derivatives_base: virtual lower_chol_base<Type> {
   virtual matrix<Type> get_inverse_chol(std::vector<int> visits, matrix<Type> dist) = 0;
   virtual matrix<Type> get_sigma_derivative1(std::vector<int> visits, matrix<Type> dist) = 0;
   virtual matrix<Type> get_sigma_derivative2(std::vector<int> visits, matrix<Type> dist) = 0;
-  virtual matrix<Type> get_sigma(std::vector<int> visits, matrix<Type> dist) = 0;
   virtual matrix<Type> get_inverse_derivative(std::vector<int> visits, matrix<Type> dist) = 0;
+  virtual ~derivatives_base() {};
```
clarkliming, line 75: needed because there are multiple inheritance, to avoid the destructor of base class being called

danielinteractive (approved):

src/derivatives.h
danielinteractive: Can you please add that as code comment?

danielinteractive (commented):
Thanks you Liming just one small comment

danielinteractive (approved):

PR #296, opened 2023-06-19
190 simulate method

### LINKED ISSUE #190: Add `simulate()` method for `mmrm_tmb` (or `mmrm` if we need adjusted cov matrix of fixed effects)
Motivation: 1) kind of a standard for models, 2) needed for support in the simulation framework [DHARMa](https://github.com/florianhartig/DHARMa/wiki/Adding-new-R-packages-to-DHARMA#integrating-a-package-with-dharma).

To do:

- [ ] Compare `?simulate` for the general semantics of this generic function for simulating from the predictive distribution
	- stats: simple
	- Ecfun: for `glm`s. Added `newdata` argument.
	- lme4: 
		- `simulate.formula`: which formula to use
		- `simulate.merMod`: simulate from fitted model.
- [ ] Continue prediction vignette by adding simulation section instead of separate design doc or vignette, because they refer to the same predictive distribution
- [ ] Add method
- [ ] Add tests
- [ ] Add to NEWS

danielinteractive:
closes #190


danielinteractive:
@danleibovitz I just opened this draft PR so we have this on the board, can you then move things as discussed? Thanks

R/mmrm-methods.R
```diff
@@ -226,3 +226,50 @@ print.summary.mmrm <- function(x,
...
+    visits <- object$tmb_data$visits_zero_inds[inds] + 1
+
+    # get relevant covariance matrix for subject i
+    covmat_i <- if(grouped){
+        covmat[[object$tmb_data$subject_groups[i]]]
+      }else{
+        covmat
+      }
+    # subset covariance matrix
+    covmat_i <- covmat_i[visits, visits]
```
clarkliming, line 268: @danielinteractive are we considering the simulation conditional on the observed values? to be consistent, I think using conditional simulaiton (just like predict) should be better

clarkliming: this is what I am not sure. but maybe have the functionality of conditional simulation is a plus given that it is implemented in `predict()` ( i am not sure how useful it will be)

danielinteractive: I think we are talking about partially observed visits for a given individual, not partially observed covariates. Covariates always need to be complete. But yeah, `simulate()` should just draw samples from the same predictive distribution summarized by `predict()`

danleibovitz: Thanks @danielinteractive, that's what I meant. Ok, then I can add a test case for partially observed visits to be sure we cover this subcase as well.

R/tmb-methods.R
```diff
@@ -427,3 +427,50 @@ h_residuals_response <- function(object) {
...
+#' @param seed an object specifying if and how the random number generator should be initialized (‘seeded’).
+#' @param newdata data frame for which to evaluate predictions.
+#' @param ...
+#'
+#' @return A \code{data.frame} of dimension [n, m] where n is the number of observations
+#' fitted in the model, and m is the number \code{nsim} of simulated responses.
+simulate.mmrm_tmb <- function(object, nsim = 1,
+                          seed = NULL, newdata = object$data,
+                          ...){
+  if(!is.null(seed)) set.seed(seed)
```
clarkliming, line 447: whether the seed should be reverted back to the original status? (otherwise it is causing some side effects with this seed argument)

R/tmb-methods.R
```diff
@@ -427,3 +427,50 @@ h_residuals_response <- function(object) {
...
+
+  # build data.frame of prediction vectors
+  ret <- predict(object, newdata = newdata, se.fit = FALSE, interval = "none")
```
clarkliming, line 450: I think `newdata` here is confusing that it is only used for the prediction; the covariance is not taken into consideration. It can be true that user provided new data does not match the old data (not ordered by subject and visit etc) can cause issues in the next steps

clarkliming: e.g. newdata and fit$tmb_data do not have the same dimension

R/tmb-methods.R
```diff
@@ -427,3 +427,50 @@ h_residuals_response <- function(object) {
...
+    visits <- object$tmb_data$visits_zero_inds[inds] + 1
+
+    # get relevant covariance matrix for subject i
+    covmat_i <- if(grouped){
+      covmat[[object$tmb_data$subject_groups[i]]]
+    }else{
+      covmat
+    }
+    # subset covariance matrix
+    covmat_i <- covmat_i[visits, visits]
```
clarkliming, line 469: this covariance is not the conditional covariance matrix given some observed response

R/tmb-methods.R
danleibovitz: What's the side-effect being caused here?

R/tmb-methods.R
danleibovitz: The point about data not ordered by subject/visit is a good one -- I'll address this in the next commit and add a test for differently ordered data.

R/tmb-methods.R
danielinteractive: Do we have the conditional covariance matrices calculated inside `predict()` somewhere? Could we expose them in the returned list (maybe optional) or use internal `predict` helpers to get them here?

R/tmb-methods.R
clarkliming: with seed not equal to NULL, it will change the global seed status

R/tmb-methods.R
clarkliming: yes it is calculated at https://github.com/openpharma/mmrm/blob/d485c17240fe960a163fca9e6928925a8c7b9c1b/src/predict.cpp#L102

please note that only the diagonal is used here.

Since the data is not structured (not ordered probably) so maybe it is good to return the full matrix (still might be too large a matrix).  (sparse matrix might be a good option?)

R/tmb-methods.R
```diff
@@ -427,3 +427,55 @@ h_residuals_response <- function(object) {
...
+  n_subjects <- length(subjects) # TODO
+  subject_n_visits <- apply(1:n_sujbects, 1, function(x) sum(subjects[subjects == x])) # TODO
+  subject_zero_inds <- apply(1:n_subjects, 1, function(x) first(subjects[subjects == x])) # TODO get zero inds
+  visit_zero_inds <- # TODO get visits
+
+  for(i in 1:n_subjects){
+
+    # Obtain indices of data.frame belonging to subject i
+    num_visits <- subject_n_visits[i]
+    inds <- (subject_zero_inds[i] + 1) : (subject_zero_inds[i] + num_visits)
```
clarkliming, line 464: so the same issue here, if subject is not ordered, with (subject_zero_inds[i] + 1) : (subject_zero_inds[i] + num_visits) you will not get the records for that subject

R/tmb-methods.R
```diff
@@ -427,3 +427,55 @@ h_residuals_response <- function(object) {
...
+  if(!is.null(seed)) set.seed(seed)
+
+  # build data.frame of prediction vectors
+  ret <- predict(object, newdata = newdata, se.fit = FALSE, interval = "none")
+  ret <- data.frame(replicate(nsim, ret))
+  covmat <- VarCorr(object)
+  grouped <- if(object$tmb_data$n_groups > 1) TRUE else FALSE
+  subjects <- unique(newdata$subject) # TODO
+  n_subjects <- length(subjects) # TODO
+  subject_n_visits <- apply(1:n_sujbects, 1, function(x) sum(subjects[subjects == x])) # TODO
```
clarkliming, line 456: apply can not be applied to a vector; in addition, if subject is not numeric value this could fail

R/mmrm-methods.R
danielinteractive: Are we exploiting everything we can from `predict()` results? Just double checking

R/tmb-methods.R
danielinteractive: do we need the `seed` argument at all? Would it not be cleaner to avoid that internal handling completely?

R/tmb-methods.R
danielinteractive: Ah ok that is very interesting. That refers back to what I asked above re: are we using everything we can from `predict()` - so it seems we could return a bit more from C++ side to make the calculation easy here?
And re: storage- I see that we have this for each individual, I guess then we can just return the full matrix for each individual. Sparse matrix won't help I think because we don't have structure within individuals (at least not with unstructured cov matrix which is typically used)

danielinteractive:
Overall my feeling is that we need to return more from `predict()` to make the calculations easy in `simulate()` and also avoid duplicate code. Do you agree @clarkliming and @danleibovitz ?

R/tmb-methods.R
clarkliming: I mean, with sparse matrix, even if we return the full matrix we don't pay much; the elements between individuals are still missing; the storage is not that much (for each element we store the row/column index). With that we can make it clear if the input data is not correctly ordered.

If we provide the full matrix for each individual, while this works and could be even simpler, we need more documentation there that the returned list of matrix does not have the same order of original input data  (maybe also return the index for each subjects too)

clarkliming:
Hi @danleibovitz the predict vignette is updated and provided some hints on implementation of the simulate method; and `predict` method has some corresponding change should help

danielinteractive (commented):
btw before I forget: please also add your name to the package author list in the DESCRIPTION file

R/tmb-methods.R
```diff
@@ -429,3 +429,152 @@ h_residuals_response <- function(object) {
   assert_class(object, "mmrm_tmb")
   component(object, "y_vector") - unname(fitted(object))
 }
+
+
+#' @describeIn mmrm_tmb_methods Simulates from an mmrm model object
+#' @importFrom stats simulate
```
clarkliming, line 435: we have `package-mmrm.R` to hold all imported function

R/tmb-methods.R
```diff
@@ -429,3 +429,152 @@ h_residuals_response <- function(object) {
...
+#' @param nsim number of response vectors to simulate. Defaults to 1.
+#' @param seed an object specifying if and how the random number generator should be initialized (‘seeded’).
+#' @param newdata data frame for which to evaluate predictions.
+#' @param ...
+#'
+#' @return A \code{data.frame} of dimension [n, m] where n is the number of observations
+#' fitted in the model, and m is the number \code{nsim} of simulated responses.
+simulate.mmrm_tmb <- function(object, nsim = 1,
+                              newdata = NULL,
+                              method = "conditional",
```
clarkliming, line 448:
```suggestion
                              method = c("conditional", "marginal"),
```

R/tmb-methods.R
```diff
@@ -429,3 +429,152 @@ h_residuals_response <- function(object) {
...
+                              ...){
+
+  method <- match.arg(method)
+
+  if(is.null(newdata)){
+    # build data.frame of prediction vectors
+    ret <- fitted(object)
```
clarkliming, line 455: in this way we ignore the missingness in the data. This is inconsistent with the new data approach

R/tmb-methods.R
```diff
@@ -429,3 +429,152 @@ h_residuals_response <- function(object) {
...
+      covmat_i <- if(grouped){
+        covmat[[object$tmb_data$subject_groups[i]]]
+      }else{
+        covmat
+      }
+      # subset covariance matrix
+      covmat_i <- covmat_i[visits, visits]
+
+      # simulate from covariance matrix
+      ret[inds,] <- ret[inds, , drop=FALSE] + MASS::mvrnorm(nsim, rep.int(0, num_visits), covmat_i)
```
clarkliming, line 485: MASS::mvrnorm creates a matrix of size (nsim, num_visits); and `ret` seems like `(num_visits, nsim) shaped data.frame

R/tmb-methods.R
```diff
@@ -429,3 +429,152 @@ h_residuals_response <- function(object) {
...
+    }
+  }else if(method == "conditional"){
+    # order data.frame
+    newdata <- dplyr::arrange(newdata, !!object$formula_parts$subject_var, !!object$formula_parts$visit_var)
```
clarkliming, line 489: reordering can cause issues as the input and output are not corresponding to each other

clarkliming: and this reordering can be wrong; suggest to use `model.frame.mmrm_tmb` to obtain the correct data

clarkliming: e.g. data and fit do not have identical levels

R/tmb-methods.R
```diff
@@ -429,3 +429,152 @@ h_residuals_response <- function(object) {
...
+#' @return A \code{data.frame} of dimension [n, m] where n is the number of observations
+#' fitted in the model, and m is the number \code{nsim} of simulated responses.
+simulate.mmrm_tmb <- function(object, nsim = 1,
+                              newdata = NULL,
+                              method = "conditional",
+                              ...){
+
+  method <- match.arg(method)
+
+  if(is.null(newdata)){
```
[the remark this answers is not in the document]
danleibovitz: Agreed, this is much more elegant

R/tmb-methods.R
```diff
@@ -429,3 +429,114 @@ h_residuals_response <- function(object) {
...
+
+      for(i in 1:n_subjects){
+        # Obtain indices of data.frame belonging to subject i (iterate by 1, since indices from cpp are 0-order)
+        inds <- mu$index[[i]] + 1
+
+        # get relevant covariance matrix for subject i
+        covmat_i <- mu$covariance[[i]]
+
+        # simulate from covariance matrix
+        ret[inds,] <- ret[inds, , drop=FALSE] + MASS::mvrnorm(nsim, rep.int(0, length(inds)), covmat_i)
```
clarkliming, line 501: `MASS::mvrnorm(nsim, rep.int(0, length(inds)), covmat_i)` return a matrix of size `(nsim, length(inds)` and the `ret[inds,]` shape is its transpose

R/tmb-methods.R
```diff
@@ -429,3 +429,114 @@ h_residuals_response <- function(object) {
...
+    ret <- mu$prediction[,1]
+    ret <- data.frame(replicate(nsim, ret))
+
+    # get dataframe values
+    n_subjects <- tmb_data$n_subjects
+    subject_n_visits <- tmb_data$subject_n_visits
+    subject_zero_inds <- tmb_data$subject_zero_inds
+    visit_zero_inds <- tmb_data$visits_zero_inds
+
+      for(i in 1:n_subjects){
```
clarkliming, line 493:
```suggestion
      for(i in seq_len(n_subjects)) {
```

seq_len is usually robust than 1:n_subjcts

R/tmb-methods.R
```diff
@@ -429,3 +429,114 @@ h_residuals_response <- function(object) {
...
+
+        # Obtain indices of data.frame belonging to subject i
+        inds <- mu$index[[i]] + 1
+
+        # get relevant covariance matrix for subject i
+        covmat_i <- mu$covariance[[i]]
+
+        # simulate from covariance matrix
+        ret[inds,] <- ret[inds, , drop=FALSE] + MASS::mvrnorm(nsim, rep.int(0, length(inds)), covmat_i)
```
clarkliming, line 535: n_sim has to be one here

R/tmb-methods.R
danleibovitz: Yep, that's a very good catch, thank you.

R/tmb-methods.R
```diff
@@ -429,3 +429,114 @@ h_residuals_response <- function(object) {
...
+          # Obtain indices of data.frame belonging to subject i
+          inds <- mu$index[[i]] + 1
+
+          # get relevant covariance matrix for subject i
+          covmat_i <- mu$covariance[[i]]
+
+          # simulate from covariance matrix
+          ret[inds,j] <- ret[inds, j, drop=FALSE] + t(MASS::mvrnorm(1, rep.int(0, length(inds)), covmat_i))
+          }
+        }
```
clarkliming, line 536: this part seems a bit duplicated with the previous part; is it possible to reduce the duplication, like

```
h_fun <- function(mu, cov, n_sim) {
xx
}
```
so that in the conditional part we have

```
h_fun(mu, cov, nsim)
```

and in marginal part we have

```
lapply(seq_len(nsim), function(x) {
derivation of new theta; mu; cov; 
h_fun(mu, cov, 1L)
}
```

R/tmb-methods.R
```diff
@@ -429,3 +429,114 @@ h_residuals_response <- function(object) {
...
+#' @importFrom stats simulate
+#' @exportS3Method
+#'
+#' @param object an object of class 'mmrm'
+#' @param nsim number of response vectors to simulate. Defaults to 1.
+#' @param seed an object specifying if and how the random number generator should be initialized (‘seeded’).
+#' @param newdata data frame for which to evaluate predictions.
+#' @param ...
+#'
+#' @return A \code{data.frame} of dimension [n, m] where n is the number of observations
```
clarkliming, line 444: ususally using back quote is sufficient; but this may aslo work (i am not sure)

R/tmb-methods.R
```diff
@@ -429,3 +429,114 @@ h_residuals_response <- function(object) {
...
+#' fitted in the model, and m is the number \code{nsim} of simulated responses.
```
clarkliming, line 445: n is the rows of new data but not the rows in the fit object

R/tmb-methods.R
```diff
@@ -429,3 +429,114 @@ h_residuals_response <- function(object) {
...
   component(object, "y_vector") - unname(fitted(object))
 }
+
+
+#' @describeIn mmrm_tmb_methods Simulates from an mmrm model object
+#' @importFrom stats simulate
+#' @exportS3Method
+#'
+#' @param object an object of class 'mmrm'
+#' @param nsim number of response vectors to simulate. Defaults to 1.
```
clarkliming, line 439: usually we use something like (`int`)\cr as the type hint

R/tmb-methods.R
danielinteractive: yeah better use `` `data.frame` `` etc.

tests/testthat/test-tmb-methods.R
```diff
@@ -646,3 +646,49 @@ test_that("response residuals helper function works as expected", {
   expected <- component(object, "y_vector") - as.vector(fitted(object))
   expect_equal(result_rsp, expected)
 })
+
+# simulate.mmrm_tmb ----
+
+test_that("simulate.mmrm returns a data.frame object of correct dimension and with no missing values", {
+  object <- get_mmrm()
```
danielinteractive, line 653: always set a seed before `simulate()` calls in these tests (bit safer)

tests/testthat/test-tmb-methods.R
```diff
@@ -646,3 +646,49 @@ test_that("response residuals helper function works as expected", {
...
+  sims <- simulate(object, nsim = 2)
+  expect_data_frame(sims, any.missing = FALSE, nrows = nrow(object$data), ncols = 2)
+
+  # check that when nsim == 1, we still have a single-column data.frame
```
danielinteractive, line 657: better move comment text into `test_that()` label text, and make a new test if needed for this

tests/testthat/test-tmb-methods.R
```diff
@@ -646,3 +646,49 @@ test_that("response residuals helper function works as expected", {
...
+  sims <- simulate(object, nsim = 1)
+  expect_data_frame(sims, any.missing = FALSE, nrows = nrow(object$data), ncols = 1)
+})
+
+test_that("simulate works as expected for standard models", {
+  # simulate.mmrm values are correctly centered
```
danielinteractive, line 663: move into test label text

tests/testthat/test-tmb-methods.R
```diff
@@ -646,3 +646,49 @@ test_that("response residuals helper function works as expected", {
...
+test_that("simulate works as expected differently ordered/numbered data", {
+  object <- get_mmrm()
+  # construct trivial data.frame
+  df <- dplyr::slice(object$data, 1:10)
+  # TODO reorder newdata
+  sims <- simulate(object, nsim = 1000, newdata = df)
+  expect_equal(rowMeans(sims), predict(object, df), tolerance = 1e-1)
+})
+
+# test_that("simulate works as expected for spatial fits", {
```
danielinteractive, line 690: remove commented code

danielinteractive: or can we make it work for spatial? that would of course be good

R/tmb-methods.R
```diff
@@ -429,3 +429,114 @@ h_residuals_response <- function(object) {
...
+    weights = rep(1, nrow(full_frame)),
+    reml = TRUE,
+    singular = "keep",
+    drop_visit_levels = FALSE,
+    allow_na_response = TRUE,
+    drop_levels = FALSE
+  )
+
+  if(method == "conditional"){
+    # get prediction and prediction variance
```
danielinteractive, line 476: can we reduce comments, e.g. by clear variable names or helper functions?
if we have comments, please use sentence style

danielinteractive (commented):
Thanks @danleibovitz and also @clarkliming for the review team work, @danleibovitz you can also add yourself to package authors

R/tmb-methods.R
danleibovitz: All the other `describeIn` tags point to `mmrm_tmb_methods`. @clarkliming are you thinking we should change this?

tests/testthat/test-tmb-methods.R
danleibovitz: At this point, would it be better to merge this branch, and open a new issue (time/resource permitting) for spatial fits?

DESCRIPTION
```diff
@@ -18,7 +18,8 @@ Authors@R: c(
     person("Gilead Sciences, Inc.", role = c("cph", "fnd")),
     person("F. Hoffmann-La Roche AG", role = c("cph", "fnd")),
     person("Merck Sharp & Dohme, Inc.", role = c("cph", "fnd")),
-    person("AstraZeneca plc", role = c("cph", "fnd"))
+    person("AstraZeneca plc", role = c("cph", "fnd")),
+    person("Daniel", "Leibovitz", , "daniel.leibovitz@roche.com", role = "ctb")
```
clarkliming, line 22: maybe we can move the human names in front of company names

R/tmb-methods.R
clarkliming: oh true for these methods we have additional `importFrom`. then consistency is fine

R/tmb-methods.R
```diff
@@ -429,3 +429,91 @@ h_residuals_response <- function(object) {
...
+    drop_levels = FALSE
+  )
+
+  if(method == "conditional"){
+    mu <- h_get_prediction(tmb_data, object$theta_est, object$beta_est, object$beta_vcov)
+    ret <- as.data.frame(h_get_sim_per_subj(mu, tmb_data$n_subjects, nsim))
+  }else if(method == "marginal"){
+    ret <- as.data.frame(
+      sapply(seq_len(nsim), function(x) {
+        newtheta <- MASS::mvrnorm(1, object$theta_est, object$theta_vcov)
```
clarkliming, line 476: if possible, this `newtheta` can be generated in the outer loop. (can be more efficient I think because in mvrnorm there are some expensive computations)

R/tmb-methods.R
```diff
@@ -429,3 +429,91 @@ h_residuals_response <- function(object) {
...
+    allow_na_response = TRUE,
+    drop_levels = FALSE
+  )
+
+  if(method == "conditional"){
+    mu <- h_get_prediction(tmb_data, object$theta_est, object$beta_est, object$beta_vcov)
+    ret <- as.data.frame(h_get_sim_per_subj(mu, tmb_data$n_subjects, nsim))
+  }else if(method == "marginal"){
+    ret <- as.data.frame(
+      sapply(seq_len(nsim), function(x) {
```
clarkliming, line 475: sapply add additional simplification on lapply; but here I think anyway you are converting it to a data.frame and lapply should be sufficient

R/tmb-methods.R
```diff
@@ -429,3 +429,91 @@ h_residuals_response <- function(object) {
...
+
+#' Get simulated values by patient
+#'
+#' @param mu (`list`)\cr object returned by `h_get_prediction()`
+#' @param nsub (`integer`)\cr number of subjects
+#' @param nsim (`integer`)\cr number of values to simulate
+#'
+#' @keywords internal
+h_get_sim_per_subj <- function(mu, nsub, nsim){
+
```
clarkliming, line 498: can add some assertions here

R/tmb-methods.R
```diff
@@ -429,3 +429,91 @@ h_residuals_response <- function(object) {
...
+  ret <- matrix(replicate(nsim, mu$prediction[,1]), ncol = nsim)
```
clarkliming, line 499:
```suggestion
  ret <- matrix(mu$prediction[,1], ncol = nsim, nrow = nrow(mu$prediction))
```

this should be much efficient

R/tmb-methods.R
danielinteractive: As discussed, can create the matrix of all `newtheta` vectors before going into `sapply` or loop to only need one covariance matrix decomposition, hence that will be faster

R/tmb-methods.R
```diff
@@ -429,3 +429,92 @@ h_residuals_response <- function(object) {
...
+
+
+#' Get simulated values by patient
+#'
+#' @param mu (`list`)\cr object returned by `h_get_prediction()`
+#' @param nsub (`integer`)\cr number of subjects
+#' @param nsim (`integer`)\cr number of values to simulate
+#'
+#' @keywords internal
+h_get_sim_per_subj <- function(mu, nsub, nsim) {
```
danielinteractive, line 495: also please add unit tests for this helper function

R/tmb-methods.R
```diff
@@ -429,3 +429,92 @@ h_residuals_response <- function(object) {
...
+  assert_integer(nsim)
+
+  ret <- matrix(mu$prediction[, 1], ncol = nsim, nrow = nrow(mu$prediction))
+
+  for (i in seq_len(nsub)) {
+    if (length(mu$index[[i]]) > 0) {
+      # Obtain indices of data.frame belonging to subject i (iterate by 1, since indices from cpp are 0-order)
+      inds <- mu$index[[i]] + 1
+
+      # get relevant covariance matrix for subject i
```
danielinteractive, line 507:
```suggestion
      # Get relevant covariance matrix for subject i.
```

R/tmb-methods.R
```diff
@@ -429,3 +429,91 @@ h_residuals_response <- function(object) {
...
+      covmat_i <- mu$covariance[[i]]
+
+      # simulate from covariance matrix
+      if(nsim > 1){
+        ret[inds,] <- ret[inds, , drop = FALSE] + t(MASS::mvrnorm(nsim, rep.int(0, length(inds)), covmat_i))
+      }else{
+        ret[inds,] <- ret[inds, , drop = FALSE] + MASS::mvrnorm(nsim, rep.int(0, length(inds)), covmat_i)
+      }
```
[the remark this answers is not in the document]
danleibovitz: We'll assign segments before-hand, and put the if-else condition on a single line.

R/tmb-methods.R
clarkliming: oh right. the as.data.frame is slow for the list

tests/testthat/test-tmb-methods.R
```diff
@@ -665,35 +665,38 @@ test_that("simulate values are correctly centered for standard models", {
...
-  expect_equal(rowMeans(sims), predict(object, df), tolerance = 1e-1)
+  sims_mixed <- sims[neworder, ]
+  expect_equal(rowMeans(sims_mixed), predict(object, df_mixed), tolerance = 1e-1)
 })
 
 # h_get_sim_per_subj ----
 
-test_that("h_get_sim_per_subj", {
+test_that("h_get_sim_per_subj works for nsim == 1", {
+  h_get_sim_per_subj(mu, nsub, nsim)
```
danielinteractive, line 700: please compare result with expectation, and also add more tests e.g. for nsim > 1 and potential edge cases

danleibovitz:
I believe we're ready to merge :)

R/tmb-methods.R
```diff
@@ -430,20 +430,24 @@ h_residuals_response <- function(object) {
...
 
-
-#' @describeIn mmrm_tmb_methods Simulates from an mmrm model object
+#' @describeIn mmrm_tmb_methods Simulate from an mmrm model object
 #' @importFrom stats simulate
-#' @exportS3Method
 #'
-#' @param object (`mmrm`)\cr an fitted 'mmrm' object
+#' @param object (`mmrm`)\cr a fitted 'mmrm' object
+#' @param nsim (`integr`)\cr number of replications to simulate.
```
danielinteractive, line 437:
```suggestion
#' @param nsim (`integer`)\cr number of replications to simulate.
```

R/tmb-methods.R
```diff
@@ -430,20 +430,24 @@ h_residuals_response <- function(object) {
...
+#' @param seed Unused argument from simulate generic.
```
danielinteractive, line 438:
```suggestion
#' @param seed unused argument from simulate generic.
```

R/tmb-methods.R
```diff
@@ -430,20 +430,24 @@ h_residuals_response <- function(object) {
...
+#' @param newdata (`data.frame`)\cr new data.frame for which to simulate values.
+#' @param method (`character`)\cr One of "conditional" or "marginal". If "conditional",
```
danielinteractive, line 440:
```suggestion
#' @param method (`string`)\cr One of "conditional" or "marginal". If "conditional",
```

R/tmb-methods.R
```diff
@@ -470,11 +474,12 @@ simulate.mmrm_tmb <- function(object, nsim = 1,
...
     mu <- h_get_prediction(tmb_data, object$theta_est, object$beta_est, object$beta_vcov)
     ret <- as.data.frame(h_get_sim_per_subj(mu, tmb_data$n_subjects, nsim))
-
   } else if (method == "marginal") {
+    theta_chol <- t(chol(object$theta_vcov))
+    ntheta <- length(object$theta_est)
     ret <- as.data.frame(
       sapply(seq_len(nsim), function(x) {
-        newtheta <- MASS::mvrnorm(1, object$theta_est, object$theta_vcov)
+        newtheta <- theta_chol %*% t(MASS::mvrnorm(1, rep(0, ntheta), diag(ntheta))) + object$theta_est
```
danielinteractive, line 482: `matrix(rnorm(...), ..., ...)` could be simpler and faster

danleibovitz: I replaced `MASS::mvrnorm()` in this function as well as the helper function -- which has the added benefit of removing the awkward ifelse syntax for separating `mvrnorm(n=1)` from `mvrnorm(n>1)`.

danleibovitz:
@danielinteractive the latest check failed on processing the introduction vignette, with the message `there is no package called 'tidymodels'` (check # 1526). This seems unrelated to any changes made on this branch. Have you seen this error before, or do you have any insight here?

DESCRIPTION
```diff
@@ -14,6 +14,7 @@ Authors@R: c(
     person("Craig", "Gower-Page", , "craig.gower-page@roche.com", role = "ctb"),
     person("Dan", "James", , "dan.james@astrazeneca.com", role = "aut"),
     person("Jonathan", "Sidi", , "yoni.sidi@sagerx.com", role = "aut"),
+    person("Daniel", "Leibovitz", , "daniel.leibovitz@roche.com", role = "ctb"),
```
danielinteractive, line 17:
```suggestion
    person("Daniel", "Leibovitz", , "daniel.leibovitz@roche.com", role = "aut"),
```

danielinteractive (commented):
Congratulations @danleibovitz!

R/tmb-methods.R
```diff
@@ -429,3 +429,102 @@ h_residuals_response <- function(object) {
...
+
+  if (method == "conditional") {
+    mu <- h_get_prediction(tmb_data, object$theta_est, object$beta_est, object$beta_vcov)
+    ret <- as.data.frame(h_get_sim_per_subj(mu, tmb_data$n_subjects, nsim))
+  } else if (method == "marginal") {
+    theta_chol <- t(chol(object$theta_vcov))
+    ntheta <- length(object$theta_est)
+    ret <- as.data.frame(
+      sapply(seq_len(nsim), function(x) {
+        newtheta <- theta_chol %*% matrix(rnorm(ntheta), nrow = 1) + object$theta_est
```
danielinteractive, line 482: hm this cannot work with just one row but needs one column because the cholesky matrix will have more than one column?
```suggestion
        newtheta <- theta_chol %*% matrix(rnorm(ntheta), ncol = 1) + object$theta_est
```

danielinteractive (commented):
Hi @danleibovitz  just had a last review round and still saw some issues, see below

R/tmb-methods.R
```diff
@@ -429,3 +429,102 @@ h_residuals_response <- function(object) {
...
+      inds <- mu$index[[i]] + 1
+      obs <- length(inds)
+
+      # Get relevant covariance matrix for subject i.
+      covmat_i <- mu$covariance[[i]]
+      theta_chol <- chol(covmat_i)
+
+      # Simulate from covariance matrix.
+      mus <- ret[inds, , drop = FALSE]
+      sigs <- matrix(rnorm(nsim * obs), nrow = nsim) %*% theta_chol
```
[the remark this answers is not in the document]
danielinteractive: let's do it consistent with the way we do it above

R/tmb-methods.R
```diff
@@ -429,3 +429,102 @@ h_residuals_response <- function(object) {
...
+      ret[inds, ] <- mus + t(sigs)
```
danielinteractive, line 525: probably this needs to be 
```suggestion
      ret[inds, ] <- sigs + mus
```

tests/testthat/test-tmb-methods.R
```diff
@@ -646,3 +646,149 @@ test_that("response residuals helper function works as expected", {
...
+  set.seed(202)
+  sims <- simulate(object, nsim = 1)
+  expect_data_frame(sims, any.missing = FALSE, nrows = nrow(object$data), ncols = 1)
+})
+
+test_that("simulate values are correctly centered for standard models", {
+  object <- get_mmrm()
+  set.seed(323)
+  sims <- simulate(object, nsim = 1000)
+  expect_equal(rowMeans(sims), unname(predict(object, object$data)), tolerance = 1e-1)
```
danielinteractive, line 668: can save predict result and then in addition to the means also compare the standard deviation and the interval boundaries (just with 95% confidence level)

danielinteractive:
@danleibovitz Thanks Dan, just looked at the tests. I am seeing a couple of (maybe not related) issues (and not your fault of course, just trying to summarize the current state):
- differently ordered/numbered data does not seem to work consistently between simulate/predict
- predict with confidence intervals leads to se = 0
- predict with prediction intervals is quite slow
- simulate with marginal method (also slow but that is expected) gives biased results vs. predict with prediction intervals

clarkliming:
Forgive me for my silly error in predict:

1. R and C use different cholesky decomposition
2. to obtain mean of variance and variance of mean, by mistake I use mean of mean and variance of variance.

The speed of predict is also slow because this is still using simulation to obtain the marginal prediction error

danielinteractive:
Thanks @clarkliming , cool that you could find the problem so quickly!
OK I think we can keep the speed issue as is for now as we don't have a great idea at the moment how to speed it up.

How about these:
- differently ordered/numbered data does not seem to work consistently between simulate/predict
- predict with confidence intervals leads to se = 0 (or maybe that is a feature?)

are they solved or not yet?

clarkliming:
> * differently ordered/numbered data does not seem to work consistently between simulate/predict

sorry this is not yet solved

> * predict with confidence intervals leads to se = 0 (or maybe that is a feature?)

this one is a feature (because for observed data I set se = 0 because no more prediction is needed right?)

danielinteractive:
Thanks @clarkliming , @danleibovitz would you have time to look at the ordering issue? (see my note in the test) after that it should be more or less good then

clarkliming:
i have updated the order to make sure the result has the same order with the input data frame. After that I think we can clean up the comments and proceed.

danielinteractive:
@danleibovitz do you have time for cleanup?

danielinteractive:
OK I will quickly clean up

danielinteractive:
@clarkliming would you maybe have time to look at the conditional sims quantiles vs. predict with confidence intervals? https://github.com/openpharma/mmrm/blob/190-simulate-method/tests/testthat/test-tmb-methods.R#L894
Or don't they need to match?

clarkliming:
hi @danielinteractive let me summarize as follows

1. predict(type = "confidence") gives the variance of "prediction" conditional on $\theta$; here we always use $X \beta + E(\Sigma|obseved values)$ as the prediction. So here we ignore the variance of $\epsilon$ and variance of $\beta$. We can say that this "prediction" is a kind of prediction of mean (we use mean as our prediction)
2. predict(type = "prediction") gives the real variance considering both $\epsilon$ and variance of $\beta$, $\theta$.

1. simulate(type = "marginal") gives simulation that is not conditional on $\theta$ but is marginal on $\theta$. it also take $\epsilon$ into consideration so this one matches $predict(type = "prediction")$
2. simulate(type = "conditional") gives simulation conditional on $\theta$, and it also take $\epsilon$ into consideration. the variance of the mean of the conditional simulation will be within the confidence intervals.

so what we can test

1. predict(type = "prediction") and variance of simulate(type = 'marginal') 
2. predict(type = "confidence") and **mean** of simulate(type = "conditional"). The value should fall within the range of confidence interval
3. sample $\beta$ for multiple times given $\theta$, and for each $\beta$ we can obtain the **mean** of simulate(type = "conditional"). Then the variance of this mean, should be consistent with the `predict(type = "confidence")`. However this can be too much computation

clarkliming:
also updated some tests; simulation of marginal have larger tolerance because it has larger sd and smaller number of repetitions

danielinteractive:
Thanks @clarkliming , I have included this overview in the predict vignette since it will be helpful for users.

I have one doubt though (still, sorry):
As we describe in https://github.com/openpharma/mmrm/blob/190-simulate-method/vignettes/predict.Rmd#L184 and I think as we want it, the simulation with "marginal" method should take into account the full uncertainty - i.e. for all of  `theta`, `beta`, `epsilon`. However currently looking at https://github.com/openpharma/mmrm/blob/190-simulate-method/R/tmb-methods.R#L619, we indeed take into account the `theta` uncertainty, and we also generate `epsilon` so that is fine. However the `beta` we don't sample conditional on `theta`, so we don't take into account the uncertainty of `beta` as we e.g. do in https://github.com/openpharma/mmrm/blob/190-simulate-method/R/tmb-methods.R#L181. I would suggest we should sample `beta` from multivariate normal first and then pass to `h_get_prediction`. 
What do you think? @clarkliming @danleibovitz

danielinteractive:
I just added explanation how to simulate `beta`, this does not cost much additional time because we have the inverse `beta_vcov` matrix decomposition already

danleibovitz:
Thanks @danielinteractive ! I've read through your comments, and this makes sense to me (although with $\beta_{sample}$ instead of simple $\beta_{new}$ , the marginal simulation will take even longer to converge, and I wonder if we'll have to adjust the tolerance of the tests. I'll check that as well). I can implement this tomorrow.

danielinteractive:
Thanks @danleibovitz please push what you have or let me know if you don't have time, because I aim to finish this PR tomorrow latest, thanks!

R/tmb-methods.R
```diff
@@ -624,7 +624,9 @@ simulate.mmrm_tmb <- function(object,
         newtheta <- theta_chol %*% matrix(rnorm(ntheta), ncol = 1) + object$theta_est
         # Recalculate betas with sampled thetas.
         hold <- object$tmb_object$report(newtheta)
-        predict_res <- h_get_prediction(tmb_data, newtheta, hold$beta, hold$beta_vcov)
+        # Resample betas given new beta distribution.
+        beta_sample <- hold$beta + VCOV %*% rnorm(length(hold$beta))
```
danleibovitz, line 628: I might be fundamentally misunderstanding something here. Sampling $\beta_{sample}$ from the given distribution should be a very simple one-liner here. In fact, we have `beta_vcov` from `object$tmb_object$report()` that we could call directly, but we want to speed this up by returning $L$ and $D$ from the `C++` side. BUT, then $L$ and $D$ need to functions of the covariance of $\beta$, whereas if I'm reading right they're functions of $\theta$, no?

FYI: `VCOV` is just a placeholder, hence the canceled checks.

R/tmb-methods.R
```diff
@@ -625,7 +625,7 @@ simulate.mmrm_tmb <- function(object,
         # Recalculate betas with sampled thetas.
         hold <- object$tmb_object$report(newtheta)
         # Resample betas given new beta distribution.
-        beta_sample <- hold$beta + VCOV %*% rnorm(length(hold$beta))
+        beta_sample <- hold$beta + t(chol(hold$beta_vcov)) %*% rnorm(length(hold$beta))
```
danleibovitz, line 628: For clarity, this is the more time-consuming approach, as I understand it, in which we directly use the covariance matrix `beta_vcov` as provided by `report()`.

R/tmb-methods.R
danielinteractive: So we have a 1:1 correspondence between beta_vcov and L&D. Both are functions of theta and we can modify the C++ side to also report L&D

danielinteractive: But I can also do this tomorrow myself no worries if it is not clear yet.

R/tmb-methods.R
clarkliming: this is time consuming but it seems that we have to include this to sample $\beta$

danleibovitz (approved):
It looks good! I suppose the speed improvements from passing L and D from the C++ side now hinge on how much faster `backsolve()` is than `chol()`. But what matters most of all is, all tests passed :) 

May I squash & merge?

danielinteractive:
Thanks @danleibovitz, yeah `backsolve()` is much more efficient than `chol()` (I think complexity is similar as in a matrix multiplication). 
Feel free to press the green button :)

PR #307, opened 2023-06-29
return covariance for each subject

### LINKED ISSUE #306: return (conditional) covariance matrix in `predict` function
currently only the diagonal elements are returned, and to support simulate method, the full matrix is needed.

clarkliming:
close #306

the covariance are returned (for those with missing values to be simulated, conditional on existing values)


man/mmrm_tmb_methods.Rd
```diff
@@ -155,7 +155,8 @@ object <- fit_mmrm(formula, fev_data, weights = rep(1, nrow(fev_data)))
 coef(object)
 # Fitted values:
 fitted(object)
-predict(object, newdata = fev_data)
+fit <- mmrm(FEV1 ~ RACE + SEX + ARMCD * AVISIT + us(AVISIT | USUBJID), data = fev_data)
```
danielinteractive, line 158: Correct? Did it not work with object?

danielinteractive (commented):
Very nice! Thanks @clarkliming

man/mmrm_tmb_methods.Rd
clarkliming: my bad. it should have worked. I did not see the object definition on the top.

clarkliming:
the speed of `prediction(object, se.fit = T, interval = "prediction", n_sim= xxx)` has around 10% percent efficiency loss with new `predict` function but still acceptable I think

danielinteractive:
@clarkliming yeah that is fine. thanks for checking!

danielinteractive (approved):

PR #313, opened 2023-07-10
311: Catch divergence errors

### LINKED ISSUE #311: Certain errors from the function or optimizer evaluation should be caught
Examples:

- `L-BFGS-B needs finite values of 'fn'`
- `NA/NaN Hessian evaluation`

danielinteractive:
closes #311

### DIFF
--- a/R/fit.R
+++ b/R/fit.R
@@ -40,6 +40,16 @@ fit_single_optimizer <- function(formula,
                                  formula_parts,
                                  ...,
                                  control = mmrm_control(...)) {
+  to_remove <- list(
+    # Transient visit to invalid parameters.
+    warnings = c("NA/NaN function evaluation")
+  )
+  as_diverged <- list(
+    errors = c(
+      "NA/NaN Hessian evaluation",
+      "L-BFGS-B needs finite values of 'fn'"
+    )
+  )
   if (missing(tmb_data) || missing(formula_parts)) {
     h_valid_formula(formula)
     assert_data_frame(data)
@@ -56,9 +66,8 @@ fit_single_optimizer <- function(formula,
         covariance = covariance,
         control = control
       ),
-      remove = list(
-        warnings = c("NA/NaN function evaluation") # Transient visit to invalid parameters.
-      )
+      remove = to_remove,
+      divergence = as_diverged
     )
   } else {
     assert_class(tmb_data, "mmrm_tmb_data")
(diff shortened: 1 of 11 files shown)


.Rbuildignore
```diff
@@ -17,3 +17,4 @@ simulations
 ^doc$
 ^Meta$
 ^data-raw$
+cache
```
danielinteractive, line 20:
```suggestion
^cache$
```

R/fit.R
```diff
@@ -415,12 +418,21 @@ mmrm <- function(formula,
...
+      formula_parts = formula_parts,
+      control = control
+    )
+    if (length(fit) == 0) {
+      warning(paste0(
+        "Divergence with optimizer ", names(control$optimizers[1L]), " due to problems: ",
+        toString(attr(fit, "divergence"))
+      ))
+    }
+    control$optimizers <- control$optimizers[-1]
```
clarkliming, line 434: if all optimizer fails, `fit` is length 0 and there are no successful optimizers. then is there need to call `refit_multiple_optimizers` again?

R/fit.R
```diff
@@ -40,6 +40,16 @@ fit_single_optimizer <- function(formula,
...
+  to_remove <- list(
+    # Transient visit to invalid parameters.
+    warnings = c("NA/NaN function evaluation")
+  )
+  as_diverged <- list(
+    errors = c(
+      "NA/NaN Hessian evaluation",
+      "L-BFGS-B needs finite values of 'fn'"
+    )
+  )
```
clarkliming, line 52: if this is something static, should it be defined somewhere not inside the function but in the package environment?

danielinteractive: I think it is ok, we had it like that before too for the warnings to remove - it is not used by any other functions e.g.

R/fit.R
danielinteractive: this first while part is only finding the first optimizer returning a non-empty fit. Exactly, if all fail then below we will get the error to try multiple optimizers and refit_multiple_optimizers is not called. also have a look at the unit tests I added for these edge cases

clarkliming, line 434: so this part we modified the `control$optimizers`, and in line 451 we still need the names of all these optmizers to report the error. currently it says

```
Chosen optimizer '' led to problems during model fit:
```

R/fit.R
```diff
@@ -415,12 +418,21 @@ mmrm <- function(formula,
...
-
+  fit <- list()
+  while ((length(fit) == 0) && length(control$optimizers) > 0) {
+    fit <- fit_single_optimizer(
+      tmb_data = tmb_data,
+      formula_parts = formula_parts,
+      control = control
+    )
+    if (length(fit) == 0) {
+      warning(paste0(
```
clarkliming, line 429: I wonder if we really need this warning (because other divergent issues might be captured by the optimizers, like different convergence code; and other non-converged issue happens we just silently go to other optimizers).  Maybe also good to capture all the warning/errors in `do.call` part of the optimizer so that anyway we have a returned fit in `fit_mmrm` but all the errors/logs are stored. (`tmb_opt` result can be manually constructed with `list(value = NA, par = initial values etc, convergence = 1 etc, error = caught error)`

danielinteractive: Good ideas! I thought about this a bit more after our stand up.

- I think I would not move the capturing of errors and warnings further inside, because `fit_mmrm` (one level down) is supposed to be low level, i.e. not capture these. (and if we would consider this, it would be quite big change with separate issue/PR)
- So far all the examples we have seen recently about these divergence errors/warnings were actually pointing towards overparametrized covariance models (instead of just unfortunate optimizer behavior). Therefore I think the warning makes sense to alert the user, while still trying other optimizers.

clarkliming (approved):
looks good to me!

PR #316, opened 2023-07-12
add integration tests with SAS

### LINKED ISSUE #283: Explore whether we could add integration tests for `predict` by comparing with SAS `outp/outpm` options
To do:

- [x] Try out below SAS code
- [x] Try to use the `outp` or `outm` options to match `predict` results
- [x] If it works, add a few integration tests

Original issue:
**Summary**

As a statistician/programmer, I want to be able to make predictions from a fitted model using similar syntax to the outp/outpm syntax in SAS.

**Additional Information**

clarkliming:
close #283 

main changes:

1. add prediction files from SAS under `design/predict_interval` folder
2. modified `predict.cpp` to fix an error in spatial covariance structure
3. modified `predict.mmrm_tmb` to use `component(object, "beta_cov")` to use the (adjusted) covariance matrix
4. compared SAS results from outp and `h_get_prediction` (please note that, the `prediction` variance is conditional on $\theta$ for SAS, so we do the comparison with `h_get_prediction`. The interval, is different because SAS also uses a degree of freedom (in `mmrm` we use normal distribution)
5. SAS outpm is not compared (not sure how to compare)
6. observed values are not predictted in `mmrm` but predictted in `SAS`
7. vignette of predict method updated


vignettes/predict.Rmd
```diff
@@ -152,3 +152,24 @@ of $\theta$ and variance-covariance matrix of $\theta$ are provided.
 Using the generated $\theta_{new}$, we then obtain the $\beta_{new}$ and its variance-covariance matrix, with $\theta_{new}$ and
 the data used in fit. Then we simulate once with `simulate(method = "conditional", beta = beta_new, theta = theta_new)`.
 Pool all the repetition together and we get the marginal simulation results.
+
+## Comparison with `SAS`
+
+In `SAS`, from `proc mixed`, we are able to generate predictions using `outp` argument in `model` statement.
```
danielinteractive, line 158:
```suggestion
In `SAS`, from `proc mixed`, we are able to generate predictions using the `outp` argument in `the model` statement:
```

vignettes/predict.Rmd
```diff
@@ -152,3 +152,24 @@ of $\theta$ and variance-covariance matrix of $\theta$ are provided.
...
+An example will be
+
+```{sas eval=FALSE}
+PROC MIXED DATA = fev_data method=reml;
+  CLASS RACE(ref = 'Asian') AVISIT(ref = 'VIS4') SEX(ref = 'Male') ARMCD(ref = 'PBO') USUBJID;
+  MODEL FEV1 = ARMCD / ddfm=Satterthewaite solution chisq outpm=predm outp=pred;
```
danielinteractive, line 164:
```suggestion
  MODEL FEV1 = ARMCD / ddfm=Satterthewaite solution chisq outp=pred;
```
because you only mention `outp` above and I understand we did not compare with `outpm` so far?

vignettes/predict.Rmd
```diff
@@ -152,3 +152,24 @@ of $\theta$ and variance-covariance matrix of $\theta$ are provided.
 Using the generated $\theta_{new}$, we then obtain the $\beta_{new}$ and its variance-covariance matrix, with $\theta_{new}$ and
 the data used in fit. Then we simulate once with `simulate(method = "conditional", beta = beta_new, theta = theta_new)`.
 Pool all the repetition together and we get the marginal simulation results.
+
+## Comparison with `SAS`
+
+In `SAS`, from `proc mixed`, we are able to generate predictions using `outp` argument in `model` statement.
+An example will be
```
danielinteractive, line 159:
```suggestion
For example:
```

vignettes/predict.Rmd
```diff
@@ -152,3 +152,24 @@ of $\theta$ and variance-covariance matrix of $\theta$ are provided.
...
+  LSMEANS ARMCD / pdiff=all cl alpha=0.05 slice=AVISIT;
+RUN;
+```
+
+However, there are some differences between `SAS` implementations and our `mmrm` implementations, described as follows:
+
+1. While `mmrm` and `SAS` both provides prediction mean (conditional on other observations), `mmrm` do not predict those observed records.
+1. The prediction error is different in two implementations. In `SAS` the prediction error conditional on $\theta$ is provided, while
+in `mmrm` the marginal prediction error is provided.
+1. The intervals of prediction in `SAS` is using the t distribution, while in `mmrm` we use normal distribution.
```
danielinteractive, line 175:
```suggestion
However, there are some differences between the `SAS` implementation and our `mmrm` package, described as follows:

1. While `mmrm` and `SAS` both provide predicted means (conditional on other observations) for unobserved records, `SAS` also provides predicted means for observed records while `mmrm` does not. The rationale is that in the `mmrm` package we want to be consistent with the notion of predictions conditional on the observed records - which means that observed records are observed and therefore there is no prediction uncertainty anymore.
1. The prediction standard error is different between `mmrm` and `SAS`. While in `SAS` the prediction standard error is conditional on the estimated variance parameters $\theta$, in `mmrm` the marginal prediction standard error is provided. The rationale is that in the `mmrm` package we want to take into account the full uncertainty about parameter estimates including $\theta$.
1. The prediction intervals in `SAS` are based on the t distribution, while currently in `mmrm` we use the normal distribution. We will be considering an extension towards using the t distribution in the future and welcome feedback on this detail.
```

danielinteractive (commented):
thanks @clarkliming ! just some wordsmithing below

danielinteractive (approved):

PR #318, opened 2023-07-17
validate the fit object

### LINKED ISSUE #312: `h_jac_list()` to assert that there is no infinite values in returned Jacobian
To do:
- [ ] Add assertion before returning result from `C++`
- [ ] Throw meaningful error in case there are infinite values

clarkliming:
close #312


danielinteractive (commented):
Thanks @clarkliming , but this is not really addressing the issue #312 . I think these improvements are separate, and could be folded into the function https://github.com/openpharma/mmrm/blob/main/R/tmb.R#L314

clarkliming:
> Thanks @clarkliming , but this is not really addressing the issue #312 . I think these improvements are separate, and could be folded into the function https://github.com/openpharma/mmrm/blob/main/R/tmb.R#L314

updated to check in Jacobian calculations

danielinteractive (approved):
Thanks @clarkliming - I would say the inital ideas are also still valuable, could make a separate PR for those as mentioned to improve the existing helper function

PR #344, opened 2023-08-30
DOC: Add missing spatial exponential function name

ElDeveloper:
This change makes the spatial exponential title consistent with the titles for other covariance structures.


danielinteractive (approved):
Thanks @ElDeveloper !

PR #348, opened 2023-09-06
update vignettes

### LINKED ISSUE #343: Prepare 0.3 release
To do:

More spurious test failures:

- [x] Error ('test-fit.R:347:3'): mmrm falls back to other optimizers if default does not work ──
    Error in `mmrm(formula, fev_data, n_cores = 1L, optimizer_fun = c(fake = fake_optimizer,
        h_get_optimizers("L-BFGS-B")))`: Chosen optimizers 'fake, L-BFGS-B' led to problems during model fit:
    1) Model convergence problem: NEW_X.
    Consider trying multiple or different optimizers.
- [x] Failure ('test-fit.R:378:3'): mmrm works for rank deficient original design matrix by default ──
    attr(result, "converged") is not TRUE

Fix failing tests on R-devel:

- [x] Failure ('test-fit.R:227:3'): refit_multiple_optimizers works as expected with default arguments ──
     attr(result, "converged") is not TRUE
- [x] Failure ('test-fit.R:301:3'): mmrm works as expected for unstructured ───────
    attr(result, "converged") is not TRUE

Fix CRAN problems:

- [x] `Specified C++14: please drop specification unless essential` - I think we can push back on that because we looked into it already (see below)
- [x] Documented arguments not in \usage in documentation object 'h_mmrm_tmb_fit': 'data', 'weights'
- [x] Check: dependencies in R code, Result: NOTE
  Error in .make_numeric_version(x, strict, .standard_regexps()$valid_numeric_version) :
    invalid non-character version specification 'x' (type: logical) - we might look at https://github.com/R-JunmingChen/arrow/commit/2229d80f9c289b865608923af4eb5b9630035d1a to get an idea
- [x] Failure ('test-fit.R:248:3'): h_summarize_all_fits works when some list elements are try-error objects
`result` (`actual`) not equal to `expected` (`expected`).
```
    `actual$warnings[[3]]` is a character vector ('Model convergence problem: NEW_X.')
    `expected$warnings[[3]]` is NULL
      `actual$log_liks`: -1693.22493558 NA -2155.28688402
    `expected$log_liks`: -1693.22493559 NA -1693.22493812
    `actual$converged`:   TRUE FALSE FALSE
    `expected$converged`: TRUE FALSE TRUE
```
- [x] Error ('test-interop-parsnip.R:6:3'): Engine 'mmrm' is not supported for `linear_reg()`. See `show_engines('linear_reg')`.
	- something does not work here correctly (only on the Debian OS) for the registration of the `mmrm` engine...
	- trying to reproduce on r-hub: https://builder.r-hub.io/status/mmrm_0.3.0.tar.gz-a96a327d5e94fc4fe8243991ed9a0e5d
	- it might just be a follow on error after the initial package version error (above) - if the parsnip extension is not registered, then this error will follow
- [x] Error ('test-interop.R:6:3'): check_package_version warns when version is not compatible - Error in `.make_numeric_version(x, strict, .standard_regexps()$valid_numeric_version)`: invalid non-character version specification 'x' (type: logical)

Initial:
- [x] polish vignettes
	- [x] show all vignettes in drop down menu
	- [x] Update package structure vignette
	- [x] setup `sasr` 
	- [x] then rerun the comparison vignette
- [x] Get rid of note 1, because see https://stat.ethz.ch/R-manual/R-devel/doc/manual/R-exts.html#Package-structure that says we should not have hidden files or directories
- [x] We need to keep `C++14` because of #295 and https://stat.ethz.ch/R-manual/R-devel/doc/manual/R-exts.html#Using-C_002b_002b-code. 
	- [x] We could also do something like https://github.com/bstreiff/cppbits/blob/master/make_unique.h to emulate `make_unique` on older compilers but that does not seem needed (at least we did not get requests to make our package work with `C++11`)
- [x] Check on R-Hub for Windows system: https://builder.r-hub.io/status/mmrm_0.3.0.tar.gz-748e8fb621486615d52f96c204e42afb - success (just one note about the package size)
- [x] Clean NEWS
- [x] bump version

note 1:
```
* checking for hidden files and directories ... NOTE
Found the following hidden files and directories:
  vignettes/.subsections
These were most likely included in error. See section ‘Package
structure’ in the ‘Writing R Extensions’ manual.

CRAN-pack does not know about
  vignettes/.subsections
```

CRAN is asking if we can get below 5 MB for the tarball - let's try
- problem are the html files of the vignettes
- possible things to try:
  - [ ] rmarkdown::html_vignette
    - https://stackoverflow.com/questions/41237664/reduce-file-size-of-r-markdown-html-output
    - https://bookdown.org/yihui/rmarkdown/r-package-vignette.html

clarkliming:
update the cached docs


clarkliming:
@cicdguy is it possible to add an action that will execute an R script after installation of the package? It also need to update some of the data

```{sh}
Rscript vignettes/precompile.R
```

Otherwise the created doc is always using an older version of `mmrm` if we manually do this

danielinteractive:
@clarkliming Thanks a lot for rerunning!
I think it is not a big issue because we only want to run this vignette before each CRAN release. Then, we skip the version bump GHA and instead set the version manually as now in this PR. Specifically we don't want to run the vignette after each main branch update, because that is too expensive and we don't have SAS server available to run the SAS code

danielinteractive (commented):
Thanks @clarkliming can you also include the SAS PROC MIXED results? (Via sasr)

danielinteractive:
Thanks @clarkliming , will wait for the PR #349 merge

danielinteractive (approved):

PR #349, opened 2023-09-11
use glimmix instead of mixed

clarkliming:
use glimmix instead of mixed in vignettes

Will update the wording after the simulation is finished


danielinteractive (approved):

danielinteractive (approved):
Thanks @clarkliming , can you add roxygen documentation  for the new data set? Otherwise I am afraid `R CMD check` might complain

danielinteractive:
ah I just saw it in the parent PR

PR #354, opened 2023-09-28
try to work without std::make_unique [skip vbump]

danielinteractive:
[skip vbump]

CRAN team is pushing towards avoiding the need to specify the `C++14` system requirement.

So I tried to avoid the need for `std::make_unique` by emulating it with a piece of code from https://stackoverflow.com/a/64054899 (still would need to properly attribute this to the author). 
For testing I am forcing `C++11` and unfortunately I get a lot of failed tests. 

I guess it could be because here we need array support - as mentioned in https://stackoverflow.com/a/64054899 the current emulation I included does not support that, but instead we could try to take https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique#Version_1 and then modify by changing `std::enable_if_t<...>` to `std::enable_if<...>::type`... 

Quick question to @clarkliming - could we also just completely avoid the use of `make_unique` in this piece:

```c++
// Use unique pointers here to better manage resource.
if (is_spatial) {
  this->cache[r] = make_unique<D1>(theta.segment(r * theta_one_group_size, theta_one_group_size), cov_type);
} else {
  this->cache[r] = make_unique<D2>(theta.segment(r * theta_one_group_size, theta_one_group_size), n_visits, cov_type);
}
```

e.g. could we just set 
```c++
this->cache[r] = theta.segment(r * theta_one_group_size, theta_one_group_size)
```
and don't worry about the memory resources? Or would it be significantly more memory used?


danielinteractive:
Hi @clarkliming , would you maybe have time to look at my question this week? I am stuck with the CRAN release because of this and would like to know whether we can fix it this way maybe. Thanks a lot!

vignettes/package_structure.Rmd
```diff
@@ -128,7 +128,6 @@ This file implements the `P`, `Q` and `R` matrix calculations which are needed f
 ### `Makevars`
 
 This file specifies additional flags used in the source code compilation. 
-We specify here the use of `C++14` such that `std::make_unique()` is available.
 Also we ask `TMB` to use the `TMB` automatic differentiation framework (instead of the default `CppAD` framework) to be more efficient, and disable useless warnings from the `Eigen` library compilation.
```
clarkliming, line 131:
```suggestion
W ask `TMB` to use the `TMB` automatic differentiation framework (instead of the default `CppAD` framework) to be more efficient, and disable useless warnings from the `Eigen` library compilation.
```

clarkliming (approved):
looks good! only very minor comment in the vignette

PR #358, opened 2023-10-11
fix make_unique

clarkliming:
use make_shared instead of make_unique

also update the `check_package_version` function to avoid errors like

```
Cannot register mmrm for use with parsnip (v1.0.3). Version 1.0.4 to NA required
```


clarkliming:
hi @danielinteractive there is one more thing: in unit tests, the required version for parsnip is 1.1.0 where in the onload function the required version is 1.0.4. I saw you modified this in #351 but not sure which version should be the correct one.

Issue #270, opened 2023-05-18
instead of `select_matrix` and matrix calculation, use subset rows


clarkliming:
currently to obtain a subset of covariance matrix, we use `select_matrix` to obtain the subset. This involves the following calculation
1. contruct select matrix (n_subject_visits * n_visits) with filled 0
2. fill 1 for specific positions in the select matrix
3. do the matrix calculations (select_matrix %*% covariance %*% t(select_matrix)) 

instead, consider using matrix subsetting
1. construct return matrix (n_visits * n_visits), no need to initialize values
2. for the given rows, take the corresponding rows and colums

With this, we are able to harmonize the input of coordinates (they are all kind of matrix now, matrix<double> as input; then obtain the std::vector<int> subject_visits and this can be used directly in `chol_cache.R`) but this is can be resolved in another issue

related to #110

danielinteractive:
OK thanks Liming! Let's move it to the backlog for now.

danielinteractive:
Related to Eigen version 3.4 which is not available in CRAN yet:

PR #366, opened 2023-10-31
unify coordinates and visits_zero_index

### LINKED ISSUE #110: Improve handling of `visits_zero_inds` and `coordinates`
Current situation:
- for non-spatial covariance matrix, `visits_zero_inds` is used (integer vector) when subsetting the overall covariance matrix, see https://github.com/openpharma/mmrm/blob/c1e47d0e75d5b5cbe9e2fbc5df7eb811be677c1b/src/mmrm.cpp#L82
- for spatial covariance matrix, `coordinates` is used (numeric matrix), see https://github.com/openpharma/mmrm/blob/c1e47d0e75d5b5cbe9e2fbc5df7eb811be677c1b/src/mmrm.cpp#L94

The question is whether this can be improved so that we e.g. put the `visits_zero_inds` in `coordinates` with a single column or something similar.

clarkliming:
close #110

current implementation unifies coordinates and visits.

There is an extra data conversion so it is expected there is a small cost.
Average computation time for `FEV1 ~ ARMCD + us(AVISIT|USUBJID)`  on `fev_data` decrease from 37.53347 ms to 37.89274 ms (but still very small and neglectable I think)

In addition, the select matrix is removed but instead use "subset_matrix" functionality

### DIFF
--- a/src/derivatives.h
+++ b/src/derivatives.h
@@ -106,9 +106,9 @@ struct derivatives_nonspatial: public lower_chol_nonspatial<Type>, virtual deriv
       return target->second;
     } else {
       int n_visits_i = visits.size();
-      matrix<Type> ret = matrix<Type>(this->n_theta * n_visists_i, n_visists_i);
+      matrix<Type> ret = matrix<Type>(this->n_theta * n_visits_i, n_visits_i);
       for (int i = 0; i < this->n_theta; i++) {
-        ret.block(i  * n_visists_i, 0, n_visists_i, n_visists_i) = subset_matrix<matrix<Type>, vector<int>>(this->sigmad1_cache[this->full_visit].block(i  * this->n_visits, 0, this->n_visits, this->n_visits), visits, visits);
+        ret.block(i  * n_visits_i, 0, n_visits_i, n_visits_i) = subset_matrix<matrix<Type>, vector<int>>(this->sigmad1_cache[this->full_visit].block(i  * this->n_visits, 0, this->n_visits, this->n_visits), visits, visits);
       }
       this->sigmad1_cache[visits] = ret;
       return ret;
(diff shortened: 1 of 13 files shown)


danielinteractive:
Thanks @clarkliming , does this then also close https://github.com/openpharma/mmrm/issues/270 ?

src/derivatives.h
```diff
@@ -99,28 +98,17 @@ struct derivatives_nonspatial: public lower_chol_nonspatial<Type>, virtual deriv
...
-    }
   }
   // Cache and return the first order derivatives using select matrix.
   matrix<Type> get_sigma_derivative1(std::vector<int> visits, matrix<Type> dist) override {
     auto target = this->sigmad1_cache.find(visits);
      if (target != this->sigmad1_cache.end()) {
       return target->second;
     } else {
-      Eigen::SparseMatrix<Type> sel_mat = this->get_sel_mat(visits);
       int n_visists_i = visits.size();
```
danielinteractive, line 108: there seems to be a typo here
```suggestion
      int n_visits_i = visits.size();
```
and similarly rename below

src/test-utils.cpp
```diff
@@ -3,24 +3,6 @@
 
 using namespace Rcpp;
```
danielinteractive, line 5: please add unit tests for the two kinds of `subset_matrix` functions

danielinteractive (commented):
nice simplification, thanks @clarkliming !

danielinteractive (approved):
very nice, thanks!!

clarkliming:
maybe we can wait for CRAN 0.3.5 release accepted and then merge

danielinteractive:
I would merge now. this will not affect the release, it will be part of next version

PR #383, opened 2023-12-11
375: remove again staged.dependencies now that we are on CRAN

danielinteractive:
closes https://github.com/openpharma/mmrm/issues/375


clarkliming (approved):
looks good

PR #384, opened 2023-12-13
67 anova

clarkliming:
add initial ideas of type 1,2,3 hypothesis testing

### DIFF
--- a/design/anova/anova.Rmd
+++ b/design/anova/anova.Rmd
@@ -329,7 +329,6 @@ df_md(fit, l) # also close
 # Prototypes
 
 ```{r}
-
 count_levels <- function(x) {
   UseMethod("count_levels")
 }
--- a/design/anova/anova.Rmd
+++ b/design/anova/anova.Rmd
@@ -30,18 +30,20 @@ ods output Tests1 = Tests1 Tests2 = Tests2 Tests3 = Tests3;
 
 PROC MIXED DATA = fev cl method=reml;
   CLASS AVISIT(ref = 'VIS4') ARMCD(ref = 'PBO') USUBJID;
-  MODEL FEV1 = ARMCD AVISIT ARMCD*AVISIT / ddfm=satterthwaite htype=1,2,3;
+  MODEL FEV1 = ARMCD*AVISIT ARMCD AVISIT  / ddfm=satterthwaite htype=1,2,3 e3;
   REPEATED AVISIT / subject=USUBJID type=ar(1);
 RUN;
 "
 
 library(sasr)
+library(mmrm)
+library(car)
 df2sd(fev_data, "fev")
-run_sas(sas_code)
+res <- run_sas(sas_code)
 df3 <- sd2df("tests3")
 df2 <- sd2df("tests2")
 ```
-Using `anova.sas` we have the following results.
+
 Please note that type 1 result is dependent on the order of testing so it is not included here for simplicity.
 Type 2 and 3 tests are independent of the order of model parameter provided to the procedure.
 
@@ -69,7 +71,7 @@ emmeans::joint_tests(fit)
 
 however, the result do not agree well between SAS and R.
 
-Using "satterthwaite df" the F statistic is close, the d.f. is not.
+Using "satterthwaite df" the F statistic is close, the ddf is not.
 
 ## Finding suitable contrasts
 
@@ -144,7 +146,7 @@ df_md(fit, normalize_mat(dl[6:8, ]))
 df_md(fit, dl[2, ]) # d.f. do not match, stat matches a lot ()
 ```
 
-Note: the doolittle method is used to create type 1 contrasts. Not relavent here.
+Note: the doolittle method is used to create type 1 contrasts. Not really relavent here.
 
 
 # Type II Estimable Functions
@@ -241,18 +243,25 @@ df_md(fit, l)
 For the interaction, it is close to type 3 if the model is `ARMCD * AVISIT + ARMCD + AVISIT`.
 But if the model is `ARMCD + AVISIT + ARMCD * AVISIT`, the result is not so close. Why?
 
+The answer might be that the order of levels. using e3 option we can see that the contrast are different.
+However in R we will not have such issue because we always use `factor` and for the interaction we also have
+decided reference level!
+
+
+#### Changing the order of parameter
+
 ```{r}
 sas_code <- "
 ods output Tests1 = Tests1 Tests2 = Tests2 Tests3 = Tests3;
 
 PROC MIXED DATA = fev cl method=reml;
   CLASS AVISIT(ref = 'VIS4') ARMCD(ref = 'PBO') USUBJID;
-  MODEL FEV1 = ARMCD*AVISIT ARMCD AVISIT / ddfm=satterthwaite htype=1,2,3;
+  MODEL FEV1 = ARMCD*AVISIT ARMCD AVISIT / ddfm=satterthwaite htype=1,2,3 e3;
   REPEATED AVISIT / subject=USUBJID type=ar(1);
 RUN;
 "
 
-run_sas(sas_code)
+res <- run_sas(sas_code)
 df3_2 <- sd2df("tests3")
 df2_2 <- sd2df("tests2")
 ```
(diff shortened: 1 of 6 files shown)


design/anova/anova.Rmd
```diff
@@ -0,0 +1,139 @@
...
+```{r}
+library(mmrm)
+fit <- mmrm(FEV1 ~ ARMCD + AVISIT + ARMCD * AVISIT + ar1(AVISIT | USUBJID), data = fev_data)
+emmeans::joint_tests(fit)
+```
+
+
+however, the result do not agree well between SAS and R.
+
+Using "satterthwaite df" the F statistic is close, but the d.f. is not.
```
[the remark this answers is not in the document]
clarkliming: now using the contrasts constructed samely in SAS we get very very close results (even in the ddf)

design/anova/anova.Rmd
```diff
@@ -0,0 +1,139 @@
...
+ model term   df1    df2 F.ratio p.value
+ ARMCD          1 177.85  33.004  <.0001
+ AVISIT         3 309.60 116.540  <.0001
+ ARMCD:AVISIT   3 309.60   0.346  0.7921
+```
+
+## Finding suitable contrasts
+
+similar to emmeans::joint_tests() obtain a contrasts. In `emmeans` the contrasts can be obtained for Type 3, and
+use these contrasts to send to `df_md` we get exactly the same results with SAS.
```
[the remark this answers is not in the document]
clarkliming: it should be; however I did look into the so detail so just an infer from my side. there can be other differences too

clarkliming:
Hi @danielinteractive in general the design is completed (for categorical values) and should be working. with the simplest case of the mmrm fit it is now giving very close result to SAS.

design/anova/anova.Rmd
```diff
@@ -0,0 +1,394 @@
...
+zap_small <- function(x) {
+  x[abs(x) < sqrt(.Machine$double.eps)] <- 0
+  x
+}
+
+fit <- mmrm(FEV1 ~ RACE + AVISIT + RACE * AVISIT + ar1(AVISIT | USUBJID), data = fev_data)
+
+x0 <- mx[, c(1, 4:6)]
+x1 <- mx[, 2:3, drop = FALSE]
+x2 <- mx[, c(7:12)]
```
danielinteractive, line 383: this fails, I guess column index needs to be corrected?

design/anova/anova.Rmd
```diff
@@ -0,0 +1,395 @@
...
+get_l_matrix_type3 <- function(fit, var) {
+  mx <- component(fit, "x_matrix")
+  asg <- attr(mx, "assign")
+  formula <- fit$formula_parts$model_formula
+  tms <- terms(formula)
+  fcts <- attr(tms, "factors")[-1, , drop = FALSE] # discard the response
+  ods <- attr(tms, "order")
+  idx <- which(var == colnames(fcts))
+  cols <- which(asg == idx)
+  all_vars <- row.names(fcts)[which(fcts[, idx] == 1)]
```
danielinteractive, line 333: quite some duplicated code between the two functions - can we split this out in a helper function that is used by both?

design/anova/anova.Rmd
```diff
@@ -0,0 +1,395 @@
...
+get_l_matrix_type2(fit, "ARMCD:AVISIT")
+get_l_matrix_type3(fit, "ARMCD:AVISIT")
+
+# call df_md on these l matrix to obtain the results.
+
+l_mt <- get_l_matrix_type3(fit, "AVISIT")
+df_md(fit, l_mt)
+```
+
+Note: design matrix involved with numerics are still not implemented yet.
```
danielinteractive, line 370: will this be easier? can you add a paragraph to this design doc explaining the general approach for continuous variables?

design/anova/anova.Rmd
danielinteractive: I think I fixed it by adding line 380?

danielinteractive (commented):
Thanks @clarkliming , very good progress!
See a few questions below.
Also can you please add considerations regarding the user interface to these tests? (e.g. options like adding a method for `joint_tests`, ...)

design/anova/anova.Rmd
clarkliming: yes this is only a prototype that works. not necessarily be the final function.

design/anova/anova.Rmd
```diff
@@ -0,0 +1,395 @@
...
+
+With the proposed method, we have very close result to what SAS have.
+And given that we do not allow singularity in fitting (those aliased cols are removed from model fitting),
+the process can be further simplified to use identity matrix for the f1 cols.
+
+Note: Zapping might be needed to remove very small numbers.
+
+# Type III Estimable Functions
+
+Type III is constructed assuming that we have a balanced design.
```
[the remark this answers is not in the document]
clarkliming: it is recommended https://www.researchgate.net/profile/Oyvind-Langsrud/publication/220286726_ANOVA_for_unbalanced_data_Use_type_II_instead_of_type_III_sums_of_squares/links/556f3ffd08aefcb861dd622f/ANOVA-for-unbalanced-data-Use-type-II-instead-of-type-III-sums-of-squares.pdf

design/anova/anova.Rmd
```diff
@@ -0,0 +1,395 @@
...
+```
+
+### AVISIT part
+
+```{r}
+l <- matrix(0, nrow = 3, ncol = 8) # initial values
+l[, c(3, 4, 5)] <- diag(rep(1, 3))
+l[, c(6, 7, 8)] <- diag(rep(1, 3)) / 2
+
+df_md(fit, l) # also close
```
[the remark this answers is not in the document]
clarkliming: in SAS, using model  ARMCD * AVISIT ARMCD AVISIT  gives 335.368 under type 3. using model ARMCD AVISIT ARMCD * AVISIT gives a slightly different result 335.929 under type 3.

design/anova/anova.Rmd
clarkliming: the method is the same but the data handling can be different (e.g. count number of levels, count the interaction if the effect is contained)

design/anova/anova.Rmd
danielinteractive: yeah sure, just it might be helpful to already do the split up now in the design

design/anova/anova.Rmd
clarkliming: oh yes. this just store some random code for testing, should be removed later (or just store it here for reference)

clarkliming:
after adding some complex cases, it is found that

1. it should be true that the order of arguments should not change the type2/3 testing. however in SAS it is
2. when numeric values are included, the ddf do not match well in the interaction terms. why?

design/anova/anova.Rmd
clarkliming: updated now

design/anova/anova.Rmd
clarkliming: it is in the same function now

design/anova/anova.Rmd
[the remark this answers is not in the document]
clarkliming: the order changes the reference group for the interaction term. not a issue now.

design/anova/anova.Rmd
```diff
@@ -0,0 +1,395 @@
...
+```{r}
+x0 <- mx[, c(1, 2, 3, 4, 5)]
+x1 <- mx[, c(6, 7, 8), drop = FALSE]
+
+l <- matrix(0, nrow = 3, ncol = 8) # initial values
+m <- x0 %*% MASS::ginv(t(x0) %*% x0) %*% t(x0)
+l1_coef <- MASS::ginv(t(x1) %*% m %*% x1) %*% t(x1) %*% m %*% x1
+l[, c(6, 7, 8)] <- l1_coef
+
+df_md(fit, l) # close to type 3?
```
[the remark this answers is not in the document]
clarkliming: depending on the reference level, it can be very close. SAS changes its reference level automatically but if the contrast is the same the result also is very close

danielinteractive (approved):

clarkliming:
@danielinteractive could you please review again ? 
image

Issue #380, opened 2023-12-07
Convergence problem


anatsa1:
Hi,

The script to generate data for N replications is both attached as docx and in the details chunk below

<details><summary>simulation script</summary>

```r
library(MASS)
library(mvtnorm)

######
# gen.data - generates individual FEV1 using slope. FEV1 is in cc and times is measured in weeks

# Parameters for gen.data
# N - number of repetitions
# n - sample size
# mu - mean annual slope of the placebo (cc/week)
# delta - effect of the drug on slope. Effect of drug = mu + delta (cc/week)
# mua - overall mean of the intercept (cc)
# sigmaa - SD of the intercept
# sigmab - SD of the slope
# corab - covariance between slope and intercept
# sigma - residual SD
# times - the time (weeks) in which FEV1 is assessed 

gen.FEV1 = function(
  N=10, 
  n=200,
  mu=-100/52,
  delta = 50/52 ,
  mua = 2000,
  sigmaa = 300, 
  sigmab = 60, 
  corab = 0.2, 
  sigma = 10,
  times = c(0,2,6,12,24,36,52,70,88,104)
){
   nt = length(times)                    # no of repeated measures
   out = as.data.frame(list(1,1,"Tmp",1,1))
   names(out) = c("Sim","Pts","Trt","Time","FEV1")                 # out is the output

   outi = as.data.frame(list(rep(0,n*nt),rep(0,n*nt),rep("Trt",n*nt),rep(0,n*nt),rep(0,n*nt)))
   names(outi) = names(out)                                        # outi is the output for each replication

   outi[,"Pts"] = rep(1:n,each=nt)                                 
   outi[,"Trt"] = c(rep("Placebo",n*nt/2),rep("Active",n*nt/2))    
   outi[,"Time"] = rep(times,n)                                   # the time of each FEV1 measurement
   covab = corab * sigmaa * sigmab                                # cov between a and b
   COV = matrix(c(sigmaa^2,covab,covab,sigmab^2),ncol=2)          # Cov matrix for the slope and intercept 

   Pts.ID = 1:n
   Trt = c(rep("Placebo",n/2),rep("Active",n/2))
   
   for (i in 1:N)                                                 # Loop over the N simulated studies
        {outi[,"Sim"] = rep(i,n*nt)
        for (j in 1:n)                                            # Loop over the n patients
            {si = rmvnorm(1, mean = c(mua, mu + delta*(Trt[j]=="Active")),sigma=COV)   
[...]

PR #386, opened 2023-12-13
380: Idea to solve convergence problem

danielinteractive:
part of https://github.com/openpharma/mmrm/issues/380

### DIFF
--- a/design/unstructured_optimization.Rmd
+++ b/design/unstructured_optimization.Rmd
@@ -228,7 +228,7 @@ eps_hat <- residuals(lin_mod)
 
 ```{r}
 eps_hat_df <- data.frame(id = dat$Pts, visit = dat$Visit, eps = eps_hat)
-eps_hat_wide <- tidyr::pivot_wider(eps_hat_df, names_from = visit, values_from = eps) |> 
+eps_hat_wide <- tidyr::pivot_wider(eps_hat_df, names_from = visit, values_from = eps) |>
   dplyr::select(-id)
 head(eps_hat_wide)
 sigma_emp <- cov(eps_hat_wide)
--- a/design/unstructured_optimization.Rmd
+++ b/design/unstructured_optimization.Rmd
@@ -354,12 +354,21 @@ sigma_emp3[na_inds[, c(2, 1), drop = FALSE]] <- sigma_emp3[na_inds]
 
 Let's hope this is positive definite.
 
-```{r}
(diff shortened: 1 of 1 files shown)


danielinteractive:
@yonicd maybe you are interested in reviewing this design PR - I think I solved the puzzle why SAS converges much better in the example

yonicd (commented):
In general i think that the computation cost of the empirical covmat is small and would relax the constraint that this simulated data showed. 

To the question of MIVQUE0, i think that could be a longer term goal in the next release. It would give more flexibility for larger dimension designs (per [reference paper](https://support.sas.com/documentation/onlinedoc/v82/techreport_r105.pdf)) so we may hit against that problem anyways as more people use the package.

yonicd (commented):
this [blogpost](https://yetanothermathprogrammingconsultant.blogspot.com/2019/11/sdp-model-imputing-covariance-matrix.html) may give a direction on a consistent way to resolve `NA` in the covmat

here is an R [package](http://cran.nexr.com/web/packages/sdpt3r/vignettes/sdpt3r.pdf) that has a solution for the "nearsest correlation matrix" problem

yonicd (approved):

danielinteractive:
> this [blogpost](https://yetanothermathprogrammingconsultant.blogspot.com/2019/11/sdp-model-imputing-covariance-matrix.html) may give a direction on a consistent way to resolve `NA` in the covmat
> 
> here is an R [package](http://cran.nexr.com/web/packages/sdpt3r/vignettes/sdpt3r.pdf) that has similar solutions

Thanks Yoni - yeah I think we don't need those more complex solutions because we can start from the residuals data set

design/unstructured_optimization.Rmd
```diff
@@ -354,12 +354,21 @@ sigma_emp3[na_inds[, c(2, 1), drop = FALSE]] <- sigma_emp3[na_inds]
...
+```{r, eval = FALSE}
 L_mat3 <- t(chol(sigma_emp3))
 ```
 
-Hm it is not ... So maybe let's go back to pivoting the Cholesky decomposition,
-fill in NAs there and then come up with a complete matrix?
+Force the PD using Matrix package
+
+```{r}
+sigma_emp3_pd <- Matrix::nearPD(sigma_emp3)$mat
```
danielinteractive, line 364: @yonicd ah this is cool! thanks

PR #390, opened 2023-12-18
Pass formula as unnamed argument to gls()

ndunnewind:
`nlme::gls()` does not have a `formula` argument.


danielinteractive:
Thanks @ndunnewind !

PR #391, opened 2023-12-19
update anova type II and III

### LINKED ISSUE #67: Design for Type II and Type III tests a la `car::Anova()`
To do: 
- [x] Read `?car::Anova`
- [x] Go through code material:
	- [x] `glmmTMB` implementation: https://github.com/glmmTMB/glmmTMB/blob/master/glmmTMB/R/Anova.R
	- [x] `car` package code: https://github.com/cran/car/tree/master/R
- [x] Write a design doc first (so no production package code yet)
    - [x] strategy (and several options for it if it makes sense)
	- [x] prototype code

Note (from `?car::Anova?` Details): 
The designations "type-II" and "type-III" are borrowed from SAS, but the definitions used here do not correspond precisely to those employed by SAS. 
- Type-II tests are calculated according to the principle of marginality, testing each term after all others, except ignoring the term's higher-order relatives; This definition of Type-II tests corresponds to the tests produced by SAS for analysis-of-variance models, where all of the predictors are factors, but not more generally (i.e., when there are quantitative predictors).
- so-called type-III tests violate marginality, testing each term in the model after all of the others.  Be very careful in formulating the model for type-III tests, or the hypotheses tested will not make sense.

### LINKED ISSUE #392: Implement type II/III testing
Design of the method is in #67 (and related PR)

- [x] Add interface for `car::Anova`
- [x] Add unit test for all functions (including internal ones)
- [x] Add integration test to compare the SAS results
- [x] Add a vignette to illustrate the method used
- [x] #394
- [x]  Adjust style for Roxygen documentation
- [x] title style for titles
- [x] blank lines for readability between sections
- [x] parentheses for type with \cr afterwards
- [x] At least one sentence description before arguments

clarkliming:
close #392


danielinteractive:
Thanks @clarkliming , can you add a new issue for this production work? (ideally also with checkboxes for the objectives of that issue) The other issue was only for the design I think

danielinteractive (commented):
Thanks @clarkliming , looks great!
Few suggestions:
- Could we use a similar mechanism as for `emmeans` so that we don't need to import `car` but just suggest it?
- could we add a `NEWS` item
- For the vignette I remember we somewhere have something on hypothesis testing already, would it make sense to integrate those together somehow?
- please add an example in introduction vignette

clarkliming:
> Thanks @clarkliming , looks great! Few suggestions:
> 
> * Could we use a similar mechanism as for `emmeans` so that we don't need to import `car` but just suggest it?
Updated. However added a new function (copy from emmeans) but it is difficult to do any testing on it

> * could we add a `NEWS` item
Already had an entry under "new features". do we need more?

> * For the vignette I remember we somewhere have something on hypothesis testing already, would it make sense to integrate those together somehow?

Is it the section in introduction? Maybe it is clearer to be separated here because it involves a lot of details?

> * please add an example in introduction vignette

Added now

R/mmrm-methods.R
```diff
@@ -20,6 +20,10 @@
 #'    2. If the covariance structure is spatial, the covariance matrix of two time points with unit distance
 #'       will be displayed.
 #'
+#' `Anova` will return `anova` object with one row per variable and columns
+#' `Num Df`(numerator degrees of freedom), `Denom Df`(denominator degrees of freedom),
+#' `F Statistic` and `Pr(>=F)`.
+#'
```
clarkliming, line 26: I don't want to export `Anova.mmrm` so no place to properly document it. so still put it here.

R/zzz.R
```diff
@@ -15,6 +15,11 @@
     callback = parsnip_add_mmrm,
     message = emit_tidymodels_register_msg
   )
+  register_on_load(
+    "car", c("3.1.2", NA),
```
clarkliming, line 19: the version can be very flexible but let's use latest version

vignettes/hypothesis_testing.Rmd
```diff
@@ -0,0 +1,129 @@
...
+contrasts.
+
+For more details, see @goodnight1980tests.
+
+# Contained Effect
+
+Before we discuss the testing, it is important that we understand the concept of "contained effect".
+The definition of "contained effect" is:
+
+For effect "E2", it is said to contain "E1" if
```
danielinteractive, line 39: Can we rather use $E_2$ etc. that is more mathematical notation?

vignettes/hypothesis_testing.Rmd
```diff
@@ -0,0 +1,129 @@
...
+1. they both involve the same continuous variable (and no more other continuous variable is involved)
+1. "E2" have equal or more "categorical variable" than "E1"
```
danielinteractive, line 41:
```suggestion
1. "E2" have equal or more categorical covariates than "E1"
```

vignettes/hypothesis_testing.Rmd
```diff
@@ -0,0 +1,129 @@
...
+
+For more details, see @goodnight1980tests.
+
+# Contained Effect
+
+Before we discuss the testing, it is important that we understand the concept of "contained effect".
+The definition of "contained effect" is:
+
+For effect "E2", it is said to contain "E1" if
+1. they both involve the same continuous variable (and no more other continuous variable is involved)
```
danielinteractive, line 40:
```suggestion
1. they both involve the same continuous covariates
```
I guess we don't need the parenthesis? Or we need to reword to make it clearer?

vignettes/hypothesis_testing.Rmd
```diff
@@ -0,0 +1,129 @@
...
+1. "E2" have equal or more "categorical variable" than "E1"
+
+For example, for the following formula
```
danielinteractive, line 43:
```suggestion
For example, for the following model formula
```

vignettes/hypothesis_testing.Rmd
```diff
@@ -0,0 +1,129 @@
...
+
+$Y = A + B + A*B$
+
+we have
+
+1. If $A$ and $B$ are both categorical, then $F_{A*B}$ contains $F_{A}$ and $F_{B}$.
```
danielinteractive, line 49: please first define what $F_{X}$ means, otherwise this is not clear

vignettes/hypothesis_testing.Rmd
```diff
@@ -0,0 +1,129 @@
...
+
+1. All columns of $L$ associated with effect not containing $F_1$ (except $F_1$) are set to 0.
+1. The submatrix of $L$ associated with $F_1$ is set to $(X_1^\top M X_1)^{-}X_1^\top M X_1$
+1. Each of the remaining submatrices of L associated with an effect $F_2$ that contains $F_1$ is $(X_1^\top M X_1)^{-}X_1^\top M X_2$
+
+where $X_0$ stands for columns of $X$ whose associated effect do not contain $F_1$,
+$X_1$ stands for columns of $X$ associated with $F_1$,
+$X_2$ stands for columns of $X$ whose associated effect contains $F_1$,
+$M = I - X_0(X_0^\top X_0)^{-}X_0^\top$,
+and $Z^{-}$ stands for the g2 inverse of $Z$.
```
danielinteractive, line 68:
```suggestion
and $Z^{-}$ stands for the g2 inverse of $Z$.
```
please include a link to a g2 inverse definition (I don't think this is commonly known)

vignettes/hypothesis_testing.Rmd
```diff
@@ -0,0 +1,129 @@
...
+
+Note: In `mmrm` package, we do not allow singularity in general, so the g2 inverse is the inverse.
```
danielinteractive, line 70:
```suggestion
Note: Here we do not allow for singularity in general, so the g2 inverse is just the usual inverse.
```

vignettes/hypothesis_testing.Rmd
```diff
@@ -0,0 +1,129 @@
...
+Thus we can use identity matrix in step 2.
```
danielinteractive, line 71:
```suggestion
Thus we can use an identity matrix as the submatrix in step 2.
```

vignettes/hypothesis_testing.Rmd
```diff
@@ -0,0 +1,129 @@
...
+
+# Type III Hypothesis Testing
+
+For type III hypothesis testing, we assume that the hypothesis tested should be the same for all
+designs with the same general form of estimable functions.
+It can be defined with the following step:
```
danielinteractive, line 77:
```suggestion
It can be defined with the following steps:
```

vignettes/hypothesis_testing.Rmd
```diff
@@ -0,0 +1,129 @@
...
+
+where $X_0$ stands for columns of $X$ whose associated effect do not contain $F_1$,
+$X_1$ stands for columns of $X$ associated with $F_1$,
+$X_2$ stands for columns of $X$ whose associated effect contains $F_1$,
+$M = I - X_0(X_0^\top X_0)^{-}X_0^\top$,
+and $Z^{-}$ stands for the g2 inverse of $Z$.
+
+Note: In `mmrm` package, we do not allow singularity in general, so the g2 inverse is the inverse.
+Thus we can use identity matrix in step 2.
+
```
danielinteractive, line 72: can you add an example here? I think that would be helpful

vignettes/hypothesis_testing.Rmd
```diff
@@ -0,0 +1,129 @@
...
+columns equal to the number of all parameters.
+
+1. All columns of $L$ associated with effect not containing $F_1$ (except $F_1$) are set to 0.
+1. The submatrix of $L$ associated with $F_1$ is set to identity matrix.
+1. For each of the remaining submatrices of L associated with an effect $F_2$ that contains $F_1$, equate the
+values.
+
+Here equate the values means that, given that the coefficients in step 2 are all 1 (diagonal), in step 3
+this 1 is divided by the total number of new levels.
+
```
danielinteractive, line 89: also here an example will be great

vignettes/hypothesis_testing.Rmd
```diff
@@ -0,0 +1,129 @@
...
+## Special Considerations
+
+For `PROC MIXED`, it is important that the categorical values have the correct order, especially
+for structured covariance (otherwise the correlation between visits can be messed up).
+We usually use `CLASS variable(REF=)` to define the reference level.
+However, SAS will put the reference level in the last, so if the visit variable is included in fixed effect,
+there can be issues with the comparison between SAS and R, as in R we still use the first level as our reference.
+
+When we conduct the hypothesis testing, we will also face this issue.
+The result will be slightly different because of different reference groups, however, if correct contrast is used
```
danielinteractive, line 104: but is it possible always to match results by custom options for reference levels? or not?
would be good to make that clear

vignettes/hypothesis_testing.Rmd
```diff
@@ -0,0 +1,129 @@
...
+we should have closer result.
+
+### Why Model Covariate Order Changes My Testing in SAS?
+
+For type II/III testing, it is true that the order of covariate should not affect the result.
+However, if there is an interaction term, the reference level can be changed.
+With different reference level the result will be slightly different.
```
danielinteractive, line 111:
```suggestion
With different reference levels the result will be slightly different.
```

vignettes/hypothesis_testing.Rmd
```diff
@@ -0,0 +1,129 @@
...
+For `PROC MIXED`, if a model is specified as follows
+
+```
+MODEL FEV1 = ARMCD SEX ARMCD * SEX ARMCD * FEV1_BL
+```
+
+Then it is equivalent to the following `mmrm` model
+
+```
+FEV1 ~ ARMCD * SEX + ARMCD * FEV1_BL - FEV1_BL
```
danielinteractive, line 124:
```suggestion
FEV1 ~ ARMCD * SEX + ARMCD:FEV1_BL
```
would be simpler

vignettes/hypothesis_testing.Rmd
```diff
@@ -0,0 +1,129 @@
...
+```
+
+Because SAS will not include the covariate `FEV1_BL` by default, unless manually added.
+However in R we will add `FEV1_BL` by default.
+Please note that in this example we excludes the covariance structure part.
```
danielinteractive, line 129:
```suggestion
Please note that in this example we exclude the covariance structure part.
```

danielinteractive (commented):
Thank you @clarkliming ! Great stuff. I gave it a thorough last round of review.

vignettes/hypothesis_testing.Rmd
clarkliming: I think it is already clear that `if correct contrast is used we will get closer result`. (I used closer because anyway we will not have identical results)

However, we do not include the algorithm to create the correct reference.

vignettes/hypothesis_testing.Rmd
clarkliming: this is on purpose to use `ARMCD*FEV1_BL` to better illustrate the difference: in SAS using `ARMCD*FEV1_BL` will only add the interaction term, but not the two main effect.

R/interop-car.R
```diff
@@ -0,0 +1,109 @@
...
+#' @param ...Not used.
+#' @inheritParams h_get_contrast
+#'
+#' @keywords internal
+#' @noRd
+Anova.mmrm <- function(mod, type = c("II", "III", "2", "3"), tol = sqrt(.Machine$double.eps), ...) { # nolint
+  assert_double(tol, finite = TRUE, len = 1L)
+  type <- match.arg(type)
+  vars <- colnames(attr(terms(mod$formula_parts$model_formula), "factors"))
+  ret <- lapply(
```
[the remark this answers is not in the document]
clarkliming: slightly safer but causes more issues later. vapply always simplifies the result to an array/matrix however we return a list. The mode is a list and this is not we expected. We will need to further update the code to create a normal data frame

vignettes/hypothesis_testing.Rmd
danielinteractive: Hm it is not clear to me yet :-) Is the way we show the code currently getting us "closest" to SAS results? or is there another way to get even closer?
If there is an algorithm to create the correct reference/contrast then let's include it

vignettes/hypothesis_testing.Rmd
danielinteractive: can we then add a comment that in R you can do the "ARMCD:FEV1_BL" instead?

vignettes/hypothesis_testing.Rmd
clarkliming: well this is not "correct reference" but user's choices. e.g. for RACE we really don't need to say "which reference is correct, Aisan or White"? they essentially try to answer the same question, however the result can be a little different. For type II it will be a little complicated to obtain the contrast for the with different reference level, for type III it is easier, e.g. if there are three levels, using lvl1 as reference, we are testing `lvl2 - lvl1` and `lvl3 - lvl1` are both zero, and the contrast should be
```
1 0
0 1
```

however if we are using `lvl3` as reference, then we should test `lvl1 - lvl3`, `lvl2 - lvl3`. However in R we have `lvl2-lvl1`, `lvl3 - lvl1`. With a linear combination of these we can use

```
0 1
-1 -1
```
So here I use "correct contrast" to say that, e.g. in SAS we are testing `lvl1 - lvl3`, `lvl2 - lvl3`, in R we shoud not use identity matrix if we only have `lvl2 - lvl1` and `lvl3 - lvl1` and identity matrix, but use the previous form of matrix to correctly test `lvl1 - lvl3`, `lvl2 - lvl3`. It is not related to "correct reference", but "correct contrast" to reflect the same testing.

However a better solution is to use a different model to create the same design matrix.

In summary, I think we should not provide such functionality to update the contrast afterwards

danielinteractive: ok makes sense, can you add this example to the vignette? I think that helps the user

vignettes/hypothesis_testing.Rmd
clarkliming: added

vignettes/hypothesis_testing.Rmd
clarkliming: updated

_pkgdown.yml
```diff
@@ -33,18 +33,18 @@ articles:
     navbar: Covariance Structures
     contents:
       - covariance
-  - title: Degrees of Freedom
-    navbar: Degrees of Freedom
+  - title: Degrees of Freedom and Inference
+    navbar: Degrees of Freedom and Inference
```
danielinteractive, line 37:
```suggestion
  - title: Degrees of Freedom and Testing
    navbar: Degrees of Freedom and Testing
```
(inference is more estimating parameters etc)

inst/WORDLIST
```diff
@@ -1,3 +1,4 @@
+ARMCD
```
danielinteractive, line 1: maybe set in backticks to avoid the need for this?

vignettes/hypothesis_testing.Rmd
```diff
@@ -57,58 +55,207 @@ For type II hypothesis testing, the contrasts can be obtained through the follow
...
+
+$$
+\begin{matrix} 
+\mu & ARMCD & RACE & RACE & ARMCD:RACE & ARMCD:RACE\\
+0 & 0 & 0 & 0 & 0 & 0\\ 
+0 & 0 & 0 & 0 & 0 & 0\\ 
+\end{matrix}
+$$
+
+Please note that this is $2\times 6$ matrix, the rows is the number of coefficients that is associated with `RACE`,
```
danielinteractive, line 90:
```suggestion
Please note that this is a $2\times 6$ matrix, the rows is the number of coefficients that is associated with `RACE`,
```

vignettes/hypothesis_testing.Rmd
```diff
@@ -57,58 +55,207 @@ For type II hypothesis testing, the contrasts can be obtained through the follow
...
+and the columns is the total number of coefficients `r length(coef(fit))`
```
danielinteractive, line 91:
```suggestion
and the columns is the total number of coefficients `r length(coef(fit))`.
```

vignettes/hypothesis_testing.Rmd
```diff
@@ -57,58 +55,207 @@ For type II hypothesis testing, the contrasts can be obtained through the follow
...
+
+Then following step 2, we filled in identity matrix in the corresponding submatrix
```
danielinteractive, line 93:
```suggestion
Then following step 2, we filled in the identity matrix in the corresponding submatrix
```

vignettes/hypothesis_testing.Rmd
```diff
@@ -57,58 +55,207 @@ For type II hypothesis testing, the contrasts can be obtained through the follow
...
+
+$$
+\begin{matrix} 
+\mu & ARMCD & RACE & RACE & ARMCD:RACE & ARMCD:RACE\\
+0 & 0 & 0 & 0 & 0 & 0\\ 
+0 & 0 & 0 & 0 & 0 & 0\\ 
+\end{matrix}
+$$
+
+Then following step 2, we filled in identity matrix in the corresponding submatrix
```
danielinteractive, line 151:
```suggestion
Then following step 2, we filled in the identity matrix in the corresponding submatrix
```

vignettes/hypothesis_testing.Rmd
```diff
@@ -57,58 +55,207 @@ For type II hypothesis testing, the contrasts can be obtained through the follow
...
+0 & 0 & 1\\
+\end{matrix}
+$$
+
+But given that the reference level is different, we are testing different things.
+In SAS we are testing `VIS1 - VIS4`, `VIS2 - VIS4` and `VIS3 - VIS4`,
+while in R we are testing `VIS2 - VIS1`, `VIS3 - VIS1` and `VIS4 - VIS1`.
+
+To correctly test `VIS1 - VIS4`, `VIS2 - VIS4` and `VIS3 - VIS4` in R, we need to update the contrast
+to the following matrix(ignoring other part) to get closer result.
```
danielinteractive, line 243:
```suggestion
to the following matrix (ignoring the other part) to get a closer result.
```

danielinteractive (approved):
Congrats @clarkliming , good to merge after last few suggestions are in!

inst/WORDLIST
clarkliming: it is in the matrix part so adding backtick is ugly and really not necessary


```math
\begin{bmatrix}
\mu & ARMCD & RACE & RACE & ARMCD:RACE & ARMCD:RACE\\
0 & 0 & 1 & 0 & 0 & 0\\ 
0 & 0 & 0 & 1 & 0 & 0\\ 
\end{bmatrix}
```

```math
\begin{bmatrix}
\mu & `ARMCD` & `RACE` & `RACE` & `ARMCD`:`RACE` & `ARMCD`:`RACE`\\
0 & 0 & 1 & 0 & 0 & 0\\ 
0 & 0 & 0 & 1 & 0 & 0\\ 
\end{bmatrix}
```

PR #401, opened 2024-01-04
use data instead of full_frame in emp

### LINKED ISSUE #400: transformation in response is not allowed
it is natural and allowed in other linear models that the transformation applied to response can work, e.g.

```
lm(log(FEV1) ~ ARMCD + FEV1_BL, data = fev_data)
```

however mmrm has some issues related to it

```
mmrm(log(FEV1) ~ ARMCD + FEV1_BL + ar1(AVISIT|USUBJID), data = fev_data)
```
works

however in the vignette of #396 

```
mmrm(log(FEV1) ~ ARMCD + FEV1_BL + ar1(AVISIT|USUBJID), data = fev_data, start = emp_start)
```
do not work and gives error like 

Error in fit_single_optimizer(tmb_data = tmb_data, formula_parts = formula_parts,  : 
  Assertion on 'all.vars(model_formula)' failed: Must be a subset of {'log(fev1)','trt','time','pts','(weights)'}, but has additional elements {'fev1'}.

in #396 we may break something and we should fix them

- [x] add unit test to cover transformation in response
- [x] fix issue in #396 branch

clarkliming:
close #400

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -403,6 +403,10 @@ std_start <- function(cov_type, n_visits, n_groups, ...) {
 #' to obtain the empirical variance-covariance, and it is then used to obtain the
 #' starting values.
 #'
+#' @note `data` is used instead of `full_frame` because `full_frame` is already
+#' transformed if model contains tranformations, e.g. `log(FEV1) ~ exp(FEV1_BL)` will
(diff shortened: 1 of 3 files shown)


danielinteractive (approved):
Thanks @clarkliming , can you add a `@note` or so to explain why `data` is used instead of `full_frame`? (for posterity)

PR #417, opened 2024-01-23
update confint

### LINKED ISSUE #414: Add a `confint` method to use t-distribution based quantiles
Currently we implicitly use Gaussian distributed based quantiles. Which does not make really sense when we are making so much effort to derive adjusted d.f. for the t-statistics. 

In particular, this can lead to inconsistencies, where the p-value could be larger than 0.05 while the confidence interval would exclude 0.

Example:
```r
library(dplyr)
library(mmrm)

df <- fev_data |> 
  filter(AVISIT %in% c("VIS1", "VIS2"), USUBJID %in% paste0("PT", 1:50)) |> 
  droplevels()

mm2 <- mmrm(data = df, formula = log(FEV1) ~ AVISIT + us(AVISIT | USUBJID))
mm2_sum <- summary(mm2)
mm2_sum

# This is Gaussian based:
confint(mm2)
# We want actually here:
mm2_coefs <- as.data.frame(mm2_sum$coefficients) |> 
  mutate(t_quantile = stats::qt(0.975, df = df)) |> 
  mutate(
    ci_lower = Estimate - t_quantile * `Std. Error`,
    ci_upper = Estimate + t_quantile * `Std. Error`
  )
mm2_coefs
```

Hopefully we don't need to make this as complex as https://github.com/glmmTMB/glmmTMB/blob/master/glmmTMB/R/methods.R#L865 in our case.

Relatedly, we also need to test whether `emmeans` confidence intervals are correctly calculated following this update.

clarkliming:
close #414 

add confint for the coefficients

update news, update vignettes to include `confint` for `pairs` on emmeans object.


NEWS.md
```diff
@@ -5,6 +5,7 @@
 - `Anova` is implemented for `mmrm` models and available upon loading the `car` package. It supports type II and III hypothesis testing.
 - The argument `start` for `mmrm_control()` is updated to allow function/numeric input for better
 choices of initial values.
+- `confint` on `mmrm` models will give t-based confidence interval now, instead of the normal approximation.
```
danielinteractive, line 8:
```suggestion
- `confint` on `mmrm` models will give t-based confidence intervals now, instead of the normal approximation.
```

R/mmrm-methods.R
```diff
@@ -227,3 +231,32 @@ print.summary.mmrm <- function(x,
   cat("\n")
   invisible(x)
 }
+
+
+#' @describeIn mmrm_methods obtain the confidence interval for the MMRM fit results.
```
danielinteractive, line 236:
```suggestion
#' @describeIn mmrm_methods obtain the confidence interval for the coefficients.
```

R/mmrm-methods.R
```diff
@@ -227,3 +231,32 @@ print.summary.mmrm <- function(x,
...
+  if (is.numeric(parm)) parm <- pnames[parm]
+  assert_number(level, lower = 0, upper = 1)
+  a <- (1 - level) / 2
+  pct <- paste(format(100 * c(a, 1 - a), trim = TRUE, scientific = FALSE, digits = 3), "%")
+  coef_table <- h_coef_table(object)
+  df <- coef_table[parm, "df"]
+  ses <- coef_table[parm, "Std. Error"]
+  fac <- stats::qt(a, df = df)
+  ci <- array(NA, dim = c(length(parm), 2L), dimnames = list(parm, pct))
+  ci[] <- cf[parm] + (ses * fac) %o% c(1, -1)
```
danielinteractive, line 260: let's avoid `%o%` here, and keep with basic operators

tests/testthat/test-mmrm-methods.R
```diff
@@ -103,3 +103,40 @@ test_that("print.summary.mmrm works as expected for spatial fits", {
...
+    confint(object, parm = 100),
+    "Element 1 is not <= 11"
+  )
+  expect_error(
+    confint(object, parm = c("a", "b")),
+    "Must be a subset of"
+  )
+})
+
+test_that("confint give same result as emmeans if no ", {
```
danielinteractive, line 134:
```suggestion
test_that("confint give same result as emmeans if no ", {
```
if no ..? is there a word missing?

danielinteractive, line 134:
```suggestion
test_that("confint give same result as emmeans if no ", {
```
if no ..? is there a word missing?

danielinteractive (commented):
Thanks @clarkliming , nice work - can you please also add an integration test comparing with SAS results? (maybe addition to existing test)

R/mmrm-methods.R
clarkliming: updated; also use "confidence intervals" instead.

tests/testthat/test-mmrm-methods.R
clarkliming: oh sorry. updated now.

R/mmrm-methods.R
clarkliming: `%o%` is also from base, only a shortcut for `outer`. (similar to `%*%`). but we can change that to simpler forms as this is only two cols

clarkliming:
> Thanks @clarkliming , nice work - can you please also add an integration test comparing with SAS results? (maybe addition to existing test)

Hi @danielinteractive I added some tests, but did not cover all the scenarios, should be sufficient as illustration purpose?

tests/testthat/test-mmrm-methods.R
```diff
@@ -140,3 +140,39 @@ test_that("confint give same result as emmeans if no ", {
   expect_equal(conf$lower.CL, conf_coef[, 1])
   expect_equal(conf$upper.CL, conf_coef[, 2])
 })
+
+test_that("confint give same result as SAS on unstructured", {
+  object <- mmrm(FEV1 ~ ARMCD + SEX + us(AVISIT | USUBJID), data = fev_data, reml = FALSE)
```
danielinteractive, line 145: would be good to have a comment in here and below test to refer to which design/SAS script was used / which output file the numbers come from

danielinteractive (approved):
super, thanks Liming - just one last comment, then good to merge

danielinteractive (approved):

Issue #423, opened 2024-01-31
L-BFGS-B divergence however error code is not correct


clarkliming:
when fitting mmrm we can run into such issue
```
Warning message:
In mmrm(FEV1 ~ AVISIT + us(AVISIT | USUBJID), data = fev) :
  Divergence with optimizer L-BFGS-B due to problems: L-BFGS-B needs finite values of 'fn'
```

and immediately
```
Error in UseMethod("logLik") : 
  no applicable method for 'logLik' applied to an object of class "c('mmrm_fit', 'list')"
```

The reason is that `L-BFGS-B` final status is warning instead of error.

reprex.

```
fev = data.table::data.table(fev_data)

fev[seq(1, 800, 4), FEV1 := 0]

mmrm(FEV1 ~ AVISIT + us(AVISIT|USUBJID), data = fev)

```

clarkliming:
related to #311 but I wonder why for these divergent values we still want to keep them as "success"? Is there any cases that this error is reported however a result is also returned? @danielinteractive

danielinteractive:
Thanks @clarkliming , maybe there was a change in the optim here, but yeah in principle this should be an error I guess. Maybe just need to adapt the filter for warnings in the helper functions?

PR #424, opened 2024-02-01
fix issue of h_record_all_outputs

clarkliming:
I can not produce examples that `h_record_all_outputs` that can create a "try-error" object, but only returns `list()` on any error, unless there is some internal `try(expr, silent=T)` in `fit_mmrm` call, however I did not find such things that could potentially return the "try-error" object. any ideas @danielinteractive ?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,7 +6,7 @@
 
 ### Bug Fixes
 
-- Previously if an optimizer fails `h_record_all_output` will return a `list()` or `try-error`. Now only `try-error` will be returned.
+- Previously if a secondary optimizer fails `mmrm` will fail. This is fixed now.
 
 # mmrm 0.3.9
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,10 @@
 
 - Fix internal test skipping functions for MacOS R. 
 
+### Bug Fixes
+
+- Previously if an optimizer fails `h_record_all_output` will return a `list()` or `try-error`. Now only `try-error` will be returned.
+
 # mmrm 0.3.9
 
 ### Miscellaneous
--- a/R/fit.R
+++ b/R/fit.R
@@ -441,15 +441,15 @@ mmrm <- function(formula,
     drop_visit_levels = control$drop_visit_levels,
     allow_na_response = FALSE
   )
-  fit <- list()
+  fit <- structure("", class = "try-error")
   names_all_optimizers <- names(control$optimizers)
-  while ((length(fit) == 0) && length(control$optimizers) > 0) {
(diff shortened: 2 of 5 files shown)


danielinteractive:
> I can not produce examples that `h_record_all_outputs` that can create a "try-error" object, but only returns `list()` on any error, unless there is some internal `try(expr, silent=T)` in `fit_mmrm` call, however I did not find such things that could potentially return the "try-error" object. any ideas @danielinteractive ?

I think that is as it is supposed to work - this helper function `h_record_all_outputs` should always return a `list` and not fail with an error?

NEWS.md
```diff
@@ -4,6 +4,10 @@
 
 - Fix internal test skipping functions for MacOS R. 
 
+### Bug Fixes
+
+- Previously if an optimizer fails `h_record_all_output` will return a `list()` or `try-error`. Now only `try-error` will be returned.
```
danielinteractive, line 9: In principle correct, but for the NEWS let's keep it at the user level, and describe which problem for the user is now fixed

R/fit.R
```diff
@@ -441,15 +441,15 @@ mmrm <- function(formula,
     drop_visit_levels = control$drop_visit_levels,
     allow_na_response = FALSE
   )
-  fit <- list()
+  fit <- structure("", class = "try-error")
```
danielinteractive, line 444: are we sure we want to change this here?

clarkliming: image

from this part it has to be "try-error" to consistent, or change this function to use length as error indicator

clarkliming:
> > I can not produce examples that `h_record_all_outputs` that can create a "try-error" object, but only returns `list()` on any error, unless there is some internal `try(expr, silent=T)` in `fit_mmrm` call, however I did not find such things that could potentially return the "try-error" object. any ideas @danielinteractive ?
> 
> I think that is as it is supposed to work - this helper function `h_record_all_outputs` should always return a `list` and not fail with an error?

maybe some time ago there is some try-catcher in the fit_mmrm so that it will return a "try-error"? or we can update the error checks to "length(result) == 0"

R/fit.R
clarkliming: (and this will affect more docs and unit tests, just because previously it is all taken as "try-error")

danielinteractive: ok sounds good

danielinteractive:
@clarkliming hm not sure if anything else to be done here - as long as the new unit test passes, this should close #423 ? Or would you like to add more tests around the `try-error` behavior and start fixing afterwards?

danielinteractive (approved):

PR #438, opened 2024-04-22
add conditional arg for predict

### LINKED ISSUE #437: support for `{marginaleffects}`
see https://github.com/vincentarelbundock/marginaleffects/pull/1000

clarkliming:
close #437 

allow unconditioanl prediction


tests/testthat/test-tmb-methods.R
```diff
@@ -173,6 +173,19 @@ test_that("predict will return NA if data contains NA in covariates", {
   )
 })
 
+test_that("predict can give unconditional predictions", {
```
danielinteractive, line 176: thanks Liming, can we also add in addition
- one test to check that it works with formula-based response variable
- one test to see that we can modify the prediction result when changing the saved coefficients (similar as it is needed for the interface to the {marginaleffects} package)

R/tmb-methods.R
```diff
@@ -610,20 +619,23 @@ simulate.mmrm_tmb <- function(object,
...
-    data = newdata,
-    include = c("subject_var", "visit_var", "group_var", "response_var"),
-    na.action = "na.pass"
-  )
+  all_vars <- all.vars(object$formula_parts$formula)
+  assert_names(colnames(newdata), must.include = all_vars)
+  full_frame <- object$tmb_data$full_frame
+  for (v in setdiff(all_vars, object$formula_parts$subject_var)) {
+    if (is.factor(full_frame[[v]]) || is.character(full_frame[[v]])) {
+      newdata[[v]] <- h_factor_ref(newdata[[v]], full_frame[[v]])
```
danielinteractive, line 627: this is duplicate from the other method, let's factor it out

R/tmb-methods.R
```diff
@@ -610,15 +615,14 @@ simulate.mmrm_tmb <- function(object,
...
   # Ensure new data has the same levels as original data.
-  full_frame <- model.frame(
-    object,
-    data = newdata,
-    include = c("subject_var", "visit_var", "group_var", "response_var"),
-    na.action = "na.pass"
+  newdata <- h_factor_ref_data(
+    newdata,
+    object$tmb_data$full_frame,
+    setdiff(all.vars(object$formula_parts$formula[[3]]), object$formula_parts$subject_var)
```
danielinteractive, line 621: this line also seems duplicate, would it maybe be better to also handle this in the helper function? or another small one?

R/tmb.R
```diff
@@ -135,34 +135,29 @@ h_mmrm_tmb_data <- function(formula_parts,
...
+      na.action = "na.pass"
     ))
   )
   if (drop_levels) {
     full_frame <- droplevels(full_frame, except = formula_parts$visit_var)
   }
-  # If `y` is allowed to be NA, replace it with original y.
   if (allow_na_response) {
-    full_frame[[temp_var]] <- y_original[full_frame[[temp_var]]]
+    keep_ind <- complete.cases(full_frame[, colnames(full_frame) != vn])
```
danielinteractive, line 156: do we need a `drop = FALSE` here to be safe?

R/utils.R
```diff
@@ -351,6 +351,16 @@ h_factor_ref_data <- function(x, ref, vars) {
   x
 }
 
+#' Obtain Right Hand Side Variables of `mmrm_tmb_formula_parts` Object
+#'
+#' @param object (`mmrm_tmb_formula_parts`)\cr object.
+#'
+#' @keywords internal
+h_mmrm_vars <- function(object) {
```
danielinteractive, line 359: nice, wondering if this could be directly a component of `mmrm_tmb_formula_parts` upon its construction? then we could just take it from there

danielinteractive (commented):
Thanks @clarkliming , last comments, I promise :-)

R/tmb.R
clarkliming: complete.cases works for both data.frame or vectors, so should be fine here without drop

R/utils.R
```diff
@@ -327,6 +327,29 @@ h_factor_ref <- function(x, ref, var_name = vname(x)) {
   factor(x, levels = h_default_value(levels(ref), sort(uni_ref)))
 }
 
+#' Convert Character to Factor Following Reference `MMRM` Fit.
+#'
+#' @param object (`mmrm_tmb`)\cr the fitted MMRM object.
+#' @param data (`data.frame`)\cr input data.
+#'
```
danielinteractive, line 334: Return is missing

clarkliming: added now; also added for some other functions in utils.R

PR #456, opened 2024-07-26
clean up simulations

clarkliming:
there are just too many updates done within this merge request, @danielinteractive please have a careful review.

The reason to have this is to achieve smooth simulation. The changes (and reasons) are listed as below

1. totally remove simChef usage.
    1. current structure of using simChef is not efficient, because all model objects are saved. with large dataset, a tmb model can be over 1GB size. it is impossible to store all these models without crashing.
    2. we don't really need the simChef simulation framework. simpler simulations easier to configure
1. remove sasr usage.
    1. sasr with ssh tunnels is not stable enough for large simulations
1. fix convergence criteria.
    1. previously for mmrm it only check if there is an error but not the [convergence code](https://github.com/openpharma/mmrm/blob/b0741f8c6f25edfa799ce122c987dd2f7b820ece/simulations/missing-data-benchmarks/R/method/mmrm-wrapper.R#L77)
1. ML and REML are both included in the simulation
    1. previously glmmTMB is using ML instead of REML, but for other simulations they all use REML
1. missing data pattern is updated
    1. previously only in extreme missing scenario the missing is [related to treatment](https://github.com/openpharma/mmrm/blob/b0741f8c6f25edfa799ce122c987dd2f7b820ece/simulations/missing-data-benchmarks/R/dgp/helpers.R#L112). to have more consistency the treatment is not affecting the missing pattern now
1. simulation for nlme::gls is updated
    1. previously for nlme::gls, the covariance structure is [not correct](https://github.com/openpharma/mmrm/blob/b0741f8c6f25edfa799ce122c987dd2f7b820ece/simulations/missing-data-benchmarks/R/method/nlme-wrapper.R#L49). it should be ` ~ visit|id`
    2. the toeplitz covariance structure is also added
1. simulation data updated
    1. now for the same configuration but only different covariance structure in data generation, all the other covariates are the same. the only different thing is the error term
    2. now the toeplitz covariance structure has the same variance ([toep](https://github.com/openpharma/mmrm/blob/b0741f8c6f25edfa799ce122c987dd2f7b820ece/simulations/missing-data-benchmarks/R/meal.R#L24), [us](https://github.com/openpharma/mmrm/blob/b0741f8c6f25edfa799ce122c987dd2f7b820ece/simulations/missing-data-benchmarks/R/dgp/helpers.R#L52)
1. improve the covariance structure
    1. previously for gls and sas, the covariance structure for the first subject is returned. now it is re-constructed and no matter if there is anyone with full visits, a full covariance structure will be returned.


danielinteractive:
Apologies @clarkliming for the delayed review, will do this tomorrow

danielinteractive (approved):
Looks so much simpler. Thanks a lot @clarkliming !!

PR #463, opened 2024-09-03
make unconditional prediction default

### LINKED ISSUE #461: `predict()` should not require response variable to be present in `newdata`
Reprex:

```r
library("mmrm")

fit <- mmrm(
  formula = FEV1 ~ RACE + SEX + ARMCD * AVISIT + us(AVISIT | USUBJID),
  data = fev_data
)

predict(fit, subset(fev_data, select = - FEV1))
```

fails with:

```
Error in eval(predvars, data, env) : object 'FEV1' not found
```

because `h_mmrm_tmb_data()` is invoked and this uses the `full_formula` which includes the response variable from the fit.

Reported originally in https://github.com/vincentarelbundock/marginaleffects/pull/1000#issuecomment-2282352569

clarkliming:
close #461 

for simulate we can open another issue


danielinteractive (commented):
Thanks a lot @clarkliming !

danielinteractive:
we should change the approach to "X * beta" for unconditional, because then marginal effects package will also work better with duplicate subject index and visit index

clarkliming:
to achieve that, we first need to call `model.matrix`. however it seems that there is a subset https://github.com/openpharma/mmrm/blob/17e417b8b783d2bf8b74fbe3f50650471b4f332b/R/tmb-methods.R#L346 in model.matrix. Do you know why there is @danielinteractive

danielinteractive:
Good question, seems like from the comment that we want to be sure to only include the utilized observations. What happens if you delete the subset? Does it fail any tests?

clarkliming:
> Good question, seems like from the comment that we want to be sure to only include the utilized observations. What happens if you delete the subset? Does it fail any tests?

of course it fails some tests related to `model.matrix`, but as you know there are not really much downstream part of model.matrix so hard to say what is the impact.

I can imagine that we should follow

1. if no data provided, always return the utilized obs in that model
2. if new data provided, we should always return based on new data; if NA is in response we should ignore it?
3. however we have the choices to select different "include" option; whether we want to make them consistent (in rows of observations) given different "include"?
4. do we really need the "include" option for model.matrix? (do we really want other dummy columns added to the model.matrix,  like subject var, visit_var and group_var? why they should be?)

danielinteractive:
@ddsjoberg Do you have any opinions on this, because I vaguely remember you added this method? 😃

ddsjoberg:
> @ddsjoberg Do you have any opinions on this, because I vaguely remember you added this method? 😃

I also vaguely recall adding this method 😜 I think when I added this, I had in mind its use with the modeling object. It may be the case that predictions/new data are not being handled they way they are meant to be (I am no expert in the expected behaviour of `model.matrix()` in various cases).

If there is an update needed, can we ensure the update doesn't break the integration with {broom.helpers}, which does tasks like identifying reference rows and more.

image

Should we simply aim to mimic the behaviour of `model.matrix.glmmTMB()` ?

ddsjoberg:
If you're not in a rush, I can investigate in the next one to two weeks and make the update to model.matrix()

danielinteractive:
Thanks a lot @ddsjoberg , that is good to know. We are kind of in a rush because we want to ship another CRAN release soon, and this would be great to include in there to unblock the downstream package `marginaleffects`. 
@clarkliming maybe we could add the `broom.helpers` example as additional test under https://github.com/openpharma/mmrm/blob/main/tests/testthat/test-tmb-methods.R#L571 (adding the corresponding package in `Suggests`), and then try to refactor?

NEWS.md
```diff
@@ -6,11 +6,14 @@
...
 - Previously `car::Anova` will fail if the model does not contain intercept. This is fixed now.
 - Previously, `mmrm` will ignore contrasts defined for covariates in the input data set. This is fixed now.
+- Previously, `predict` will always require the response to be valid, even for unconditional predictions. This is fixed now and unconditional prediction do not require the response to be valid anymore.
 - When running with `TMB` package versions below 1.9.15, MMRM fit results are not completely reproducible. While this may not be relevant for most applications, because the numerical differences are very small, we now issue a warning to the user if this is the case. We advise users to upgrade their `TMB` package versions to 1.9.15 or higher to ensure reproducibility.
 
 ### Miscellaneous
 
 - Upon fitting an MMRM, it is checked whether a not reproducible optimization feature of `TMB` is turned on. If so, a warning is issued to the user once per session.
+- `model.matrix` is updated to ensure that the `NA` values are dropped.
+- `model.frame` is updated to ensure that the `na.action` works.
```
danielinteractive, line 16: can merge these two items

R/utils.R
```diff
@@ -314,6 +314,18 @@ h_warn_na_action <- function() {
   }
 }
 
+#' Obtain na.action as Function
```
danielinteractive, line 317:
```suggestion
#' Obtain `na.action` as Function
```

tests/testthat/test-tmb-methods.R
```diff
@@ -378,11 +386,6 @@ test_that("h_construct_model_frame_inputs works with all columns", {
     FEV1 ~ RACE + USUBJID + AVISIT,
     ignore_attr = TRUE
   )
-  expect_equal(
```
danielinteractive, line 381: why is this deleted? does it not work anymore?

tests/testthat/test-tmb-methods.R
```diff
@@ -580,62 +573,46 @@ test_that("model.matrix works as expected with defaults", {
...
   out_frame <- expect_silent(
-    model.matrix(fit1, data = fev_data, include = "visit_var")
+    model.matrix(fit1, data = fev_data)
   )
-  expect_identical(colnames(out_frame), c("(Intercept)", "ARMCDTRT", "AVISITVIS2", "AVISITVIS3", "AVISITVIS4"))
+  expect_identical(colnames(out_frame), c("(Intercept)", "ARMCDTRT"))
+})
+
+test_that("model.matrix works with broom.helpers", {
+  fit <- get_mmrm()
```
danielinteractive, line 613: maybe make this test conditional on `broom.helpers` being installed?

NEWS.md
danielinteractive: I think we need a separate news item that the interface of `model.matrix` changed, instead of `include` there is now `drop_response`

R/utils.R
```diff
@@ -314,6 +314,18 @@ h_warn_na_action <- function() {
...
+#' @keywords internal
+h_get_na_action <- function(na_action) {
```
danielinteractive, line 319: can you add 1-2 unit tests for this helper?

tests/testthat/test-tmb-methods.R
```diff
@@ -580,62 +573,46 @@ test_that("model.matrix works as expected with defaults", {
   )
 })
 
-test_that("model.matrix works as expected with includes", {
-  object <- get_mmrm_tmb()
```
danielinteractive, line 584: can we include these tests still as much as possible with the new interface? and also test the `drop_response` thing?

danielinteractive (commented):
thanks a lot @clarkliming !
one additional thing which would be good to test - is it now possible to have stacked duplicate observations as input for predict and it will return the correct unconditional predictions? because that was what was blocking `marginaleffects`

NEWS.md
clarkliming: these two are different; model.frame can use `na.action` however, `model.matrix` always drop NA in data.

tests/testthat/test-tmb-methods.R
clarkliming: `h_construct_model_frame_inputs` does not need that many attributes anymore thus removed

NEWS.md
```diff
@@ -12,8 +12,9 @@
 ### Miscellaneous
 
 - Upon fitting an MMRM, it is checked whether a not reproducible optimization feature of `TMB` is turned on. If so, a warning is issued to the user once per session.
-- `model.matrix` is updated to ensure that the `NA` values are dropped.
+- `model.matrix` is updated to ensure that the `NA` values are dropped. Additionally, an argument `use_response` is added to decide whether records with `NA` values in the response should be dicarded.
 - `model.frame` is updated to ensure that the `na.action` works.
+- `predict` is updated to allow duplicated subject id for unconditional prediction(fixed effect only).
```
danielinteractive, line 17:
```suggestion
- `predict` is updated to allow duplicated subject IDs for unconditional prediction.
```
we anyway only have fixed effects so I would omit that

danielinteractive (approved):

danielinteractive:
ok this seems ok apart from the spurious url check problem

PR #507, opened 2025-05-23
506: optimize empirical vcov / Satterthwaite d.f. calculations

### LINKED ISSUE #506: Improving performance of `mmrm(vcov = "Empirical")`
## The issue

`mmrm(vcov = "Empirical")` can be slow because the C++ function `get_empirical()` performs the cross product of a potentially large matrix (`g`) on itself: https://github.com/openpharma/mmrm/blob/7d27878b1772b305b8e63ad809c2b8a7e8f7538d/src/empirical.cpp#L84

`mmrm()` includes `gtvg` as an element of its return value, named `empirical_df_mat`, but it is not needed to obtain the primary fitted model. Nonetheless, its calculation can cause `mmrm()` to go from taking a few seconds to several minutes or hours to complete. 

The function `mmrm:::h_quad_form_mat()` is the only `mmrm` function designed to accept `empirical_df_mat` (as its `center` argument), but `mmrm()` does not necessarily need to run it.
https://github.com/openpharma/mmrm/blob/7d27878b1772b305b8e63ad809c2b8a7e8f7538d/R/satterthwaite.R#L126 https://github.com/openpharma/mmrm/blob/7d27878b1772b305b8e63ad809c2b8a7e8f7538d/R/satterthwaite.R#L245

## Proposed changes

1. `get_empirical()` will not create `gtvg` and will return `g` in its place.
2. Add an `apply_crossproduct` argument to `mmrm:::h_quad_form_mat()` wherein the user can indicate whether or not `center` shall be (in effect) preprocessed with `crossprod()` (which is equivalent to `t(center) %*% center`). The default will be `FALSE`, so any objects outside the scope of this issue will be unaffected.
3. Calls to `mmrm:::h_quad_form_mat()` that expect `empirical_df_mat` will have the argument `apply_crossproduct = TRUE` added.

danielinteractive:
closes #506 

This PR optimizes empirical covariance estimation and Satterthwaite degrees-of-freedom calculations by replacing the large `df_mat` (crossproduct of `g`) with a more efficient `g_mat`, introducing a new helper, updating docs, and adding tests.

- Return `g_mat` instead of `df_mat` in `get_empirical` and R wrappers, reducing memory/time overhead.
- Add `h_df_1d_sat_empirical()` to compute Satterthwaite d.f. from `g_mat` and wire it into existing routines.
- Update documentation, vignette, NEWS, and add tests for the new empirical d.f. helper.
- Add a new author, Nikolas Krieger, to the `DESCRIPTION` file. (`DESCRIPTION`)


R/satterthwaite.R
```diff
@@ -98,6 +97,46 @@ h_gradient <- function(jac_list, contrast) {
...
+      "mmrm fit was obtained with package version < 0.3.15, ",
+      "using deprecated calculation of d.f., consider refitting the model"
+    )
+    h_quad_form_mat(contrast_matrix, object$empirical_df_mat)
+  } else if (!is.null(object$empirical_g_mat)) {
+    contrast_times_g_transposed <- tcrossprod(
+      object$empirical_g_mat,
+      contrast_matrix
+    )
+    crossprod(contrast_times_g_transposed)
```
NikKrieger, line 131: Aha, seeing this is a moment of integration for me. I understand now what you've done here---doing this operation in only two efficient steps:
```
crossprod(tcrossprod(g, C)) =
    t(tcrossprod(g, C)) %*% tcrossprod(g, C) =
    t(g %*% t(C)) %*% g %*% t(C) =
    t(t(C)) %*% t(g) %*% g %*% t(C) =
    C %*% crossprod(g) %*% t(C)
```

danielinteractive: Yeah that's it!
Actually my variable name was misleading, fixed it now.

NikKrieger:
Review complete.

PR #519, opened 2025-07-03
Adding anova.mmrm

NikKrieger:
# Adding `anova.mmrm()`

This begins with adding the design document specifying how `anova.mmrm()` should be added to the package.


design/anova/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,157 @@
...
+
+### Plain language
+
+#### Special case: supplying only one model fit
+
+If the user provides only one model fit to `anova.mmrm()`, the function will calculate and return the significance of the model terms.
+
+#### Comparing multiple model fits
+
+If the user provides more than one model fit, a `tibble` of results is produced. This `tibble` will have a row for each model, and the columns will contain identifying information as well as standard diagnostics (e.g., AIC, log likelihood, etc.).
```
danielinteractive, line 78: is a `tibble` result standard, e.g. for `nlme` and `lme4` or `glmmTMB`? I would try to keep close to those reference packages in terms of UI (as you have described above already) but also in terms of return values and behavior

design/anova/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,157 @@
...
+anova.mmrm <- function (object, ..., test = TRUE, refit = FALSE) {
+  
+  # If there are objects supplied to "..."
+  if (...length()) {
+    
+    # Gather all objects into one list
+    fits <- list(object, ...)
+    
+    # Ensure all the objects are mmrm fits
+    validate_mmrm_objects(fits)
```
danielinteractive, line 94: side remark: use existing helper and assertion functions as much as possible, avoid creating too many new ones if not needed

design/anova/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,157 @@
...
+    
+    # If only one model fit was provided, calculate the significance of model
+    # terms, which is what ?stats::anova promises
+    out <- calculate_significance_of_model_terms(object)
+  }
+  
+  out
+}
+```
+
```
danielinteractive, line 154: Please start with prototype code here (just functions, without roxygen2 chunks or tests, but with text explanations to illustrate the flow etc.), please see other design docs as examples

design/anova/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,157 @@
...
+# Examples
```
danielinteractive, line 155: yes, great to have 3-4 typical examples here (single model passed, multiple models with same cov structure, multiple models with same formula but different cov structure, errors when this is not the case.)

design/anova/adding_anova_to_mmrm_design_document.md
```diff
@@ -0,0 +1,181 @@
...
+          
+          # Refit all models using harmonized data
+          fits <- lapply(fits, refit_mmrm, harmonized_data)
+        } else {
+          stop("Data underlying mmrm fits are not all the same and refit = FALSE")
+        }
+      }
+      
+      
+      results <- results |>
```
danielinteractive, line 155: other side remark: please avoid pipe operations in mmrm package code where possible, and use base functions if possible, we don't have this in our https://github.com/openpharma/mmrm/blob/main/.github/CONTRIBUTING.md but is my personal preference 😄 
is more relevant for the production PRs later, but easier if you start prototyping below like that

danielinteractive (commented):
Nice start, thanks a lot @NikKrieger ! Please see initial direction pointers below

design/anova/adding_anova_to_mmrm_design_document.md
```diff
@@ -0,0 +1,181 @@
+# Adding anova() to mmrm
```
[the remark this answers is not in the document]
NikKrieger: Done

design/anova/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,157 @@
...
+
+### Providing only one fit to `anova.mmrm()`
+
+`?stats::anova` asserts that if only one model fit is provided, the function will calculate the significance of model terms. The `anova.mmrm()` should follow this lead and provide these results accordingly.
+
+### Following the lead of `anova()` methods in `nlme` and `lme4`
+
+In order that `anova.mmrm()` may be comparable to `anova()` methods in the `nlme` and `lme4` packages, it shall contain the following arguments:
+
+-   `test` argument to opt in/out of obtaining LRT results (see `?nlme::anova.lme`)
```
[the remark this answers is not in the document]
NikKrieger: `anova.mmrm()` could still provide a table of other results. For instance, `nlme::anova.lme` provides degrees of freedom, AIC, BIC, and log likelihood. 

If this is not desired we can do away with the `test` argument and always include LRT results.

design/anova/adding_anova_to_mmrm_design_document.qmd
NikKrieger: The `anova()` methods in `nlme` and `lme4` supply data frames. `mmrm` already imports `tibble` so I leveraged `tibble::tibble()`. Let me know if I should use a `data.frame` instead.

I modeled the `anova.mmrm()` output off `nlme::anova.lme`.
```
    Model df      AIC      BIC    logLik   Test   L.Ratio p-value
fm1     1  6 454.6367 470.6173 -221.3183                         
fm2     2  5 453.3146 466.6318 -221.6573 1 vs 2 0.6778943  0.4103
```

Here's what `lme4::anova.merMod` yields:
```
Data: sleepstudy
Models:
m1: Reaction ~ Days + (Days | Subject)
m2: Reaction ~ predict(splines::ns(Days, df = 2)) + (Days | Subject)
   npar    AIC    BIC  logLik -2*log(L)  Chisq Df Pr(>Chisq)
m1    6 1763.9 1783.1 -875.97    1751.9                     
m2    7 1764.5 1786.9 -875.25    1750.5 1.4313  1     0.2315
```

danielinteractive: Yeah I would say let's keep to a simple `data.frame` in the similar way as `nlme::anova.lme`

design/anova/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,157 @@
...
+| `test` | Logical indicating whether likelihood ratio tests should be used to compare the fitted models represented by `object` and the object(s) in `...`. Defaults to `TRUE`. |
+| `refit` | Logical indicating if the models should be refitted before comparing. Defaults to `FALSE`. |
+
+## Function outline
+
+### Plain language
+
+#### Special case: supplying only one model fit
+
+If the user provides only one model fit to `anova.mmrm()`, the function will calculate and return the significance of the model terms.
```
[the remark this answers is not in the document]
NikKrieger: Can we do this iteratively by running `df_1d()` once for each model term and compiling the results into a data frame?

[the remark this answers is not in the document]
danielinteractive: I think it should be LRT, for consistency.

So you need to refit the model dropping each term (not coefficient) in turn

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,579 @@
...
+### Prototype code
+
+#### `anova.mmrm()`
+
+```{r }
+anova.mmrm <- function (object, ..., test = TRUE, refit = FALSE) {
+  checkmate::assert_class(object, "mmrm")
+  
+  # If there are objects supplied to "..."
+  if (...length()) {
```
danielinteractive, line 151: interesting that this works!
But still better to keep it readable:
```suggestion
  extra_args <- list(...)
  if (length(extra_args)) {
```

danielinteractive: and adapt accordingly below

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,579 @@
...
+      if (!refit && !h_fits_data_identical(fits, refit = FALSE)) {
+        stop("Not all data sets underlying the model fits are identical and ",
+             "refit = FALSE.", call. = FALSE)
+      }
+      
+      # Label for the pair being compared (e.g., "3 vs 4")
+      out$test <- c(NA, paste(head(out$Model, -1), "vs", tail(out$Model, -1)))
+      
+      # Calculate the likelihood ratio between the pairs using base::diff()
+      out$likelihood_ratio <- c(NA, diff(out$logLik))
```
danielinteractive, line 194:
```suggestion
      # Calculate the log of likelihood ratios between the pairs using base::diff()
      out$log_likelihood_ratio <- c(NA, diff(out$logLik))
```

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,579 @@
...
+      
+      # Calculate the p-value using the ratios and differences in df
+      out$p_value <-
+        pchisq(
+          2 * abs(out$likelihood_ratio),
```
danielinteractive, line 199:
```suggestion
          2 * abs(out$log_likelihood_ratio),
```

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,579 @@
...
+          df = abs(diff(out$df)), 
+          lower.tail = FALSE
+        )
+    }
+    
+  } else { 
+    
+    # attributes of terms.object for the model formula without the cov struct
```
danielinteractive, line 207: the term wise testing probably makes sense to factor out in a small dedicated helper function

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,579 @@
...
+#### Supporting functions:
+
+```{r}
+# Compares the data sets underlying a list of fits. 
+
+# If refit = FALSE, this is a predicate function indicating whether all the data
+# sets are the same.
+
+# If refit = TRUE, all data sets are harmonized, the models are refit using the
+# harmonized data set, and the refits are returned.
```
danielinteractive, line 245: It adds complexity that this function returns different things depending on one argument. Can you refactor such that we have the overall flow:

1) `h_check_same_data()` - whether the data sets are the same or not
2) if not the same and not refit -> error
3) if not the same and refit ->
 a) `h_smallest_common_data()` returns smallest common data set included in all fits (and not the largest data set e.g.)
b) `h_refit_on_data()` refits all models on this smallest common data set
4) if the same -> all good

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,579 @@
...
+      if (
+        h_check_cov_struct_nesting(model_basic, model_augmented) == "identical"
+      ) {
+        warning("Models have identical covariates and covariance structures.")
+      },
+    "nested" = 
+      if (
+        h_check_cov_struct_nesting(model_basic, model_augmented) == "nested"
+      ) {
+        stop("Models cannot have both nested covariates and nested covariance structures.",
```
danielinteractive, line 370: this should be allowed now that I changed my mind, see above

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,579 @@
...
+    envir <- environment(fit[["formula_parts"]][["full_formula"]])
+    # Evaluate the model call in the environment of the original fit's formula
+    fit <- eval(call, envir)
+  }
+
+  fit
+}
+
+
+# Compares two models to see if they are properly nested.
```
danielinteractive, line 330: btw we also need to think about the use of the REML vs. ML fits. probably easiest to just assert that always the same method was used for the fit. I would not bother to cater for the unlikely case that sometimes REML and sometimes ML was used and then refit, that is too rare and makes things more complex

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,579 @@
...
+editor: visual
+---
+
+## Setup
+
+```{r setup}
+library(mmrm)
+library(tidyverse)
+set.seed(20250708)
+data <-
```
danielinteractive, line 14: let's go with a simpler example that does not refer to the other tutorial (which is not available to others here)

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,579 @@
...
+  cols <- sort(colnames(df))
+  Reduce(function(df, col) df[order(df[[col]]), ], rev(cols), init = df[cols])
+}
+```
+
+# Examples
+
+Using a modified version of the `fit_reduced` and `fit_full` models in the tutorial:
+
+```{r}
```
danielinteractive, line 452: currently this example shows a case where the df are the same for the two models (note that this won't be the case anymore after correcting the d.f. definition above). but it is interesting because in that case the models cannot be nested. That could also be a simple first step in the nesting function above : for nesting, the d.f. of the first model must be strictly smaller than the d.f. of the second model.

design/adding_anova_to_mmrm_design_document.qmd
NikKrieger: `...length()` is a base function that returns the number of arguments supplied to `...`. But if it's too obscure then we can certainly grab the elements with `list(...)` as you suggest.

design/adding_anova_to_mmrm_design_document.qmd
NikKrieger: Yes, I agree. It dawned on me that there is no upside to using the tutorial's example in this repo

design/adding_anova_to_mmrm_design_document.qmd
NikKrieger: To reference the description of the `refit` argument found in `?lme4::anova.merMod` again:

> logical indicating if objects of class lmerMod should be refitted with ML before comparing models. The default is TRUE to prevent the common mistake of inappropriately comparing REML-fitted models with different fixed effects, whose likelihoods are not directly comparable.

It seems that the original purpose of the `refit` argument in `lme4:::anova.merMod()` was to allow the refitting of REML models. 

Can we make it so that `refit = TRUE` gives the function permission to refit models with the same data AND to also refit any REML models?

```
if (refit) {

  if (different_data_sets(fit)) {
    fits <- refit_all_with_new_data_and_ML(fits)
  } else if (any_REML_models(fit)) {
    fits <- refit_REML_models_with_ML(fits)

} else {
  error if not all data same
  error if any REML present
}
```

design/adding_anova_to_mmrm_design_document.qmd
danielinteractive: Yes please, it is more common

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,579 @@
...
+  nall <- component(object, "n_obs")
+
+  # was REML used (ML was used if FALSE)
+  reml_used <- component(object, "reml")
+
+  # Number of estimated coefficients (excluding aliased coefficients)
+  n_fixed_effects <- length(component(object, "beta_est"))
+
+  # Adjusted number of observations when REML is used
+  nobs <- nall - reml_used * n_fixed_effects
```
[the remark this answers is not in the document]
danielinteractive: Hm then maybe let's drop those

design/adding_anova_to_mmrm_design_document.qmd
danielinteractive: We don't necessarily need to match lme4 behavior here. In MMRMs so far I have only ever seen REML been used in practice. Therefore I would prefer to focus only on datasets when refitting

design/anova/adding_anova_to_mmrm_design_document.qmd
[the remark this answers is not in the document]
NikKrieger: We have agreed to use `df_md()` at this point.

design/anova/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,157 @@
...
+```
+
+### Arguments
+
+|  |  |
+|----|----|
+| `object` | An object inheriting from class `mmrm`: a fitted mixed model with repeated measures. |
+| `...` | Other optional fitted model objects inheriting from class `mmrm`. |
+| `test` | Logical indicating whether likelihood ratio tests should be used to compare the fitted models represented by `object` and the object(s) in `...`. Defaults to `TRUE`. |
+| `refit` | Logical indicating if the models should be refitted before comparing. Defaults to `FALSE`. |
```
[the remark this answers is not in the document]
NikKrieger: Outstanding questions concerning `REML`:

1. Can REML models with different fixed effects be compared via LRT?
2. If the user supplies an `ML` and `REML` model along with `refit=TRUE`, should we refit one or the other? Or just throw an error? Or should we attempt the correction to the log likelihood, as in https://github.com/cran/nlme/blob/cf4770a47fdb2ebfe94a3e75eaaa12c7cbb036b6/R/lme.R#L1415-L1422 ?

design/adding_anova_to_mmrm_design_document.qmd
NikKrieger: I have made it so that `refit` only applies when the data sets are different.

Just checking—does the assertion in `?lme4::anova.merMod` apply here (viz., you can't compare REML models with different fixed effects)? I suppose the question behind my question is whether models with nested covariates have "different" fixed effects.

design/adding_anova_to_mmrm_design_document.qmd
NikKrieger: Done

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,579 @@
...
+
+The title of the issue is centered on incorporating the Likelihood Ratio Test for `mmrm` fits: "\[feature request\] add likelihood ratio test."
+
+There is consensus that this goal of obtaining LRT results would be available via a new `anova()` method for `mmrm` model fits.
+
+### Models should be compared to check if they are nested:
+
+-   Nested sets of covariates. See: <https://github.com/openpharma/mmrm/issues/164#issuecomment-1446364890>
+-   "Nested" covariance structures (e.g., unstructured + any structured; AR1 being a special case of Toeplitz). See: <https://github.com/openpharma/mmrm/issues/164#issuecomment-1446627674> and the **Nested covariance structures proofs** appendix below.
+-   **Not both** of the above, as this would be too complicated.
```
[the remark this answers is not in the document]
NikKrieger: Done

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,579 @@
...
+  reml_used <- component(object, "reml")
+
+  # Number of estimated coefficients (excluding aliased coefficients)
+  n_fixed_effects <- length(component(object, "beta_est"))
+
+  # Adjusted number of observations when REML is used
+  nobs <- nall - reml_used * n_fixed_effects
+
+  # Number of parameters
+  df <- component(object, "n_theta")
```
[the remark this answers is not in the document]
NikKrieger: I believe this is fixed. However, I did not add the "fixed sigma patch" because it didn't seem to me that `mmrm()` has this option. Let me know if I'm mistaken.

design/adding_anova_to_mmrm_design_document.qmd
NikKrieger: Once `logLik.mmrm_tmb()` correctly calculates and reports `df` (which is hopefully now), is the current nesting assertion (covariates and covariance structure) sufficient?

design/anova/adding_anova_to_mmrm_design_document.qmd
NikKrieger: Done

design/anova/adding_anova_to_mmrm_design_document.qmd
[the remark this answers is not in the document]
danielinteractive: Good point about comparing models via LRT when estimating with REML: Indeed it seems we can only then compare nested covariance models, but not different covariates. 

See e.g. " This entails that when comparing multiple models by their REML likelihoods,
they must contain the same fixed effects, and that REML, by itself, does not estimate the fixed effects. A
benefit of this is that the number of parameters in the model is restricted" in [here](https://psoerensen.github.io/qgnotes/REML.pdf)

So that means for us then: When REML is used, we assert identical covariates and check for nested covariance model. Whereas if ML is used (for all models), we check nested covariates and nested covariance model.

And the d.f. is only the number of variance parameters for REML, and the sum of coefficients and variance parameters for ML

design/adding_anova_to_mmrm_design_document.qmd
danielinteractive: Yes, thanks indeed it applies here, see my comment above

design/anova/adding_anova_to_mmrm_design_document.qmd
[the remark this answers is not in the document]
NikKrieger: I have implemented this. If any of the models use REML, all the models must have the same covariates (but they can have a nested covariance structure). 

I have tweaked `logLik.mrmm_tmb()` so that it adds the number of coefficients to `df` if the model uses ML instead of REML.

design/adding_anova_to_mmrm_design_document.qmd
NikKrieger: Changed accordingly; use comment thread above for further discussion.

design/adding_anova_to_mmrm_design_document.qmd
NikKrieger: I have now done so. When LRT is requested (i.e., `test = TRUE`), degrees of freedom are now compared before anything else.

I did notice the edge case in which there are only 1 or 2 timepoints (i.e., $m=1$ or $m=2$), two models with identical covariates and nested covariance structures can have the same number of degrees of freedom. For instance, when $m=2$ and using `toeph`:

$$
N_{params} = m(m+1)/2 = 2(2+1)/2 = 3
$$ 

And if using `us`, then:

$$
N_{params} = 2m - 1 = 2\cdot 2 - 1 = 3
$$

This will result in the two models having the same degrees of freedom even though they are nested but not identical. Should these two models indeed not be allowed to be compared via LRT?

danielinteractive: Yes, correct, in that case they should not be allowed to be compared via LRT. So LRT comparison is valid if 1) covariates are nested, 2) covariance is nested, and 3) at least one d.f. difference.

design/anova/adding_anova_to_mmrm_design_document.qmd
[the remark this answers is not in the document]
danielinteractive: perfect, thanks Nik!

design/adding_anova_to_mmrm_design_document.qmd
NikKrieger: Cool, then the current code accomplishes this.

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,855 @@
...
+## Key points from the GitHub issue
+
+The title of the issue is centered on incorporating the Likelihood Ratio Test for `mmrm` fits: "\[feature request\] add likelihood ratio test."
+
+There is consensus that this goal of obtaining LRT results would be available via a new `anova()` method for `mmrm` model fits.
+
+### Models should be compared to check if they are nested:
+
+-   Nested sets of covariates. See: <https://github.com/openpharma/mmrm/issues/164#issuecomment-1446364890>
+-   "Nested" covariance structures (e.g., unstructured + any structured; AR1 being a special case of Toeplitz). See: <https://github.com/openpharma/mmrm/issues/164#issuecomment-1446627674> and the **Nested covariance structures proofs** appendix below.
```
danielinteractive, line 27: for REML or ML

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,855 @@
...
+-   **Not both** of the above, as this would be too complicated.
```
danielinteractive, line 28:
```suggestion
-   Both of the above, but only for ML estimated models
```

danielinteractive (commented):
Thanks Nik, see a few minor comments below for ML/REML etc.

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,855 @@
...
+
+## Key points from the GitHub issue
+
+The title of the issue is centered on incorporating the Likelihood Ratio Test for `mmrm` fits: "\[feature request\] add likelihood ratio test."
+
+There is consensus that this goal of obtaining LRT results would be available via a new `anova()` method for `mmrm` model fits.
+
+### Models should be compared to check if they are nested:
+
+-   Nested sets of covariates. See: <https://github.com/openpharma/mmrm/issues/164#issuecomment-1446364890>
```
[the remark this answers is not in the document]
NikKrieger: This is in the section that distills key points from the GitHub issue, which doesn't mention REML nor ML. Should I still include this here? Or should I add some language into the **Additional considerations** section that adds nuance?

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,855 @@
...
+-   `test` argument to opt in/out of obtaining LRT results (see `?nlme::anova.lme`). The function will return degrees of freedom, AIC, BIC, and log-likelihood regardless.
+-   `refit` arguments to allow for refitting the models (see `?lme4::anova.merMod`).
+
+# Prerequisite: augment `logLik.mmrm_tmb()`
+
+This function needs to return more data in the `attributes()` of the returned value:
+
+-   `df`: the degrees of freedom shall be returned as an attribute.
+    -   If REML estimation was used, it shall equal the number of parameters.
+    -   If ML estimation was used, it shall equal the sum of the number of parameters and the number of estimated coefficients.
```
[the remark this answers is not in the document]
NikKrieger: Done.

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -0,0 +1,855 @@
...
+
+  # REML
+  if (component(object, "reml")) {
+    df <- n_params
+  } else {
+    # ML
+    # Number of estimated coefficients
+    n_fixed_effects <-
+      length(
+        component(object, if (complete) "beta_est_complete" else "beta_est")
```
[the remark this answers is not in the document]
danielinteractive: Hm yeah I think that is the case. Just use `coef(complete = FALSE)` then

design/adding_anova_to_mmrm_design_document.qmd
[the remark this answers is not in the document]
danielinteractive: Yeah additional considerations is also fine. The issue was just a starting point. We have evolved from there already 😄

design/adding_anova_to_mmrm_design_document.qmd
danielinteractive: let's not worry about the issue details anymore - now that we started the work, things have changed

NikKrieger: Done. Ready for you to take another look.

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -144,22 +153,36 @@ anova.mmrm <- function (object, ..., test = TRUE, refit = FALSE) {
     
     # If the data are not all the same and refit = TRUE, refit all with the 
     # largest common data set
-    if (refit && !h_check_same_data(fits)) {
-      fits <- lapply(fits, h_refit_mmrm, data = h_fits_common_data(fits))
+    if (refit && !h_check_fits_all_data_same(fits)) {
+      common_data <- h_fits_common_data(fits)
+      refit_lgl <-
```
danielinteractive, line 158:
```suggestion
      needs_refit <- 
```

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -144,22 +153,36 @@ anova.mmrm <- function (object, ..., test = TRUE, refit = FALSE) {
...
+    if (refit && !h_check_fits_all_data_same(fits)) {
+      common_data <- h_fits_common_data(fits)
+      refit_lgl <- 
+        !vapply(
+          lapply(fits, h_get_minimal_fit_data), 
+          h_check_columns_nested,
+          FUN.VALUE = logical(1L),
+          data_augmented = common_data
+        )
+      fits[refit_lgl] <-
```
danielinteractive, line 165:
```suggestion
      fits[need_refit] <-
```

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -144,22 +153,36 @@ anova.mmrm <- function (object, ..., test = TRUE, refit = FALSE) {
...
+        lapply(fits[refit_lgl], h_refit_mmrm, data = common_data)
+    } else {
+      refit_lgl <- rep_len(FALSE, length(fits))
```
danielinteractive, line 168:
```suggestion
      need_refit <- rep_len(FALSE, length(fits))
```

design/adding_anova_to_mmrm_design_document.qmd
```diff
@@ -144,22 +153,36 @@ anova.mmrm <- function (object, ..., test = TRUE, refit = FALSE) {
...
     
     # vector of log likelihood results for the fits
-    LL_vec <- lapply(fits, logLik.mmrm_tmb)
+    logLik_vec <- lapply(fits, logLik.mmrm_tmb)
     
     # Calculate the standard diagnostics.
     out <-
       data.frame(
         Model = seq_along(fits),
+        refit = refit_lgl,
```
danielinteractive, line 178:
```suggestion
        refit = need_refit,
```

danielinteractive (approved):
Thanks Nik, just a minor naming suggestion below, otherwise good to merge this design doc and then start production PRs

design/adding_anova_to_mmrm_design_document.qmd
NikKrieger: done

design/adding_anova_to_mmrm_design_document.qmd
NikKrieger: done

PR #527, opened 2025-08-11
525: make `predict` work when response is missing

### LINKED ISSUE #525: `predict` does not work for unconditional prediction if `interval = "prediction"` and response does not exist
`predict` works for unconditional prediction if response does not exist - but only if interval is not `prediction`. When that is the case, it still does not work.

To do:

- [x] add test
- [x] fix

danielinteractive:
closes #525


NEWS.md
```diff
@@ -3,6 +3,7 @@
 ### Bug Fixes
 
 - Previously, the compound symmetry correlation models (`cs` and `csh`) could have divergence issues, especially when the correlation was negative. This was because the range of the correlation parameter was from -1 to 1, whereas the correct range is between -1/(m-1) and 1, where m is the number of observations per subject. This is fixed now by using another mapping from the real line to the correct range, and therefore compound symmetry correlation models will work more robustly. 
+- Previously, the `emp_start` starting values could only work with variables included in `data`. Now they can also work when variables are not included in `data`.
```
clarkliming, line 6: it sounds a bit strange, that the issue itself describes that the prediction is not working if response is not included, indeed this issue arise from the fact that we want to build the tmb_data to ensure everything is consistent (including the factor levels etc). however why the issue goes to `emp_start`? because `emp_start` use a linear regression and then use the residuals to obtain the covariance matrix shape. It is indeed strange that it is still working when the response is not available, how would we provide this "empirical start values"?

clarkliming: by the way, in terms of "variables" included in the `data`, does it mean

1. response is `log(y)` however data contains only `y`
2. response is `y` however data does not contain any `y`
3. independent variable include `x` while data does not contain any `x`
4. independent variable include `log(x)` while data does contain `x`

R/tmb-methods.R
```diff
@@ -80,6 +82,10 @@ predict.mmrm_tmb <- function(object,
   assert_flag(conditional)
   interval <- match.arg(interval)
   formula_parts <- object$formula_parts
+  # Make sure we have at least a completely NA filled response variable in newdata.
+  if (isFALSE(formula_parts$response_var %in% names(newdata))) {
```
clarkliming, line 86: whether `response_var` can be a chracter vector? if it is a vector, assiging values may not work (and initially I think adding these variables should already work to predict the prediction intervals)

NEWS.md
```diff
@@ -3,6 +3,8 @@
 ### Bug Fixes
 
 - Previously, the compound symmetry correlation models (`cs` and `csh`) could have divergence issues, especially when the correlation was negative. This was because the range of the correlation parameter was from -1 to 1, whereas the correct range is between -1/(m-1) and 1, where m is the number of observations per subject. This is fixed now by using another mapping from the real line to the correct range, and therefore compound symmetry correlation models will work more robustly. 
+- Previously, the `predict()` method failed when requesting an unconditional prediction interval. This is fixed now.
+- Previously, the `emp_start()` starting values could only work with variables included directly in `data`. Now they can also work when variables are not included in `data` but are available in the parent environment: The variables will be correctly included in the design matrix. This is achieved by an internal update to `h_mmrm_tmb_data()`.
```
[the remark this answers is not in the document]
clarkliming: oh it is variable outside `data.frame`. indeed for that issue we skip the  emp_start function line 3, remove the assertion should work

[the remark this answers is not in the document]
danielinteractive: I will check whether there was an initial problem, or whether it was just a follow up thing to correct when I fixed the design matrix creation, will take a few days though

danielinteractive:
Thanks @clarkliming for having a look already. I would be happy with a simpler solution, but please note what I mentioned that also for me initially this looked like a minor thing, but then I ended up working long time on it... 
If you have a good idea also feel free to create a new branch of course.

R/tmb-methods.R
clarkliming: since this is adding much complexity now, I suggest we do the following:

1. add an argument "conditional" to indicate that the prediction is conditional on the current observations
2. in the prediction, if not "conditional" we use `delete.response` to remove the response from terms, and use that terms to re-construct the frame
3. if "conditional" we then create the `h_mmrm_tmb` object

now we don't need to try if the model frame can be construted

danielinteractive: @clarkliming please have a look at my latest commit which accounts also for this situation. I think it works ok.

clarkliming: indeed it solves most of the issue but on rare cases like

```{r}
a = function(){parent.frame()$FEV1}
mmrm(a() ~ AVISIT + ar1(AVISIT|USUBJID),data = fev_data)
```

then the obtained response variable is length 0; this could still lead to issues if FEV1 is not available in the data. but we can proceed with this solution


note: glm works in this scenario because it totally removes the response

```{r}
fit <- glm(a() ~ AVISIT, data = fev_data)

fev2 <- fev_data
fev2$FEV1 <- NULL
predict(fit, newdata = fev2)
```

PR #528, opened 2025-08-12
Add h_refit_mmrm() and h_generate_new_name()

NikKrieger:
Adds the helper functions `h_refit_mmrm()` and `h_generate_new_name()` along with documentation and tests.


tests/testthat/test-tmb-methods.R
```diff
@@ -1379,3 +1379,23 @@ test_that("h_assert_lrt_suitability() ensures suitability for LRT testing", {
                                        dfs = c(attr(logLik(get_mmrm_cs()), "df"), attr(logLik(get_mmrm()), "df")),
                                        is_reml = c(component(get_mmrm_cs(), "reml"), component(get_mmrm(), "reml"))))
 })
+
+
+test_that("h_generate_new_name() generates a string without a binding in env", {
```
danielinteractive, line 1384: maybe here could be worth to have one or two more tests where you explicitly create an environment within another environment and see that the while loop strategy works

danielinteractive (commented):
Thanks @NikKrieger , looks good, just one comment below

tests/testthat/test-tmb-methods.R
NikKrieger: Yes, I agree that it's prudent to make the function run the `while` loop more than once. I've implemented this.

PR #534, opened 2025-08-26
fix nesting checks for interaction terms

NikKrieger:
Fix the variable nesting check for `anova.mmrm()` so that it will properly recognize when an interaction term is nested inside of another interaction term.


danielinteractive:
Thanks @NikKrieger for explaining, cool - in that case I think it would still be valuable to get this improved code. Can you maybe add tests that show where the new method works better (if there is such a case) compared to the old way of checking?

NikKrieger:
The covariate nesting check now accounts for interaction terms whose components were specified in a different order. Tests have also been added.

danielinteractive:
Thanks @NikKrieger , let's first merge the Anova() PR and then continue with this one

danielinteractive (approved):

PR #549, opened 2025-11-20
548: Fix `model.frame()` behavior for missing values

### LINKED ISSUE #548: `model.frame.mmrm_tmb()` applies `na.action` too early
Line 328 in `mmrm:::model.frame.mmrm_tmb()` applies the user-specified `na.action` too early, not allowing `model.frame.default()` to properly apply the `na.action`: https://github.com/openpharma/mmrm/blob/832ee0c698dca3952417697e12ea014dfab6c228/R/tmb-methods.R#L325-L331

This can cause errors if an environment variable is present:
``` r
library(mmrm)
env_FEV1_BL <- fev_data$FEV1_BL
fit <-
  mmrm(
    FEV1 ~ env_FEV1_BL + SEX + ar1(as.ordered(VISITN) | USUBJID),
    fev_data
  )
model.frame(fit, data = fev_data)
#> Error in model.frame.default(formula = lst_formula_and_data$formula, data = h_get_na_action(na.action)(lst_formula_and_data$data), : variable lengths differ (found for 'env_FEV1_BL')

# This works fine because no environment variable is involved:
fit_no_env_var <-
  mmrm(
    FEV1 ~ FEV1_BL + SEX + ar1(as.ordered(VISITN) | USUBJID),
    fev_data
  )
head(model.frame(fit_no_env_var, data = fev_data))
#>        FEV1  FEV1_BL    SEX USUBJID as.ordered(VISITN)
#> 2  39.97105 25.27144 Female     PT1                  2
#> 4  20.48379 25.27144 Female     PT1                  4
#> 6  31.45522 45.02477   Male     PT2                  2
#> 7  36.87889 45.02477   Male     PT2                  3
#> 8  48.80809 45.02477   Male     PT2                  4
#> 10 35.98699 43.50070 Female     PT3                  2
```
<sup>Created on 2025-11-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

The error occurs because `data`, which is preprocessed with `na.omit()` before being passed to `model.frame.default()`, has fewer observations than the environment variable `env_FEV1_BL`. `model.frame.default()` doesn't tolerate this. 

Unless there is a good reason to apply `na.action` to `data` before sending it on to `model.frame.default()`, line 328 should just be: 
``` r
data = lst_formula_and_data$data,
```

`model.frame.default()` has an `na.action` argument, so just let it do its thing. 

If this change is made, the only test that is violated is this one:
https://github.com/openpharma/mmrm/blob/832ee0c698dca3952417697e12ea014dfab6c228/tests/testthat/test-tmb-methods.R#L695-L707

I don't totally understand the purpose of this test. All it seems to be doing is checking that `USUBJID` gets subjected to `na.action` in view of the response variable `FEV1` (which has missing values), even though the user is explicitly leaving `response_var` out of the `include` argument of `model.frame.mmrm_tmb()`. (Maybe there's a good reason for this behavior and this test sheds light on why `data` is preprocessed with `na.action`, but I'm not understanding.)

``` r
devtools::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.5.1 (2025-06-13)
#>  os       Ubuntu 24.04.3 LTS
#>  system   x86_64, linux-gnu
#>  ui       X11
#>  language (EN)
#>  collate  C.UTF-8
#>  ctype    C.UTF-8
#>  tz       America/New_York
#>  date     2025-11-19
#>  pandoc   3.6.3 @ /usr/lib/rstudio/resources/app/bin/quarto/bin/tools/x86_64/ (via rmarkdown)
#>  quarto   1.7.32 @ /usr/lib/rstudio/resources/app/bin/quarto/bin/quarto
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version     date (UTC) lib source
#>  backports     1.5.0       2024-05-23 [1] CRAN (R 4.5.1)
#>  cachem        1.1.0       2024-05-16 [1] CRAN (R 4.5.1)
#>  checkmate     2.3.3       2025-08-18 [1] CRAN (R 4.5.1)
#>  cli           3.6.5       2025-04-23 [1] CRAN (R 4.5.1)
#>  devtools      2.4.6       2025-10-03 [1] CRAN (R 4.5.1)
#>  digest        0.6.37      2024-08-19 [1] CRAN (R 4.5.1)
#>  ellipsis      0.3.2       2021-04-29 [1] CRAN (R 4.5.1)
#>  evaluate      1.0.5       2025-08-27 [1] CRAN (R 4.5.1)
#>  fastmap       1.2.0       2024-05-15 [1] CRAN (R 4.5.1)
#>  fs            1.6.6       2025-04-12 [1] CRAN (R 4.5.1)
#>  generics      0.1.4       2025-05-09 [1] CRAN (R 4.5.1)
#>  glue          1.8.0       2024-09-30 [1] CRAN (R 4.5.1)
#>  htmltools     0.5.8.1     2024-04-04 [1] CRAN (R 4.5.1)
#>  knitr         1.50        2025-03-16 [1] CRAN (R 4.5.1)
#>  lattice       0.22-5      2023-10-24 [4] CRAN (R 4.3.3)
#>  lifecycle     1.0.4       2023-11-07 [1] CRAN (R 4.5.1)
#>  magrittr      2.0.4       2025-09-12 [1] CRAN (R 4.5.1)
#>  Matrix        1.7-4       2025-08-28 [4] CRAN (R 4.5.1)
#>  memoise       2.0.1       2021-11-26 [1] CRAN (R 4.5.1)
#>  mmrm        * 0.3.15.9001 2025-11-19 [1] Github (openpharma/mmrm@832ee0c)
#>  nlme          3.1-168     2025-03-31 [4] CRAN (R 4.4.3)
#>  pkgbuild      1.4.8       2025-05-26 [1] CRAN (R 4.5.1)
#>  pkgload       1.4.1       2025-09-23 [1] CRAN (R 4.5.1)
#>  purrr         1.1.0       2025-07-10 [1] CRAN (R 4.5.1)
#>  R6            2.6.1       2025-02-15 [1] CRAN (R 4.5.1)
#>  rbibutils     2.4         2025-11-07 [1] CRAN (R 4.5.1)
#>  Rcpp          1.1.0       2025-07-02 [1] CRAN (R 4.5.1)
#>  Rdpack        2.6.4       2025-04-09 [1] CRAN (R 4.5.1)
#>  remotes       2.5.0       2024-03-17 [1] CRAN (R 4.5.1)
#>  reprex        2.1.1       2024-07-06 [1] CRAN (R 4.5.1)
#>  rlang         1.1.6.9000  2025-11-17 [1] local
#>  rmarkdown     2.30        2025-09-28 [1] CRAN (R 4.5.1)
#>  rstudioapi    0.17.1      2024-10-22 [1] CRAN (R 4.5.1)
#>  sessioninfo   1.2.3       2025-02-05 [1] CRAN (R 4.5.1)
#>  TMB           1.9.18      2025-10-13 [1] CRAN (R 4.5.1)
#>  usethis       3.2.1       2025-09-06 [1] CRAN (R 4.5.1)
#>  vctrs         0.6.5       2023-12-01 [1] CRAN (R 4.5.1)
#>  withr         3.0.2       2024-10-28 [1] CRAN (R 4.5.1)
#>  xfun          0.53        2025-08-19 [1] CRAN (R 4.5.1)
#>  yaml          2.3.10      2024-07-26 [1] CRAN (R 4.5.1)
#> 
#>  [1] /home/nik/R/x86_64-pc-linux-gnu-library/4.5
#>  [2] /usr/local/lib/R/site-library
#>  [3] /usr/lib/R/site-library
#>  [4] /usr/lib/R/library
#>  * ── Packages attached to the search path.
```

danielinteractive:
closes #548

### DIFF
--- a/tests/testthat/test-tmb-methods.R
+++ b/tests/testthat/test-tmb-methods.R
@@ -278,7 +278,7 @@ test_that("predict will return NA if data contains NA in covariates", {
       fit,
       data = fev_data2[new_order, ],
       include = NULL,
-      drop_response = TRUE,
+      exclude = "response_var",
       na.action = "na.pass"
     )
   )


ddsjoberg:
Hello! I do have vague recollections on this update 😆 

The logic here is that all the other `model.frame()` methods return the data frame that was used to build model by default. In mmrm, depending on which columns were being returned, we were getting data sets of various sizes, so it was often unexpected and inconsistent with the other methods' default.

Also, I think this was causing an issue in the `broom.helpers` package, which is why I made the result consistent regardless which columns are returned.

Is there a reason to change this behaviour for mmrm? 🤔

ddsjoberg:
By consistency, I meant that the other methods return the data the model was estimated by, and providing a consistent experience for mmrm users would be important. 

Do we think it is a feature to return different data sets depending on which variables are requested? Users would more often be interested in the data the model was built on I would think (that is 100% of the time what I want when I call `model.frame()`). If  a user ever wanted a larger set of the data, we provide them that option with the `na.action` argument. I would strongly recommend to keep the behavior as it is (while of course addressing the env issue).

From broom.helpers, I think the package is using model.frame() for a couple of things, like getting the model N, the number of obs within each level of a categorical level and these are not possible to do if the function is updated.

NEWS.md
```diff
@@ -11,7 +11,7 @@
...
 - Previously, the `predict()` method failed when requesting an unconditional prediction interval. This is fixed now.
 - Previously, the `emp_start()` starting values could only work with variables included directly in `data`. Now they can also work when variables are not included in `data` but are available in the parent environment: The variables will be correctly included in the design matrix. This is achieved by an internal update to `h_mmrm_tmb_data()`.
-- Previously, the `model.frame()` method failed when the model formula included variables from the environment. This is fixed now.
+- Previously, the `model.frame()` method failed when the model formula included variables from the environment. This is fixed now. In addition, the `model.frame()` method now only considers selected columns (using the `include` argument) when removing rows with missing values. This can lead to more rows being kept in the returned data frame compared to previous versions.
```
NikKrieger:
```suggestion
- Previously, the `model.frame()` method failed when the model formula included variables from the environment. This is fixed now. In addition, whereas `model.frame()` used to apply `na.action` before processing the `include` argument, `na.action` is now applied afterwards. As a result, any variables excluded from `include` are no longer considered when `na.action` is applied. Therefore, in the default case when `na.action = "na.omit"`, this can lead to more rows being kept in the returned data frame compared to previous versions.
```

tests/testthat/test-tmb-methods.R
```diff
@@ -692,17 +692,35 @@ test_that("model.frame makes the levels match", {
   expect_identical(levels(fev_data$AVISIT), levels(out_frame$AVISIT))
 })
 
-test_that("model.frame ignores subject levels", {
+test_that("model.frame only considers included columns when applying na.action", {
```
NikKrieger:
```suggestion
test_that("model.frame excludes columns not in `include` when applying na.action", {
```

NikKrieger:
> If we make this update, would there be any way to get the model frame of the data used to build the model? My impression is that when a user calls `model.frame()` getting the utilized data is the desired output in nearly every call.

@ddsjoberg, I agree that `model.frame(fit)` should return the observations used to build the model, but this should not necessarily be the case if the user supplies a value to the `data`, `na.action`, or `subset` argument. In the case of the user supplying `data` in particular, the user is explicitly asking `model.frame()` to use `data` *instead of* the observations used to build the model, and `data` may very well have a totally different complexion of missingness. I think this method should follow the lead of `?stats::model.frame` if it can:

> All the variables in `formula`, `subset` and in `...` are looked for first in `data` and then in the environment of `formula` [...] and collected into a data frame. Then the `subset` expression is evaluated, and it is used as a row index to the data frame. Then the `na.action` function is applied to the data frame [...]. 

Can the `include` argument (which in practice serves to *exclude* undesired variables) be applied last—after these steps? For instance, if the user runs `model.frame(fit, include = "subject_var")` (i.e., return the original data but exclude the response, group, and visit variables), then the method can—in effect—do this:

``` r
ret <- model.frame.default(fit$formula_parts$full_formula, data = fit$data, na.action = na.action)
ret <- ret[*remove the response_var, group_var, and visit_var*]
ret
```

This would make it so that the `na.action` is applied in full view of the original data, resulting in only the observations used to build the model, and then the response, group, and visit variables are removed.

ddsjoberg:
Thanks @NikKrieger ! Apologies, for not being clear. It sounds like we're in agreement that the _default_ `model.frame.mmrm(fit)` method should return the data used to create the model. 

By specifying the other args, the default should be able to alter what is returned similar the other `model.frame(subset, na.action)` arguments and mmrm even gives further options with the `include` argument.

NikKrieger:
Seems like we're on the same page, then, @ddsjoberg! @danielinteractive, I'll propose a tweak that I think will work.

danielinteractive:
Thanks @ddsjoberg and @NikKrieger . If I understand correctly, we're in agreement that the default `model.frame.mmrm(fit)` method should return the data used to create the model.

That is still the case after this PR being merged.

So I don't see any reason to change the PR. Right?

ddsjoberg:
@danielinteractive as a user, I would be confused if I called `model.frame()` on the fit and I got 100 obs data frame returned. Then decided that I only need the group var, added that to the `include` argument, and was returned a data frame with far more obs. From a user perspective, I would keep the subsetting behaviour consistent for `include`, and leave the `subset` or `na.action` arguments to control any filtering of the data.

Regardless of what you decide, thanks for reaching out! ✌🏼 🤗

NikKrieger:
If you're interested, I implemented the approach I discussed above in which the `include` argument is applied at the end. It doesn't call `h_construct_model_frame_inputs()` anymore but it borrows its applicable pieces.

As expected, it violates the `subject_var` test, but all other unit tests pass:

``` r
model.frame.mmrm_tmb <- function(
  formula,
  data,
  include = c("subject_var", "visit_var", "group_var", "response_var"),
  full,
  na.action = "na.omit", # nolint
  ...
) {
  # nolint

  assert_class(formula, classes = "mmrm_tmb")
  include_choice <- c("subject_var", "visit_var", "group_var", "response_var")
  assert_subset(include, include_choice)

  # Only if include is default (full) and also data is missing, and also na.action is na.omit we will
  # use the model frame from the tmb_data.
  if (
    missing(data) &&
      setequal(include, include_choice) &&
      identical(h_get_na_action(na.action), stats::na.omit)
  ) {
    ret <- formula$tmb_data$full_frame
    # Remove weights column.
    ret[, "(weights)"] <- NULL
    ret
  } else {
    if (missing(data)) {
      data <- formula$data
    }
    assert_data_frame(data)

    # Construct data frame to return to users.
    ret <-
      stats::model.frame(
        formula = formula[["formula_parts"]][["full_formula"]],
        data = data,
        na.action = na.action,
        xlev = component(formula, "xlev")
      )

    if (!setequal(include, include_choice)) {
      drop_response <- !"response_var" %in% include
      add_vars <- unlist(formula$formula_parts[setdiff(include, "response_var")])
      new_formula <- h_add_terms(
        formula$formula_parts$model_formula,
        add_vars,
        drop_response
      )
      new_terms <- stats::terms(new_formula)
      cols_to_keep <- rownames(attr(new_terms, "factors"))
      cols_to_drop <- setdiff(colnames(ret), cols_to_keep)
      ret[cols_to_drop] <- NULL
      attr(ret, "terms") <- new_terms
    }
  }
  ret
}
```

I would also recommend passing `...` to `model.frame.default()`. I tested this, and it violates one test, but it's only because the test has an extraneous argument that `model.frame.default()` interprets as a column to add to the model frame:
https://github.com/openpharma/mmrm/blob/fe98ccc88b3cf879f4eab4b1b1cf1d8e8b8efa5a/tests/testthat/test-tmb-methods.R#L281

Issue #539, opened 2025-10-10
0-row contrast matrices when fit has no intercept


NikKrieger:
When a model has no intercept, `mmrm:::h_get_contrast()` yields zero-row contrast matrices for certain model terms. This causes `mmrm:::Anova.mmrm(type = "III", test.statistic = "Chisq")` to fail:

``` r
library(devtools)
#> Loading required package: usethis
library(car)
#> Loading required package: carData
library(mmrm)
#> mmrm() registered as car::Anova extension

fit <- mmrm(FEV1 ~ FEV1_BL + SEX + ARMCD + ar1(AVISIT | USUBJID), fev_data)

fit_no_intercept <-
  mmrm(FEV1 ~ FEV1_BL + SEX + ARMCD - 1 + us(AVISIT | USUBJID), fev_data)

# Works fine
car::Anova(fit, type = "III", test.statistic = "Chisq")
#> Analysis of Fixed Effect Table (Type III Chisq tests)
#>         Df Chisq Statistic Pr(>=Chisq)    
#> FEV1_BL  1         11.4571   0.0007122 ***
#> SEX      1          0.1306   0.7177725    
#> ARMCD    1         21.4769   3.581e-06 ***
#> ---
#> Signif. codes:  0 '***' 0.001 '**' 0.01 '*' 0.05 '.' 0.1 ' ' 1

# Errors out
car::Anova(fit_no_intercept, type = "III", test.statistic = "Chisq")
#> Error in solve.default(h_quad_form_mat(contrast, component(object, "beta_vcov"))): 'a' is 0-diml
```
<sup>Created on 2025-10-10 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

This is probably caused by this line 44:
https://github.com/openpharma/mmrm/blob/452ce6b63f0e6897c16af15e399d095b58888629/R/interop-car.R#L44 

A zero is in the assign vector (here called `asg`) when there is an intercept (see `?model.matrix`). The code `(!0 %in% asg)` seems to be doing the opposite of what was intended because it yields `TRUE` if there is NOT a zero in the assign matrix. Thus, when a term is only represented in one column in the model matrix (i.e., `component(fit, "x_matrix")`) and when `contains_intercept` is `TRUE`, the next line will assign `coef_rows` a value of zero:

https://github.com/openpharma/mmrm/blob/452ce6b63f0e6897c16af15e399d095b58888629/R/interop-car.R#L45

[...]

PR #552, opened 2025-11-25
Attempt to fix type-2 0-row contrast matrices

NikKrieger:
This is a second attempt to fix the zero-row contrast matrix Issue #539, which only affects type-2 matrices now. 

This replaces pull request #540 .

A few more tests may be necessary but I don't have the expertise/capacity right now to know exactly what the expected values should be.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,7 +3,6 @@
 ### New Features
 
 - `Anova.mmrm()` now has a `test.statistic` argument allowing the user to choose a Chi-squared test as opposed to the default F-test.
-- `Anova.mmrm()` can now handle intercept-free models, even when the first effect containing a categorical variable has an aliased parameter.
 - `logLik.mmrm_tmb()` now includes a `nobs` attribute containing the number of subjects in the inputted model (i.e., `component(mmrm_object, "n_subjects")`). As a result, `BIC(logLik(mmrm_object))` will now be successfully calculated.
 - `mmrm_control()` now has a `disable_theta_vcov` argument allowing the user to disable computation of the variance-covariance matrix of the covariance parameters. In that case, the `nlminb` optimizer cannot be used and other optimizers should be selected. Avoiding this computation can be useful in cases where the number of covariance parameters is large and computation of this very large matrix is slow and memory-intensive.
 - `model.frame()` has a new argument `exclude` which allows to exclude specific model variables from the returned data frame, *after* creating the initial `model.frame`. This is useful when the same rows should be kept as in the original model fit (e.g. after `na.action` has been applied), but some variables are not needed in the returned data frame. The `terms` attribute of the returned data frame is updated accordingly.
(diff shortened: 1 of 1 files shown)


vignettes/hypothesis_testing.Rmd
```diff
@@ -237,3 +238,28 @@ FEV1 ~ ARMCD * SEX + ARMCD:FEV1_BL
...
+
+#### Type-II Contrast Matrices in Intercept-free Models
+
+To accurately run a type-II ANOVA when all levels of a categorical variable are represented in a design matrix, one of the levels of each categorical variable must contain subtractive (i.e., negative) values its corresponding contrast matrix columns. Since `PROC MIXED` includes all factor levels of all categorical variables, the incorporation of negative values is a commonplace occurrence. Negative values are invariably assigned to the last columns, which are always the reference levels' columns. 
+
+However, in `mmrm::mmrm()`, reference level columns only exist in intercept-free models for the first term containing a categorical variable. Thus, mmrm's helper functions incorporate a special handling step for such first-categorical-terms in which their last contrast matrix columns are assigned straight -1s.
```
danielinteractive:
```suggestion
However, in `mmrm::mmrm()`, reference level columns only exist in intercept-free models for the first term containing a categorical variable. Thus, `mmrm`'s helper functions incorporate a special handling step for such first-categorical-terms in which their last contrast matrix columns are assigned straight -1s.
```

R/component.R
```diff
@@ -90,6 +90,7 @@ component <- function(
     "beta_est_complete",
     "beta_aliased",
     "x_matrix",
+    "x_matrix_complete",
```
danielinteractive: please add `x_matrix_complete` in the doc above

R/interop-car.R
```diff
@@ -33,52 +33,96 @@ h_type2_contrast <- function(
...
     }
   }
+  # Finally, we drop aliased columns. But we needed them to properly calculate
+  # the others.
+  l_mx <- l_mx[, !component(object, "beta_aliased"), drop = FALSE]
   l_mx[abs(l_mx) < tol] <- 0
   l_mx
 }
 
+h_first_term_containing_categ <- function(factors, categorical) {
```
danielinteractive: please add a little doc chunk above this function (we do this for all functions including internal, to help with maintenance later)

tests/testthat/test-interop-car.R
```diff
@@ -84,6 +97,50 @@ test_that("h_type2_contrast works if intercept is not given", {
...
+    matrix(
+      rep(rep(c(0, 1), 3), c(6, 1, 9, 1, 9, 1)),
+      nrow = 3,
+      byrow = TRUE,
+      dimnames = mmrm_trans_dimnames
+    )
+  )
+
+  # Testing an intercept-free model whose first term that contains a categorical
+  # variable also has an aliased coefficient
```
danielinteractive:
```suggestion
  # variable also has an aliased coefficient.
```

danielinteractive (commented):
Thanks @NikKrieger , almost there - please see few comments below, and can you please also add a `NEWS` bullet for this?

PR #556, opened 2025-12-22
Prevent Rcpp from masking valid uses of Rf_error

Enchufa2:
This PR sets the `-DRCPP_NO_MASK_RF_ERROR` flag to prevent Rcpp from masking valid calls to `Rf_error`.

At Rcpp, we plan to warn about this (see [this issue](https://github.com/RcppCore/Rcpp/issues/1247) for further discussion), and at some point stop the compilation with an error message, forcing the maintainers to take action. For valid uses of `Rf_error`, like in this package, we will provide this opt-out switch. Therefore, we kindly ask you to apply this patch and send an update to CRAN at your earliest convenience.


danielinteractive:
Thanks a lot @Enchufa2 for your work on `Rcpp` and in particular this PR! Much appreciated. 

I will then submit a new `mmrm` package version to CRAN after the CRAN holidays.

danielinteractive (approved):

Issue #560, opened 2026-03-23
Estimated means using the `emmeans` with the `counterfactuals` argument.


runehaubo:
Dear MMRM authors.

Thank you for making the `mmrm` package available to us. This is a much
needed addition to the R landscape for the analysis of RCTs.

I am fitting MMRMs and extracting estimated means using the `emmeans`
package but run into errors that I cannot find a way around. Any help
would be much appreciated.

Rather than the conventional estimated means that pertain to the
subjects actually randomized to each treatment arm, I want to compute
the causally interpretable estimated means that pertain to all subjects
had they all been randomized to each treatment arm. I usually obtain
these using the `counterfactuals` argument to `emmeans` or `ref_grid`
but either this is not supported by the `mmrm` package or I am doing
something wrong. Any assistance would be highly appreciated!

Please see the following example for details.

Get packages:

``` r
library(data.table)
library(mmrm)
library(emmeans)
```

Get data and fit MMRM:

``` r
dt <- copy(fev_data)
setDT(dt)
dt
#>      USUBJID AVISIT  ARMCD                      RACE    SEX  FEV1_BL     FEV1
#>       <fctr> <fctr> <fctr>                    <fctr> <fctr>    <num>    <num>
#>   1:     PT1   VIS1    TRT Black or African American Female 25.27144       NA
#>   2:     PT1   VIS2    TRT Black or African American Female 25.27144 39.97105
#>   3:     PT1   VIS3    TRT Black or African American Female 25.27144       NA
#>   4:     PT1   VIS4    TRT Black or African American Female 25.27144 20.48379
#>   5:     PT2   VIS1    PBO                     Asian   Male 45.02477       NA
#>  ---                                                                         
#> 796:   PT199   VIS4    TRT                     White   Male 41.05401       NA
#> 797:   PT200   VIS1    PBO Black or African American   Male 51.19101 35.70341
#> 798:   PT200   VIS2    PBO Black or African American   Male 51.19101 41.64454
#> 799:   PT200   VIS3    PBO Black or African American   Male 51.19101       NA
#> 800:   PT200   VIS4    PBO Black or African American   Male 51.19101 54.25081
#>         WEIGHT VISITN     VISITN2
#>          <num>  <int>       <num>
#>   1: 0.6767231      1 -0.62645381
#>   2: 0.8006186      2  0.18364332
#>   3: 0.7087859      3 -0.83562861
#>   4: 0.8088997      4  1.59528080
[...]

PR #567, opened 2026-05-21
Add G-computation correction for emmeans, contrasts argument

apanagos-lilly:
Adds emmeans_gcomp_vars to mmrm_control(). When set, emmeans() returns the average treatment effect with standard errors that account for covariate variability across subjects. Default is NULL, when NULL, mmrm runs identically as before. Related to #560, though the counterfactual path does not support variance correction and has 50x runtime overhead only addressable through emmeans changes.

mmrm() and fit_mmrm() now accept a contrasts argument for specifying contrast matrices or functions, matching lm(). When a contrast matrix includes levels not present in the fitting data, those levels are preserved and marked as aliased, enabling prediction on new data


NEWS.md
```diff
@@ -1,3 +1,11 @@
+# mmrm 0.3.17.9005
+
+### New Features
+
+- `mmrm()` and `fit_mmrm()` now accept a `contrasts` argument, allowing users to specify contrast matrices or functions for factor variables, matching the interface of `lm()`. When an explicit contrast matrix includes levels not present in the fitting data, those levels are preserved in the model and marked as aliased, enabling prediction on new data containing those levels (#562).
```
danielinteractive:
```suggestion
- `mmrm()` and `fit_mmrm()` now accept a `contrasts` argument, allowing users to specify contrast matrices or functions for factor variables, matching the interface of `lm()`. When an explicit contrast matrix includes levels not present in the fitting data, those levels are preserved in the model and marked as aliased, enabling prediction on new data containing those levels.
```
so far we don't link to any GH issues so no need for this

DESCRIPTION
danielinteractive, line 23: please add yourself as author or contributor here

tests/testthat/test-utils.R
```diff
@@ -296,11 +296,10 @@ test_that("emp_start works", {
...
       full_frame = full_frame,
-      visit_var,
-      subject_var,
-      subject_groups,
-      n_visits, # Additional arguments are ignored.
-      n_subjects
+      visit_var = visit_var,
+      subject_var = subject_var,
+      subject_groups = subject_groups,
+      cov_type = "us"
```
danielinteractive: should we still check that additional arguments are ignored?

tests/testthat/test-utils.R
```diff
@@ -314,21 +313,128 @@ test_that("emp_start also works with transformed variables", {
...
-    0.1593,
-    0.3227,
-    0.1882,
-    0.1708
-  )
-  expect_equal(result, expected, tolerance = 1e-4)
+    subject_groups = fit$tmb_data$subject_groups,
+    cov_type = fit$formula_parts$cov_type
+  )
+  expect_length(result, 2L)
```
danielinteractive: can we please keep comparing to the numbers as before? I want to see that this stays stable

tests/testthat/test-utils.R
```diff
@@ -314,21 +313,128 @@ test_that("emp_start also works with transformed variables", {
...
+
+test_that("h_get_par_from_emp errors for sp_exp", {
+  mat <- diag(4)
+  expect_snapshot(
+    h_get_par_from_emp(mat, cov_type = "sp_exp"),
+    error = TRUE
+  )
+})
+
+# emp_start per covariance type ----
```
danielinteractive: this should be moved upwards such that all emp_start tests are together in one section of this file

tests/testthat/test-gcomp.R
```diff
@@ -0,0 +1,234 @@
...
+  expect_true(isSymmetric(correction$delta, tol = sqrt(.Machine$double.eps)))
+
+  eigen_vals <- eigen(correction$delta, symmetric = TRUE, only.values = TRUE)$values
+  expect_true(all(eigen_vals >= -sqrt(.Machine$double.eps)))
+})
+
+# emmeans integration ----
+
+test_that("emmeans with gcomp produces larger SEs than without", {
+  skip_if_not_installed("emmeans", minimum_version = "1.6")
```
danielinteractive, line 158: Can we somewhere in this test file also compare numbers with reference values e.g. from SAS or other software or simulations? I want to make sure that we get correct results, not just the right dimensions and format.

tests/testthat/test-fit.R
```diff
@@ -1027,3 +1039,67 @@ test_that("mmrm gives warning if not reproducible TMB option is used", {
...
+  fit <- mmrm(
+    FEV1 ~ RACE + ARMCD * AVISIT + ar1(AVISIT | USUBJID),
+    data = fev_data,
+    contrasts = list(ARMCD = contr_mat)
+  )
+  expect_class(fit, "mmrm")
+})
+
+test_that("mmrm with extra-level contrasts allows prediction on new data", {
+  all_race_levels <- levels(fev_data$RACE)
```
danielinteractive: please check whether this better goes into https://github.com/openpharma/mmrm/blob/main/tests/testthat/test-tmb-methods.R#L127

vignettes/subsections/_intro-customizations.Rmd
```diff
@@ -38,7 +38,7 @@ In `mmrm` we provide two starting value functions, one is `std_start` and the ot
 `std_start` will try to use the identity matrix as the covariance, however there are convergence problems
 for `ar1` and `ar1h` if the identity matrix is provided, thus for these two covariance structures
 we use $\rho=0.5$ instead.
-`emp_start` will try to use the empirical covariance matrix of the residuals of the ordinary least squares model as the starting value for unstructured covariance structure.
+`emp_start` will try to use the empirical covariance matrix of the residuals of the ordinary least squares model to derive starting values for all non-spatial covariance types. The empirical covariance matrix will be used as the starting value for the unstructured covariance type. For other parametrized covariance types, the empirical covariance is decomposed into standard deviations and correlations. Simple moment estimators based on this decomposition are used as starting values. For example, the upper-triangular empirical correlations are averaged for the compound-symmetry correlation parameter.
```
danielinteractive: Thanks for adding this part here.

In addition, we need:

- please also mention the `contrasts` argument somewhere in this vignette
- separate, full vignette on g-computation, covering theory, algorithm and usage in the package

R/utils.R
```diff
@@ -484,6 +491,60 @@ h_get_theta_from_cov <- function(covariance) {
   unname(ret)
 }
 
+# h_theta_from_cor ----
```
danielinteractive: please add full documentation chunks for below functions, too.

R/fit.R
```diff
@@ -222,6 +222,14 @@ refit_multiple_optimizers <- function(fit, ..., control = mmrm_control(...)) {
...
 #'   variance-covariance matrix for variance parameters. This can speed up fitting
 #'   when there are many variance parameters, see details.
+#' @param emmeans_gcomp_vars (`character` or `NULL`)\cr names of variables to
+#'   treat as fixed in the G-computation correction for emmeans. When non-`NULL`,
+#'   enables the correction in [`emmeans_support`], producing the average
+#'   treatment effect (ATE) with corrected standard errors. The visit variable
+#'   (identified from the covariance structure) is computed separately per
+#'   level. All other named variables are treated as intervention (subjects
+#'   pooled across levels). All variables not named are averaged over using
+#'   each subject's actual values. Defaults to `NULL` (no correction).
```
danielinteractive:
```suggestion
#'   each subject's actual values.
```
we don't mention defaults in the text

R/fit.R
```diff
@@ -494,14 +518,21 @@ mmrm <- function(
   } else {
     attr(weights, which = "dataname") <- deparse(match.call()$weights)
   }
+
+  # Validate G-computation fixed variables if specified.
+  if (!is.null(emmeans_gcomp_vars)) {
+    assert_subset(emmeans_gcomp_vars, names(data))
+  }
```
danielinteractive: let's do this below as well to simplify the code structure

R/fit.R
```diff
@@ -466,13 +487,16 @@ mmrm <- function(
...
+  contrasts = NULL,
   reml = TRUE,
   control = mmrm_control(...),
   ...
 ) {
   assert_false(!missing(control) && !missing(...))
   assert_class(control, "mmrm_control")
   assert_list(control$optimizers, min.len = 1)
+  assert_list(contrasts, null.ok = TRUE, names = "unique")
+  emmeans_gcomp_vars <- control$emmeans_gcomp_vars
```
danielinteractive: let's just do this below for the gcomp stuff

R/fit.R
```diff
@@ -589,6 +620,20 @@ mmrm <- function(
     stop("Unrecognized coefficent variance-covariance method!")
   }
 
+  # G-computation correction: store metadata for emmeans hook.
+  fit$emmeans_gcomp_vars <- emmeans_gcomp_vars
+  if (!is.null(emmeans_gcomp_vars)) {
+    # Store subject-level covariate data from the ORIGINAL data (pre-NA-removal).
```
danielinteractive:
```suggestion
    # Store subject-level covariate data from the original data (pre-NA-removal).
```

danielinteractive: please also add a comment in the code here why this cannot be taken from `tmb_data` later

R/gcomp.R
```diff
@@ -0,0 +1,223 @@
...
+#' @noRd
+NULL
+
+#' Get Subject-Level Covariate Data
+#'
+#' @description Returns one row per subject with covariate data. Uses the
+#'   stored `emmeans_gcomp_subject_data` from the original dataset when available
+#'   (includes subjects with missing outcomes), otherwise falls back to
+#'   the fitted model's complete-case frame.
+#'
```
danielinteractive, line 21: add a note here that if there are time-varying covariates in the data then just the first value will be included in the returned data set. So the idea here really is for constant, "baseline", covariates to be returned.

R/gcomp.R
```diff
@@ -0,0 +1,223 @@
...
+#' @param object (`mmrm`)\cr the fitted MMRM.
+#' @param subj_data (`data.frame`)\cr one row per subject with covariates.
+#' @param visit_value (`string`)\cr the visit level to evaluate at.
+#' @param visit_var (`string`)\cr name of the visit variable.
+#' @param counterfactual_grid (`data.frame`)\cr unique combinations of fixed
+#'   variables (excluding visit) to loop over.
+#'
+#' @return A list with:
+#'   - `vhat`: `matrix` of fitted potential outcomes (n x K).
+#'   - `L_global`: `matrix` of globally-weighted contrast rows (K x p).
```
danielinteractive: also mention what is n, p, K here

R/gcomp.R
```diff
@@ -0,0 +1,223 @@
...
+
+  model_formula <- object$formula_parts$model_formula
+  model_terms <- stats::delete.response(stats::terms(model_formula))
+
+  K <- nrow(counterfactual_grid)
+  n_subj <- nrow(subj_data)
+  p <- length(beta_hat)
+
+  # Stack all K counterfactual assignments into one data frame.
+  # One model.frame + model.matrix call instead of K.
```
danielinteractive:
```suggestion
  # Stack all K counterfactual assignments into one data frame.
  # (for efficiency reasons, we can just do one model.frame and model.matrix call below instead of K).
```

R/gcomp.R
```diff
@@ -0,0 +1,223 @@
...
+  nd_list <- vector("list", K)
```
danielinteractive: please use a better name than `nd` here and below

R/gcomp.R
```diff
@@ -0,0 +1,223 @@
...
+  for (k in seq_len(K)) {
+    nd <- subj_data
+    nd[[visit_var]] <- factor(visit_value, levels = levels(full_frame[[visit_var]]))
+    for (var_name in names(counterfactual_grid)) {
+      val <- counterfactual_grid[[var_name]][k]
+      if (is.factor(full_frame[[var_name]])) {
+        nd[[var_name]] <- factor(val, levels = levels(full_frame[[var_name]]))
+      } else {
+        nd[[var_name]] <- val
+      }
```
danielinteractive: please use pattern 

`nd[[var_name]] <- if () ... else ...`

R/gcomp.R
```diff
@@ -0,0 +1,223 @@
...
+    nd_list[[k]] <- nd
+  }
+  nd_stacked <- do.call(rbind, nd_list)
+
+  mf <- stats::model.frame(model_terms, data = nd_stacked, na.action = stats::na.pass)
+  X_all <- stats::model.matrix(model_terms, data = mf, contrasts.arg = contrasts_arg)
+  X_all <- X_all[, names(beta_hat), drop = FALSE]
+  X_all[is.na(X_all)] <- 0
+
+  # Split back into K blocks of n_subj rows.
```
danielinteractive, line 126: will come back to review this once vignette is there with algorithm description

R/gcomp.R
```diff
@@ -0,0 +1,223 @@
...
+    L_global[k, ] <- colMeans(X_k, na.rm = TRUE)
+  }
+
+  list(vhat = vhat, L_global = L_global)
+}
+
+#' Compute G-computation Correction for emmeans Integration
+#'
+#' @description Computes the corrected L matrix (global covariate weighting)
+#'   and the p x p variance correction matrix. Together these produce the
```
danielinteractive: please explain dimensions again here

R/gcomp.R
```diff
@@ -0,0 +1,223 @@
...
+      for (jj in seq_along(v_indices)) {
+        kj <- match(grid_keys[jj], cf_keys)
+        if (is.na(kj)) next
+        S_full[v_indices[ii], v_indices[jj]] <- S_t[ki, kj]
+      }
+    }
+  }
+
+  L <- L_global
+  LLt <- L %*% t(L)
```
danielinteractive: use `tcrossprod`

R/gcomp.R
```diff
@@ -0,0 +1,223 @@
...
+
+  LLt_inv <- tryCatch(
+    solve(LLt),
+    error = function(e) {
+      svd_result <- svd(LLt)
+      tol <- max(dim(LLt)) * max(svd_result$d) * .Machine$double.eps
+      d_inv <- ifelse(svd_result$d > tol, 1 / svd_result$d, 0)
+      svd_result$v %*% diag(d_inv) %*% t(svd_result$u)
```
danielinteractive:
```suggestion
      MASS::ginv(LLt)
```
?

danielinteractive: actually I would just directly use that. because it will just give the standard inverse if it exists

R/gcomp.R
```diff
@@ -0,0 +1,223 @@
...
+
+      for (jj in seq_along(v_indices)) {
+        kj <- match(grid_keys[jj], cf_keys)
+        if (is.na(kj)) next
+        S_full[v_indices[ii], v_indices[jj]] <- S_t[ki, kj]
+      }
+    }
+  }
+
+  L <- L_global
```
danielinteractive: could be nicer to move the S_full, L_global calculation into a helper function and also document that a little bit more there

R/gcomp.R
```diff
@@ -0,0 +1,223 @@
...
+    solve(LLt),
+    error = function(e) {
+      svd_result <- svd(LLt)
+      tol <- max(dim(LLt)) * max(svd_result$d) * .Machine$double.eps
+      d_inv <- ifelse(svd_result$d > tol, 1 / svd_result$d, 0)
+      svd_result$v %*% diag(d_inv) %*% t(svd_result$u)
+    }
+  )
+
+  delta <- t(L) %*% LLt_inv %*% S_full %*% LLt_inv %*% L
```
danielinteractive: if we just want this at the end I feel we could avoid the full matrix (pseudo) inverse and just work with a decomposition of LLt and then solve a linear system?
just thinking how to optimize this `delta` calculation here.

R/interop-emmeans.R
```diff
@@ -139,15 +139,89 @@ emm_basis.mmrm <- function(
...
     estimability::all.estble
   }
-  dfargs <- list(object = object)
-  dffun <- function(k, dfargs) {
-    mmrm::df_md(dfargs$object, contrast = k)$denom_df
+
+  # Start with the model-based covariance.
+  V <- component(object, "beta_vcov")
+
+  if (!is.null(object$emmeans_gcomp_vars)) {
```
danielinteractive: can we move this into helper function and apply here? otherwise this thing gets too long

R/interop-emmeans.R
```diff
@@ -139,15 +139,89 @@ emm_basis.mmrm <- function(
...
+    # Detect aliased coefficient positions before any modifications.
+    aliased_pos <- which(is.na(beta_hat))
+
+    # Apply G-computation correction if specified.
+    visit_var <- object$formula_parts$visit_var
```
danielinteractive: will come back for review once algorithm vignette is there

R/tmb.R
```diff
@@ -177,12 +177,37 @@ h_mmrm_tmb_data <- function(
   }
   full_frame <- full_frame[data_order, ]
 
+  contrasts_preserve_vars <- NULL
+  if (!is.null(contrasts)) {
+    formula_factors <- intersect(
```
danielinteractive, line 208: ok we need some explanations here. either in comments or in the function documentation. please also refer to how lm() is doing this, is it the same or different and why

R/utils.R
```diff
@@ -401,21 +401,27 @@ std_start <- function(cov_type, n_visits, n_groups, ...) {
...
 #'
 #' @details
-#' This `emp_start` only works for unstructured covariance structure.
-#' It uses linear regression to first obtain the coefficients and use the residuals
-#' to obtain the empirical variance-covariance, and it is then used to obtain the
-#' starting values.
+#' `emp_start` supports all non-spatial covariance types.
+#' It uses linear regression to first obtain the coefficients
+#' and uses the residuals to obtain the empirical
+#' variance-covariance matrix. This is then decomposed into
```
danielinteractive: can you please include more algorithm details here? as this is the only place where this is documented

tests/testthat/test-fit.R
```diff
@@ -1027,3 +1039,67 @@ test_that("mmrm gives warning if not reproducible TMB option is used", {
   )
   TMB::config(tmbad_deterministic_hash = 1, DLL = "mmrm")
 })
+
+# mmrm contrasts argument ----
+
+test_that("mmrm accepts contrasts argument with contrast function", {
+  fit <- mmrm(
```
danielinteractive, line 1034: can we also check that the result is as expected? I mean the coefficients for ARMCD

danielinteractive (commented):
Nice start, thanks @apanagos-lilly ! Please see comments below.

For next time, let's please try to have separate PRs for the separate enhancements as much as possible (here it would be 3 as far as I can see). That usually speeds up reviews, resolution and merges.

One additional high level comment: 
I wonder if there is a nice way we can print to the user when they use `emmeans` on an `mmrm` with g-computation arguments that average treatment effect estimates are calculated. To basically highlight this as message, or hook into the show/print of `emmeans` results or similar, to make it very visible for the user.

R/gcomp.R
apanagos-lilly: note added, thanks.

R/gcomp.R
apanagos-lilly: vignette has been added.

tests/testthat/test-gcomp.R
apanagos-lilly: ref values from ind. rep. code added

apanagos-lilly:
@danielinteractive, thanks for all your great feedback. First, yes I agree, keeping these separate would have been better. We reverted out the pr which added the `emp_start` additions so now it is only g-computation and the contrast. The vignette for g-compute has been added and a `message()` too when it is called.

DESCRIPTION
apanagos-lilly: added.

danielinteractive:
Thanks @apanagos-lilly can you please check the open comments above, if they are no longer applicable please resolve them. Please let me know once all comments resolved and ready for my second review

tests/testthat/test-utils.R
apanagos-lilly: Reverted with emp_start removal

tests/testthat/test-utils.R
apanagos-lilly: Reverted with emp_start removal

tests/testthat/test-utils.R
apanagos-lilly: Reverted with emp_start removal

tests/testthat/test-fit.R
apanagos-lilly: Checked. The test uses the contrasts argument on mmrm(), so I think its natural to keep it in `test-fit.R`?

vignettes/subsections/_intro-customizations.Rmd
apanagos-lilly: Contrasts are documented in the 'Fine control over design matrix' section of this vignette.  G-computation vignette was added via `model_robust_estimator.Rmd`

R/utils.R
apanagos-lilly: Reverted with emp_start removal

R/fit.R
apanagos-lilly: Moved validation below

R/fit.R
apanagos-lilly: Moved to bottom of `mmrm()`

R/fit.R
apanagos-lilly: Made lowercase. Added comment explaining

R/gcomp.R
apanagos-lilly: Mentioned what n, p, K are now here. Thanks

R/gcomp.R
apanagos-lilly: used your wording, thanks

R/gcomp.R
apanagos-lilly: Renamed to `subj_cf`, `cf_list` and `cf_stacked`

R/gcomp.R
apanagos-lilly: Uses that pattern now, thanks

R/gcomp.R
apanagos-lilly: Added, dimensions of delta and l global explained here

R/gcomp.R
apanagos-lilly: Thanks, changed

R/gcomp.R
apanagos-lilly: Replaced, thanks

R/gcomp.R
apanagos-lilly: placed in `h_gcomp_visit_contributions()` with documentation

R/gcomp.R
apanagos-lilly: So walking through the edge cases, `solve()` would work for full rank. However, in the case of when it isn't full rank, tcrossprod is singular. In the singular case, the psuedoinverse is still needed.

`MASS::ginv` is able to compute both cases. I think that the computational cost is fairly negligible because the dimensions needed to be inverted are typically small.

R/interop-emmeans.R
apanagos-lilly: Yes, placed in `h_emm_basis_gcomp()`

R/interop-emmeans.R
apanagos-lilly: vignette has been added, thanks

R/utils.R
apanagos-lilly: Reverted with emp_start removal

R/tmb.R
apanagos-lilly: Added comments above here which explains how lm() is doing this and how that's insufficient and how contrasts fixes it

tests/testthat/test-fit.R
apanagos-lilly: check added using the coefficient for ARMCD

apanagos-lilly:
> One additional high level comment: I wonder if there is a nice way we can print to the user when they use `emmeans` on an `mmrm` with g-computation arguments that average treatment effect estimates are calculated. To basically highlight this as message, or hook into the show/print of `emmeans` results or similar, to make it very visible for the user.

Added a `message()` in `emm_basis.mmrm()`

apanagos-lilly:
@danielinteractive, I believe all comments have been tended to. Please let me know if I missed any. Thanks!

tests/testthat/test-fit.R
danielinteractive: hm let's better move it because the point is that the `predict()` result is as expected

R/fit.R
danielinteractive: ok thanks, however now looking at it again, I feel it would be much better if this data saving code part would live not here but instead in h_mmrm_tmb_data(). The result and also the `gcomp_vars` should then also be in the tmb_data (which makes more sense because these are data things and not really fit things) And again I am just double checking - in tmb_data we already save the full data set anyway... Is that not sufficient as a super set of what you need downstream? I wonder if tmb_data can just save the `gcomp_vars` thing and then later you can combine that with `data` to get what you need?

R/gcomp.R
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  contributions <- h_gcomp_visit_contributions(
+    object, model_mat, grid, fixed_vars, visit_var
+  )
+  S_full <- contributions$S_full
+  L <- contributions$L_global
+
+  LLt_ginv <- MASS::ginv(tcrossprod(L))
+  delta <- crossprod(L, LLt_ginv) %*% S_full %*% (LLt_ginv %*% L)
```
danielinteractive: can be slightly more efficient (save one multiplication operation) if you first save `LLt_ginv %*% L` in a variable and then multiply that onto `S_full` from left and right

vignettes/subsections/_intro-customizations.Rmd
```diff
@@ -319,3 +319,52 @@ the covariance matrix. As a consequence, we model the correlation between `VIS2`
...
+In some scenarios, we may want to control how the design matrix is constructed in finer detail. This can be especially of interest if there are factor levels not present in the current dataset, which nonetheless are valid levels that should not result in later errors (for example, when using `predict` functions). In the example below, `mmrm` might not run if only one factor level is available.
+
+```{r contrasts-fail-example, error = TRUE}
+ex_data <- fev_data[!(fev_data$RACE == "White" & fev_data$SEX == "Male"),]
+fit_subgroup1 <- mmrm(
+  formula = FEV1 ~ SEX + ARMCD * AVISIT + us(AVISIT | USUBJID),
+  data = ex_data[ex_data$RACE != "White",],
+  vcov = "Asymptotic"
+)
+fit_subgroup2 <- mmrm(
```
danielinteractive, line 337: can you check the example again in detail? Already this call here errors for me with:

```
Error in `contrasts<-`(`*tmp*`, value = contr.funs[1 + isOF[nn]]) : 
  contrasts can be applied only to factors with 2 or more levels
```

vignettes/model_robust_estimator.Rmd
```diff
@@ -0,0 +1,111 @@
...
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+```
+
+## Background
+
+After fitting a particular MMRM, one must commonly answer the question:
```
danielinteractive: Add the typical context where this matters (e.g. comparing different treatment groups in a clinical trial or other experiment) - MMRMs can also be used in other settings

vignettes/model_robust_estimator.Rmd
```diff
@@ -0,0 +1,111 @@
...
+
+After fitting a particular MMRM, one must commonly answer the question: 
+
+> What was the mean value of the outcome variable in a treatment group?
+
+One method to answer this question is to use G-computation to estimate the marginal means of each treatment group. In this section, we describe the G-computation estimator in the context of the MMRM and describe an approach to variance estimation.
+
+## G-Computation Estimator
+
+There are $J$ treatment groups, $T$ planned timepoints, and $n$ total subjects. Define
```
danielinteractive: Can you please check the notation again? We need to be consistent across the vignettes here. Please read through https://openpharma.github.io/mmrm/latest-tag/articles/algorithm.html 
e.g. I see that there we have $m$ possible time points (and not $T$ as here)

danielinteractive: and not just here but the whole vignette

vignettes/model_robust_estimator.Rmd
```diff
@@ -0,0 +1,111 @@
...
+The G-computation estimator and correct variance can be implemented using the `emmeans` package. To enable correct variance estimation, there is
+an option `emmeans_gcomp_vars` inside `mmrm_control()` that captures which variables can be treated as fixed or non-random. For example, since
+every subject is is predicted from the model at the same time point and under each treatment, a model including `TRTP` and `AVISIT` could include
+`emmeans_gcomp_vars = c("TRTP","AVISIT")`.
+
+Using `emmeans` with option `weights = "proportional"` and passing in the `data` with rows for every subject and timepoint can produce $\hat{\theta}_{tj}$.
+
+Under this call:
+```r
+emmeans(fit, ~TRTP | AVISIT, weights = "proportional", data=data)
```
danielinteractive: please always use correct style, e.g. here 
```suggestion
emmeans(fit, ~TRTP | AVISIT, weights = "proportional", data = data)
```

vignettes/model_robust_estimator.Rmd
```diff
@@ -0,0 +1,111 @@
...
+```
+the `emmeans` function will create a linear combination matrix. For example, the row $k=(t-1)J + j$ of this matrix is
+$$
+L^{global}_k = \frac{1}{n} \sum_{i=1}^n x_{tji}^{\top}
```
danielinteractive: better with 
```suggestion
L^{\text{global}}_k = \frac{1}{n} \sum_{i=1}^n x_{tji}^{\top}
```
etc.

danielinteractive (commented):
Thanks @apanagos-lilly for the updates! Please see additional comments below.

I am still trying to optimize how and where this data set lives or is created to avoid duplication...

apanagos-lilly:
>ok thanks, however now looking at it again, I feel it would be much better if this data saving code part would live not here but instead in h_mmrm_tmb_data(). The result and also the `gcomp_vars` should then also be in the tmb_data (which makes more sense because these are data things and not really fit things) And again I am just double checking - in tmb_data we already save the full data set anyway... Is that not sufficient as a super set of what you need downstream? I wonder if tmb_data can just save the `gcomp_vars` thing and then later you can combine that with `data` to get what you need?

So it did not let me reply below your comment, (here it is in the main body). After inspecting, yes I think your point is correct. Just include in tmb_data and this simplifies. I have modified it such that control$emmeans_gcomp_vars gets passed into `h_mmrm_tmb_data()` now instead. Much more simple.

R/gcomp.R
apanagos-lilly: Great point. Changed.

vignettes/subsections/_intro-customizations.Rmd
apanagos-lilly: So when I run `rmarkdown::render("vignettes/introduction.Rmd")` I get no error. The first contrasts chunk has `error = TRUE` to show the failure. The second chunk with the contrasts fix runs without error.

vignettes/model_robust_estimator.Rmd
apanagos-lilly: `\text` added where needed.

vignettes/model_robust_estimator.Rmd
apanagos-lilly: fixed, thanks.

vignettes/model_robust_estimator.Rmd
apanagos-lilly: Yes, should now have consistent notation, thanks.

vignettes/model_robust_estimator.Rmd
apanagos-lilly: example added

apanagos-lilly:
All comments tended to, ready for next review

vignettes/subsections/_intro-customizations.Rmd
danielinteractive: Yes, the rendering works without errors because of the `error = TRUE`. But I don't think that you want this error:

Screenshot 2026-06-05 at 9 20 38 AM

danielinteractive: Also can you please style the vignette or something, we need blanks between `=` and left and right e.g.

R/fit.R
```diff
@@ -589,6 +615,9 @@ mmrm <- function(
     stop("Unrecognized coefficent variance-covariance method!")
   }
 
+  # G-computation vars stored in tmb_data for emmeans hook
+  fit$emmeans_gcomp_vars <- attr(tmb_data, "emmeans_gcomp_vars")
```
danielinteractive: Do we need this here? Can't we look it up later from `tmb_data` attribute?

R/tmb.R
```diff
@@ -61,6 +61,17 @@ h_mmrm_tmb_formula_parts <- function(
...
+#'   matrices or contrast functions (like [stats::contr.sum] or
+#'   [stats::contr.poly]) for specific factor variables, matching the
+#'   `contrasts` argument in [stats::lm()]. The list names must correspond to
+#'   factor variable names in the model formula. When `NULL` (the default),
+#'   the contrasts set on the factor variables in `data` are used. If a
+#'   contrast matrix has rownames that include levels not present in `data`,
+#'   those levels are preserved and the corresponding model matrix columns
+#'   are marked as aliased (not estimable), enabling prediction on new data
+#'   containing those levels.
 #'
```
danielinteractive, line 78: the new argument needs to be documented

R/tmb.R
```diff
@@ -319,7 +360,8 @@ h_mmrm_tmb_data <- function(
       subject_groups = subject_groups,
       n_groups = n_groups
     ),
-    class = "mmrm_tmb_data"
+    class = "mmrm_tmb_data",
+    emmeans_gcomp_vars = emmeans_gcomp_vars
```
danielinteractive: this should not be an attribute, it should be part of the list (as it is also documented above)

R/interop-emmeans.R
```diff
@@ -139,16 +139,131 @@ emm_basis.mmrm <- function(
...
   }
-  dfargs <- list(object = object)
-  dffun <- function(k, dfargs) {
-    mmrm::df_md(dfargs$object, contrast = k)$denom_df
+
+  # Start with the model-based covariance.
+  V <- component(object, "beta_vcov")
+
+  if (!is.null(object$emmeans_gcomp_vars)) {
+    gcomp_result <- h_emm_basis_gcomp(
```
danielinteractive, line 147: instead of looking in `object` directly we should just look in the `tmb_data` list in the object

R/gcomp.R
```diff
@@ -0,0 +1,256 @@
...
+#' treatment and use their distribution to correct the L matrix and covariance.
+#'
+#' @keywords internal
+#' @noRd
+NULL
+
+#' Get Subject-Level Covariate Data
+#'
+#' @description Returns one row per subject with covariate data. Uses the
+#'   stored `emmeans_gcomp_subject_data` from the original dataset when
```
danielinteractive: please adjust documentation

R/gcomp.R
```diff
@@ -0,0 +1,256 @@
...
+#'   - `L_global`: `matrix` (n_grid x p) globally-weighted L matrix.
+#'
+#' @keywords internal
+#' @noRd
+h_gcomp_emm_correction <- function(object, model_mat, grid) {
+  assert_class(object, "mmrm")
+  assert_matrix(model_mat, mode = "numeric")
+  assert_data_frame(grid)
+
+  fixed_vars <- object$emmeans_gcomp_vars
```
danielinteractive: please take it from the `tmb_data` element in `object`

danielinteractive (commented):
Thanks @apanagos-lilly , almost there, just a few easier comments now. I also read through and polished a bit the vignette.

R/fit.R
apanagos-lilly: Good point, deleted

R/interop-emmeans.R
apanagos-lilly: Yes, updated

R/gcomp.R
apanagos-lilly: Yep, updated

R/tmb.R
apanagos-lilly: Yes it should... the reason for why it was an attribute was because `TMB::MakeADFun()` did not accept character vectors. But I figured a work around. Made `c("ARMCD", "AVISIT")` be  `"ARMCD,AVISIT"` and handled vector conversion downstream. So fixed, now part of list.

R/gcomp.R
apanagos-lilly: Adjusted

R/tmb.R
apanagos-lilly: Updated with new argument

vignettes/subsections/_intro-customizations.Rmd
apanagos-lilly: Ok thanks for the snippet, so the downstream `emmeans` aren't even being ran like they are supposed to... wrong error is occurring

apanagos-lilly: Need have scenario where the actually wanted error occurs

apanagos-lilly: > Also can you please style the vignette or something, we need blanks between `=` and left and right e.g.

Spacing between `=` added

apanagos-lilly: So now the error should occur at the 2nd emmeans example rather than mmrm step

apanagos-lilly:
@danielinteractive, all comments have been tended to, ready for review. Thanks.

R/tmb.R
danielinteractive: Ah ok I see... hm, I guess that is the downside when we add it to the list that it will be taken up by TMB as well - which is does not really have to! I am a bit worried that the workaround will break when we have variable names with commas inside (very rare I guess but possible, e.g. `data.frame(`A,B` = 3, C= 4, check.names = FALSE)` has that). I guess to avoid that we could 1) revert to attribute or 2) have an exclusion of this element before passing the list to `TMB::MakeADFun()` or 3) use a factor instead of character? (at least from the docs factor seems accepted)

apanagos-lilly: I think 2) is the cleanest fix? Keep the input as a character vector and then a one-line adjustment when passing into `TMB::MakeADFun()`? I can try and see how this looks, should be a simple modification.

danielinteractive: yep sounds good

apanagos-lilly: Ok, now keeping as character vector and stripping it down (or out) before every applicable `TMB::MakeADFun()` call.

vignettes/subsections/_intro-customizations.Rmd
danielinteractive: ok looks good now!

R/tmb.R
```diff
@@ -698,8 +697,10 @@ fit_mmrm <- function(
     start = control$start,
     n_groups = tmb_data$n_groups
   )
+  tmb_data_tmb <- tmb_data
+  tmb_data_tmb$emmeans_gcomp_vars <- NULL
```
danielinteractive, line 701: add a comment here why this is needed

danielinteractive:
Thanks @apanagos-lilly , almost there - just letting Copilot have a quick last look, please check if any meaningful comments come out of it.

R/tmb.R
```diff
@@ -61,6 +61,21 @@ h_mmrm_tmb_formula_parts <- function(
...
+#'   factor variable names in the model formula. When `NULL` (the default),
+#'   the contrasts set on the factor variables in `data` are used. If a
+#'   contrast matrix has rownames that include levels not present in `data`,
+#'   those levels are preserved and the corresponding model matrix columns
+#'   are marked as aliased (not estimable), enabling prediction on new data
+#'   containing those levels.
+#' @param emmeans_gcomp_vars (`character` or `NULL`)\cr names of variables
+#'   treated as fixed for G-computation correction. Stored as a
+#'   comma separated string in the returned list for downstream use
+#'   in the emmeans hook.
```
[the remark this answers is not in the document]
apanagos-lilly: Used suggested fix

vignettes/model_robust_estimator.Rmd
```diff
@@ -0,0 +1,131 @@
...
+= L^{\text{global}} \hat{V} L^{\text{global}\top} + S.
+$$
+where $S$ is a block-diagonal matrix corresponding to the relevant empirical covariance
+$\frac{1}{n} \hat{\Sigma}_v$ at each time point.
+
+This turns out to involve solving a linear system
+$$
+\hat{V}_c = 
+\hat{V} + (L^{\text{global}} L^{\text{global}\top})^{-} L^{\text{global}\top} S L^{\text{global}} (L^{\text{global}} L^{\text{global}\top})^{-},
+$$
```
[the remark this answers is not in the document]
apanagos-lilly: Used suggested alignment fix

R/gcomp.R
```diff
@@ -0,0 +1,255 @@
...
+  p <- length(beta_hat)
+
+  # Stack all K counterfactual assignments into one data frame.
+  # (for efficiency reasons, we can just do one model.frame and model.matrix call below instead of K).
+  cf_list <- vector("list", K)
+  for (k in seq_len(K)) {
+    subj_cf <- subj_data
+    subj_cf[[visit_var]] <- factor(
+      visit_value, levels = levels(full_frame[[visit_var]])
+    )
```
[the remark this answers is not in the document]
apanagos-lilly: I think this solution is a bit unnecessary for our use case. We can just add a `assert_subset(visit_var, names(object$tmb_data$full_frame))` to catch this edge case.

apanagos-lilly:
@danielinteractive, your one comment and the copilot comments have been tended to. Thanks.

danielinteractive (commented):
Thanks @apanagos-lilly it looks pretty good now!

Just one last thing: I had a look at the coverage, see [coverage-report.html](https://github.com/user-attachments/files/28818667/coverage-report.html) and I see that in `gcomp.R` there are a few lines not reached yet by any of the tests. Since this is all quite complex stuff, if it is not difficult for you to come up with a few more tests that can cover these lines, it would be great.

apanagos-lilly:
@danielinteractive, coverage for `gcomp.R` is now 100% from 96.3%. Added tests for additional edge cases. Two lines (185 and 191) are unreachable because a mismatch can't occur under the current pipeline. Though I think keeping the lines in place is the right move if there's ever a refactor, so added `# nocov` for those two lines.

apanagos-lilly:
Also there has been desire from groups to have a CRAN push for these changes, they are unable to simply install from GitHub. I don't know if you are tracking this request but I said I would ask you regarding a timeline for a CRAN push. Thanks!

danielinteractive (approved):
Cool, thanks a lot @apanagos-lilly ! I will try to get this to CRAN in the next few days.

Issue #561, opened 2026-04-13
`emmeans` support fails for MMRMs fitted to data with only one (post baseline) visit


runehaubo:
It is really great that the ANCOVA edge case (MMRM with only 1 post baseline visit) is supported by `mmrm` because
it simplifies code significantly when I don’t have to condition on
the number of post baseline visits. Ufortunately, it appears that
`emmeans` support fails for this case.

Please see the following example for details and let me know if you require additional information.

Get packages:

``` r
library(data.table)
library(mmrm)
library(emmeans)
```

Get data for first visit only:

``` r
dt <- copy(fev_data)
setDT(dt)
mydata <- dt[AVISIT == "VIS1"] |> droplevels()
mydata[, .N, AVISIT]
#>    AVISIT     N
#>    <fctr> <int>
#> 1:   VIS1   200
```

Fit ANCOVA model using `mmrm` (MMRM edge case):

``` r
fm2 <- mmrm(FEV1 ~ FEV1_BL + ARMCD + us(AVISIT | USUBJID),
  data = mydata)
```

We can fit the model using `lm` to confirm that this is simply an
ANCOVA:

``` r
fm2_lm <- lm(FEV1 ~ FEV1_BL + ARMCD, data = mydata)
all.equal(coef(fm2_lm), coef(fm2))
#> [1] TRUE
all.equal(logLik(fm2), logLik(fm2_lm, REML=TRUE), check.attributes = FALSE)
#> [1] TRUE
```

Unfortunately the `emmeans` support does not extend to the ANCOVA edge
case:

``` r
em <- try(emmeans(fm2, ~ ARMCD)) # Error related to factor levels
#> Error in `contrasts<-`(`*tmp*`, value = contr.funs[1 + isOF[nn]]) : 
#>   contrasts can be applied only to factors with 2 or more levels
```

This is what would have hoped for:

``` r
emmeans(fm2_lm, ~ ARMCD)
#>  ARMCD emmean    SE  df lower.CL upper.CL
#>  PBO     32.4 0.794 131     30.8     33.9
#>  TRT     36.9 0.806 131     35.3     38.5
#> 
#> Confidence level used: 0.95
```

Session-info:

``` r
sessionInfo()
#> R version 4.5.2 (2025-10-31 ucrt)
#> Platform: x86_64-w64-mingw32/x64
#> Running under: Windows 11 x64 (build 22631)
#> 
#> Matrix products: default
#>   LAPACK version 3.12.1
#> 
[...]

PR #568, opened 2026-06-17
Fix issue #561

apanagos-lilly:
This is a fix for issue #561. Currently `visit_var` with only 1 level get passed to `emm_basis.mmrm()` and this breaks on the contrasts. Removing at the `recover_data.mmrm()` along with other exclusions ensures all passes to `emm_basis.mmrm()` are covered. Confirmed to pass the exact edge case presented in #561. Three tests added confirming the fix.


apanagos-lilly:
This pr has old history that's already been merged. Will close this pr and open pr from clean branch instead.

PR #569, opened 2026-06-17
Fix issue #561

apanagos-lilly:
This is a fix for issue #561. Currently `visit_var` with only one level get passed to `emm_basis.mmrm()` and this breaks on the contrasts. Removing at the `recover_data.mmrm()` along with the existing exclusions ensures all passes to `emm_basis.mmrm()` are covered. Confirmed to pass the exact edge case presented in #561. Three tests added confirming the fix.


danielinteractive (approved):
Perfect, thanks a lot @apanagos-lilly !

apanagos-lilly:
Thanks! Yes, got lucky finding that `recover_data.mmrm()` is entirely upstream to `emm_basis.mmrm()` and already appeared to precisely handle exclusions like this so fortunately the exact structure already existed making the fix simple.

